PR #31, opened 2019-08-30
Prune Terraform Package

appilon:
This was originally an effort to prune the entire SDK. I removed some backend artifacts from helper/schema and then removed as much as possible without editing files in the terraform package. We would be able to remove quite a bit outside the API surface if `helper/resource.testing` weren't so reliant on it. If we managed to cook up a new testing framework that didn't import any go code but wrapped a TF executable most of the `terraform` package could go.

The method I used for this pruning is tedious. I would like to extend or build upon staticcheck if we want more thorough pruning as this effort might not be worth repeating manually.

For future pruners, the method is very gross but works as follows:
1) Identify your API surface of a package (files containing methods/types imported by providers)
2) Delete every non-test file (does not end in `_test.go`) from that package, excluding API surface files
3) build package over and over, restoring missing code, grep in a side terminal of the Trash bin was very helpful
4) Once the package builds, ensure other packages build via `go test ./...`. Missing code will error before any test files attempt to run
5) At this point you might need to restore some non-test files that were used to help in tests, such as mocks
6) You have now pruned "dead" code relative to projects importing this package, delete test files directly testing pruned files (if you deleted `abc.go`, delete its tests `abc_test.go`).
7) Let staticcheck identify internal code that is now dead via `staticcheck -checks U1000 ./...` and delete it

Unfortunately while this is a great "first pass". If you pruned an adjacent package (say `helper/resource`) next, you might have now caused code in the `terraform` package to now be "dead", and will need to repeat the process 🤢. This method also doesn't do smart things like realize 1 file, that might import 6 other types, contains only 1 type you need. The smart thing to do would be isolate the 1 dependency and cull the rest. This is where a tree shaking tool for Go would be great, staticcheck is too naive in its decision making.

UPDATE:

This PR will be put on hold. We will have more visibility into breaking changes once providers actually depend on this codebase. It's too difficult for any individual to review and say with certainty the changes are safe


appilon:
Technically this PR cannot be merged without forcing a major version bump as these functions and types are "public". While this code is likely not imported on much if at all in the wild. This PR cannot be merged.

PR #186, opened 2019-09-27
cherry_pick.sh: unstage added ignored files

kmoe:
Cherry-picking Core PRs often results in `website/` files being added, which is incorrect. We can't ignore all added files as we still want new files in relevant subdirectories. Without bringing back the whitelist, this PR solves the problem quickly and safely by unstaging added files that match the `IGNORE_FILES` regex.

### DIFF
--- a/scripts/cherry_pick/cherry_pick.sh
+++ b/scripts/cherry_pick/cherry_pick.sh
@@ -18,6 +18,9 @@ git cherry-pick --no-commit --mainline 1 "$COMMIT_ID"
 echo "Unstaging files removed by us..."
 git status --short | sed -n 's/^DU //p' | ifne xargs git rm
 
+echo "Unstaging added files that match the ignore list..."
+git status --short | sed -n 's/^A  //p' | grep -Ef ./scripts/cherry_pick/IGNORE_FILES | ifne xargs git rm -f
+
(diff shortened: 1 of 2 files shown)


scripts/cherry_pick/cherry_pick.sh
```diff
@@ -18,6 +18,9 @@ git cherry-pick --no-commit --mainline 1 "$COMMIT_ID"
 echo "Unstaging files removed by us..."
 git status --short | sed -n 's/^DU //p' | ifne xargs git rm
 
+echo "Unstaging added files that match the ignore list..."
+git status --short | sed -n 's/^A  //p' | grep -Ef ./scripts/cherry_pick/IGNORE_FILES | ifne xargs git rm -f
```
radeksimko: Is there a reason we are limiting this to additions only? Is it because updates and deletions will cause conflict and require manual resolution anyway?

kmoe: Good question. Lines 24-29 deal with the case where changes other than addition are made to files matching the `IGNORE_FILES` regex. 

We have to deal with the additions first using a different method than that on lines 24-29, because running `git reset`, `git checkout`, or `git restore` will unstage but not remove the file, given the unmerged state of that file in the index following an unfinished `git cherry-pick`.

PR #193, opened 2019-10-03
helper/schema: Parse timeouts for the data sources

kayrus:
For https://github.com/hashicorp/terraform/pull/21713

/cc @radeksimko

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -395,6 +395,16 @@ func (r *Resource) ReadDataApply(
 		return nil, err
 	}
 
+	// Instance Diff shoould have the timeout info, need to copy it over to the
+	// ResourceData meta
+	rt := ResourceTimeout{}
+	if _, ok := d.Meta[TimeoutKey]; ok {
+		if err := rt.DiffDecode(d); err != nil {
+			log.Printf("[ERR] Error decoding ResourceTimeout: %s", err)
+		}
+	}
+	data.timeouts = &rt
+
 	err = r.Read(data, meta)
 	state := data.State()
 	if state != nil && state.ID == "" {


radeksimko:
Hi @kayrus 
Sorry for the delay in reviewing this PR.

We appreciate you mentioned the `http_data_source` use case, but we'd still like to understand the use case better.

For context - these customizable resource timeouts were designed (and are mostly used) for timeouts after _retries_, not low-level HTTP connection timeouts (time to first byte, TLS handshake etc.).

In a typical scenario the provider code is repeatedly requesting response which is received in a timely manner (usually within sub-second or a few seconds), but the operation (e.g. database creation) is in progress which is apparent from the response.

Please don't take this as nuisance - we just want to make sure we have the right solution for the problem you're experiencing and attempting to solve. We'd also want to document at least one such problem (use case) on the website to guide future developers on how this should be used.

---

To comment on the code being submitted:

It turns out the current design of CRUD interfaces has a few flaws which also affect customizable timeouts in its current form - see https://github.com/hashicorp/terraform-plugin-sdk/issues/234

I am not sure how are we going to address these yet, but I'm worried that adding this feature in its current form would open more ways for developers to introduce more bugs. The current interface used for resources doesn't seem to be a good fit for data sources as it allows defining 4 timeouts for different purposes (CRUD), 3 of which are likely going to be irrelevant in context of data sources:

```go
func dataSource() *schema.Resource {
    return &schema.Resource{
        Read:   dataSourceRead,

            "name": {
                Type:     schema.TypeString,
                Computed: true,
            },
        },

            Read:   schema.DefaultTimeout(45 * time.Minute),
            // ▼▼▼
            Create: schema.DefaultTimeout(45 * time.Minute),
            Update: schema.DefaultTimeout(45 * time.Minute),
            Delete: schema.DefaultTimeout(45 * time.Minute),
            // ▲▲▲
        },
    }
}
```

Data sources should never attempt to create, update or delete anything and allowing them to define timeouts for these actions may suggest otherwise.

---

For these reasons I'm afraid we can't accept this patch in its current form, however we’re looking to work towards a similar goal - as such I'd suggest [opening an issue](https://github.com/hashicorp/terraform-plugin-sdk/issues/new) for this feature where we can discuss the use case(s) more in depth (and that may help us to prioritise https://github.com/hashicorp/terraform-plugin-sdk/issues/234).

PR #214, opened 2019-10-28
Remove testing package from helper/schema

dnovikoff:
"testing" package dependency causes users of helper/schema to import testing flags in their apps as they are defined on static

There is a simular problem in helpers/resource package.
TestMain defined in helpers/resource  and this brings more problems. Probably testing.go should be extracted to separate package, but that change would probably break compatibility. It also contains staticly defined flags, that could affect user applications.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/mitchellh/colorstring v0.0.0-20190213212951-d06e56a500db
 	github.com/mitchellh/copystructure v1.0.0
 	github.com/mitchellh/go-homedir v1.1.0
+	github.com/mitchellh/go-testing-interface v1.0.0
 	github.com/mitchellh/go-wordwrap v1.0.0
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/mitchellh/reflectwalk v1.0.1
--- a/helper/schema/testing.go
+++ b/helper/schema/testing.go
@@ -1,14 +1,13 @@
 package schema
 
 import (
-	"testing"
-
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	testing "github.com/mitchellh/go-testing-interface"
 )
 
 // TestResourceDataRaw creates a ResourceData from a raw configuration map.
 func TestResourceDataRaw(
-	t *testing.T, schema map[string]*Schema, raw map[string]interface{}) *ResourceData {
+	t testing.T, schema map[string]*Schema, raw map[string]interface{}) *ResourceData {
 	t.Helper()
 
 	c := terraform.NewResourceConfigRaw(raw)


radeksimko:
Hi @dnovikoff 
thank you for the contribution.

This looks like a reasonable change compatibility-wise, but I'd like to understand the motivation behind it in the context of your provider - do you mind sharing a snippet of code or describing what exactly you intend to do in your provider acceptance tests that requires this patch?

radeksimko:
Friendly ping.

I understand you may be busy, but to allow our team focusing on active PRs and features/bugs I will have to close this PR in the next 7 days if I receive no further response.

Thank you for understanding.

dnovikoff:
Sorry for late reply.

For instance we have
https://github.com/terraform-providers/terraform-provider-yandex/blob/master/main.go

That includes
github.com/terraform-providers/terraform-provider-yandex/yandex

The file
https://github.com/terraform-providers/terraform-provider-yandex/blob/master/yandex/data_source_yandex_compute_image.go

Includes
github.com/hashicorp/terraform-plugin-sdk/helper/schema

And as this package includes "testing" in its non _test.go files, the testing flags are added in static init to the terraform-provider-yandex binary. That does not cause much trouble, but I would prefer to keep the binary free of testing flags

radeksimko:
@dnovikoff I may be missing some important context here - apologies if that's the case.

What kind of testing flags do you mean and how do they get registered? I was under the impression that any binaries for test purposes are produced via `go test`, hence generally not available in a binary that was compiled through standard `go build` tooling or any tooling that wraps it (such as `gox`).

The `testing` library only has [`Init()`](https://golang.org/pkg/testing/#Init) with capital `I` which does _not_ run automatically when the package is imported, unlike `init()`.

dnovikoff:
That got fixed go1.13.5
https://github.com/golang/go/commit/fbc6a972226f889d2ab1150468755615098ee80f#diff-70e628298261565d825f7199d13042f2L252

As you see, before it was static var init. So for the users of the the earlier versions that flags would be parsed and shown in the flag.Usage()

dnovikoff:
Look at the foolowing example
example.go
```
package main

import (
	"flag"
	"testing"
)

var _ = &testing.T{}

func main() {
	flag.Parse()
	flag.Usage()
}
```

When running go1.13.5 Dockerfile
```
FROM golang:1.13.5

ADD ./example.go /example.go
RUN go run /example.go
```

You get
`Usage of /tmp/go-build698730291/b001/exe/example:`

When running go1.12 Dockerfile
```
FROM golang:1.12

ADD ./example.go /example.go
RUN go run /example.go
````

You get

```
Usage of /tmp/go-build751915921/b001/exe/example:
  -test.bench regexp
    	run only benchmarks matching regexp
  -test.benchmem
    	print memory allocations for benchmarks
  -test.benchtime d
    	run each benchmark for duration d (default 1s)
  -test.blockprofile file
    	write a goroutine blocking profile to file
  -test.blockprofilerate rate
    	set blocking profile rate (see runtime.SetBlockProfileRate) (default 1)
  -test.count n
    	run tests and benchmarks n times (default 1)
  -test.coverprofile file
    	write a coverage profile to file
  -test.cpu list
    	comma-separated list of cpu counts to run each test with
  -test.cpuprofile file
    	write a cpu profile to file
  -test.failfast
    	do not start new tests after the first test failure
  -test.list regexp
    	list tests, examples, and benchmarks matching regexp then exit
  -test.memprofile file
    	write an allocation profile to file
  -test.memprofilerate rate
    	set memory allocation profiling rate (see runtime.MemProfileRate)
  -test.mutexprofile string
    	write a mutex contention profile to the named file after execution
  -test.mutexprofilefraction int
    	if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
  -test.outputdir dir
    	write profiles to dir
  -test.parallel n
    	run at most n tests in parallel (default 12)
  -test.run regexp
    	run only tests and examples matching regexp
  -test.short
    	run smaller test suite to save time
  -test.testlogfile file
    	write test action log to file (for use only by cmd/go)
  -test.timeout d
    	panic test binary after duration d (default 0, timeout disabled)
  -test.trace file
    	write an execution trace to file
  -test.v
    	verbose: print additional output
```

**Comment by radeksimko** [comment]
I see - these examples with extra context are really useful, thank you!

I now understand the reasoning behind it, but I still struggle to understand how does this affect a Terraform provider specifically, when providers are generally executed in a very "controlled way" by Terraform core and [never outside of it](https://github.com/hashicorp/go-plugin/blob/a1756f37cec6471f10accdf15a5e5961317bea0d/server.go#L187-L191)?

i.e. How did you discover this?

---

It's worth noting that the `testing` package is actually imported by a few other packages, namely:

```
github.com/hashicorp/terraform-plugin-sdk/helper/acctest
github.com/hashicorp/terraform-plugin-sdk/helper/resource
github.com/hashicorp/terraform-plugin-sdk/terraform
```

the last 2 of which tend to be imported by providers outside of tests for valid reasons, so we'd need to work on eliminating the `testing` package from these too.

The process for these would be probably much more involved as we use more than just `testing.T` struct from the `testing` package (e.g. [`testing.M`](https://github.com/hashicorp/terraform-plugin-sdk/blob/7d2b5732beb0f5a5be6ceec1ceb522f9b53fe1a2/helper/resource/testing.go#L98) or [`testing.Verbose()`](https://github.com/hashicorp/terraform-plugin-sdk/blob/7d2b5732beb0f5a5be6ceec1ceb522f9b53fe1a2/helper/resource/testing.go#L541)) and that isn't covered by Mitchell's `go-testing-interface` (probably by design as the design intentions there were different from what we're discussing here).

I wonder if the effort outweighs the benefits and and risk (of breaking something because we didn't mimic something in `testing` package correctly), especially when this would be binned when 1.12 is deprecated. 🤔

radeksimko:
We discussed this topic in our team meeting yesterday also.

In the long run we'd like to decouple the testing framework entirely into its own dedicated package to avoid this problem altogether. As you can imagine this will involve some breaking changes and as such these cannot be taken lightly - we need to propose deprecation plan etc.

Again - thank you for bringing this up, but if we can't come up with a reasonably easy solution for this I'd prefer to close it and focus on other areas of the SDK which can eventually take us to the long-term solution I described above.

I will revisit this PR again on Friday and close it, if there's no further response. I hope the explanations make sense.

dnovikoff:
I've found this issue, while making special static linting for our code. The goal was to avoid deps for testing package for all our binaries as we found testing flags in some of our binaries.

The problem now seems not a big deal as we are moving to the next version of go and the other isses already resolved.

It is true, that for binaries of providers this does not result a problem as is is designed to be used as a plugin.

I think the issue could be resolved

PR #215, opened 2019-10-28
[DO NOT MERGE] SDK without the acceptance test framework

appilon:
This PR painstakingly cuts the acceptance test related identifiers from `helper/resource`, alongside everything it depended on. This PR should not be merged but should serve as the compass for knowing what parts of `internal/` and the blackhole sized `terraform/` packages are actually needed by our true "SDK" in the future. There is likely more that can be cut as more public identifiers in `helper/`, `plugin/` and `terraform/` might not be part of what we originally identified as [The Surface](https://github.com/radeksimko/terraform-provider-sdk-exposure). This also made no attempt to sift through the contents of `internal/`, I simply cut a package once it was no longer needed for compilation.

staticcheck is a great tool but sadly it cannot consider anything public (starts with Capital letter) as "unused", even if it resides within `internal/` (perhaps an opportunity for a PR to that project).

We might want to use this PR for reshuffling the code internally to better group which parts are needed for the SDK and what parts were only pulled along for the acceptance test framework. We could move the internal import paths around such that anything for the acceptance tests only within `internal/` are prefixed with `internal/acctestsdep/` or something like that. The files deleted from `terraform/` in this PR can't change packages without causing a "breaking change" from a SemVer point of view, but we could suffix the filename with `_acctestsdep.go`. Files that are needed for the SDK but have many parts only needed for tests (`terraform/state.go`) could have comments indicating what sections exist solely for the tests:
```
// NEEDED FOR ACCEPTANCE TESTS
 func Foo()
```


dominikh:
> staticcheck is a great tool but sadly it cannot consider anything public (starts with Capital letter) as "unused", even if it resides within internal/ (perhaps an opportunity for a PR to that project).

`staticcheck` has the `-unused.whole-program`, with which identifiers that aren't used in any of the packages specified on the command line are considered unused. Of course that creates "false positives" for public APIs that are intended to be imported by 3rd parties, but you could grep the output for `internal/`.

appilon:
@dominikh Thank you for that information and my apologies for not doing full research! I am still looking for an excuse to contribute to staticcheck!

appilon:
closing in favor of future v2 branch/pr

PR #232, opened 2019-11-02
hashibot: configure provider org migrator

appilon:
Simply label an issue `provider/<name>` and it will be migrated just like how Core is handled. The `name` is lifted from the label and the repo name is constructed as `<new_owner>/<repo_prefix><name>`


appilon:
@radeksimko Correct the code would error trying to create a new issue against a repo that doesn't exist. Should be good to go.

PR #255, opened 2019-11-19
setup proto4 guardrails via feature flag

appilon:
begin defining new resource_diff feature


appilon:
The RFC and idea was abandoned, closing for now, can be restored if the feature flag work is to be revived

Issue #248, opened 2019-11-13
Support the pseudo dynamic type


paultyng:
Tracking issue for adding [DynamicPseudoType](https://godoc.org/github.com/zclconf/go-cty/cty#Type) type support.

paultyng:
An example that could be replaced with dynamic: https://www.terraform.io/docs/providers/aws/r/iam_policy.html#example-usage

```hcl
resource "aws_iam_policy" "policy" {
  name        = "test_policy"
  path        = "/"
  description = "My test policy"

  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "ec2:Describe*"
      ],
      "Effect": "Allow",
      "Resource": "*"
    }
  ]
}
EOF
}
```

Could be:

```hcl
resource "aws_iam_policy" "policy" {
  name        = "test_policy"
  path        = "/"
  description = "My test policy"

  policy = {
    Version = "2012-10-17"
    Statement = [
      {
        Action = ["ec2:Describe*"]
        Effect = "Allow"
        Resource = "*"
      }
    ]
}
```

skylerto:
I am definitely interested in this great feature! I have a few requirements out of an internal provider I'm writing to house what can be generic JSON passed to an API, similar to the generality of IAM policies! 😄 

I'm currently doing the opposite of the JSON decode here, encoding HCL any type into a String for the provider to pass to an API.

MaciejKaras:
@paultyng So I suppose we have to stick to serializing nested content via e.g. `jsonencode` function or some other method that creates a string. Or is there other more proper workaround that you are aware of?

Do you have any idea approximately when milestone `v2.0.0` containing discussed enhancement is going to be released? I need to decide whether apply a workaround for now or just wait for new SDK release.

paultyng:
@MaciejKaras we are looking to exposing this in 2.x (not 2.0), but we are probably months out for a solution for it right now if that helps you to decide, until then, yes I think the best approach is still to use typed HCL and `jsonencode` it to a string. To prepare for this to exist, you could have like an `object_json` as `TypeString` with the plan to eventually just add `object` and deprecate the `_json` suffixed attribute at a later date.

gdavison:
I have another usecase for this feature, also from the AWS Provider:

[...]

Issue #62, opened 2016-04-18
core/schema: Add support for map w/ structured values


radeksimko:
### Terraform Version

`Terraform v0.6.14`
### Terraform Configuration Files

``` hcl
resource "aws_api_gateway_stage" "boo" {
    rest_api_id = "${aws_api_gateway_rest_api.demo.id}"
    name = "test"

    method_setting "yada" {
        metrics_enabled = true
        logging_level = "DEBUG"
    }

    method_setting "bada" {
        metrics_enabled = false
        logging_level = "ERROR"
    }
}
```
### Expected Behavior

```
+ aws_api_gateway_stage.boo
    rest_api_id:                         "" => "${aws_api_gateway_rest_api.demo.id}"
    name:                                "" => "test"
    method_setting.#:                    "" => "2"
    method_setting.yada.metrics_enabled: "" => "1"
    method_setting.yada.logging_level:   "" => "DEBUG"
    method_setting.bada.metrics_enabled: "" => "0"
    method_setting.bada.logging_level:   "" => "ERROR"
```
### Actual Behavior

```
Error running plan: 1 error(s) occurred:

* method_setting: 2 error(s) decoding:

* '[bada]' expected type 'string', got unconvertible type '[]map[string]interface {}'
* '[yada]' expected type 'string', got unconvertible type '[]map[string]interface {}'
```
### Steps to Reproduce

``` go

            "method_setting": &schema.Schema{
                Type:     schema.TypeMap,
                Optional: true,
                Elem: &schema.Resource{
                    Schema: map[string]*schema.Schema{
                        "metrics_enabled": &schema.Schema{
                            Type:     schema.TypeBool,
                            Optional: true,
                        },
                        "logging_level": &schema.Schema{
                            Type:     schema.TypeString,
                            Optional: true,
                        },
                        "data_trace_enabled": &schema.Schema{
                            Type:     schema.TypeBool,
                            Optional: true,
                        },
                    },
                },
            },
```

```
$ terraform plan
```

The current schema does support `TypeMap` which translates into `map[string]interface{}`. The interface is then decoded via `mapstructure`:
https://github.com/mitchellh/mapstructure/blob/master/mapstructure.go#L70

which seems to support slice of maps, but somehow expects string.

---

I'm creating this issue as I'm working around this by putting the key inside `TypeSet` as another field and I plan to link here.

radeksimko:
[...]

Issue #57, opened 2015-06-08
helper/schema feature: nestable resources


phinze:
I've talked about this feature on several threads now. Time to centralize the conversation!
## Background

**The context**: some resources are so closely related to each other that they are nearly always considered together: Security Group / Rule, DNS Zone / Record, IAM Group / User, etc.

**The legacy implementation**: it was common to model these as "sub-resources" in the schema, e.g. [Security Group Rules Sub-Resource](https://github.com/hashicorp/terraform/blob/cf901c90d118e45ac2083b4a2834e8bf6848136a/builtin/providers/aws/resource_aws_security_group.go#L47-L137)

**The problem**: "sub-resources" are bug-prone, difficult for provider implementors to work with, and require that all details be known at resource definition time.

**Interim solution**: lean on top-level resources for their simplicity in implementation and their flexibility (e.g. [Security Group Rule Top-Level Resource](https://github.com/hashicorp/terraform/blob/cf901c90d118e45ac2083b4a2834e8bf6848136a/builtin/providers/aws/resource_aws_security_group_rule.go))

**Remaining problem**: top-level resources are clunky and verbose to work with in configs
## Proposed Solution

Add a `helper/schema` feature I'm calling "nestable resources", allowing provider authors to configure places to nest top-level resources into the definition of a related resource.

A sketch of what this might look like from the provider implementation side, using Security Group Rules as an example:

``` go
// in SecurityGroup resource definition...
"ingress": &schema.NestedResource{
  ResourceType: "aws_security_group_rule",

  // FixedAttributes are always set on the nested resource
  FixedAttributes: map[string]interface{}{
    "type" = "ingress",
  },

  // MappedAttributes define which fields of the parent resource to
  // map into the nested resource
  //    - keys: parent attribute name
  //    - values: nested resource attribute name
  MappedAttributes: map[string]string{}{
    "id": "security_group_id",
  },
},
// ...
```

This would make the following two configs equivalent:

``` js
resource "aws_security_group" "foo" {
  // ...
  ingress {
    // ...
  }
}
```

``` js
resource "aws_security_group" "foo" {
  // ...
}

resource "aws_security_group_rule" "foo" {
  security_group_id = "${aws_security_group.foo.id}"
  // ...
}
```

paultyng:
Going to merge this to #220 

[...]

PR #259, opened 2019-11-21
add relative key link for ConflictsWith field

kdudkov:
when you have a list (f.e. in field `list`) of objects, with conflicted fields inside (f.e. `field1` and `field2`), you can't use ConflictsWith, because you have to set full path to conflicted field, but it contains index (f.e. `list.0.field1`).
I've relative path for this case in form of `..field2`

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -1466,18 +1466,26 @@ func validateConflictingAttributes(
 	}
 
 	for _, conflictingKey := range schema.ConflictsWith {
-		if raw, ok := c.Get(conflictingKey); ok {
+		resolvedKey := resolveKey(k, conflictingKey)
+		if raw, ok := c.Get(resolvedKey); ok {
 			if raw == hcl2shim.UnknownVariableValue {
 				// An unknown value might become unset (null) once known, so
 				// we must defer validation until it's known.
 				continue
 			}
 			return fmt.Errorf(
-				"%q: conflicts with %s", k, conflictingKey)
+				"%q: conflicts with %s", k, resolvedKey)
 		}
 	}
 
 	return nil
+}
+
+func resolveKey(rootKey string, key string) string {
+	if strings.HasPrefix(key, "..") && strings.Contains(rootKey, ".") {
+		return rootKey[:strings.LastIndex(rootKey, ".")+1] + key[2:]
+	}
+	return key
 }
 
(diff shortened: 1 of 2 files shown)


radeksimko:
Hi @kdudkov 
thank you for the PR. Sadly I will have to close it for a couple of reasons I'll try to outline below.

* We are in early planning stages of transitioning the SDK from "0.11 state" to 0.12, which includes transitioning to a new addressing scheme. While the dot-based addressing scheme is in use by the SDK (and hence providers) today, it is practically phased out from core itself and there's a thick layer of code providing backward compatibility, so that these providers didn't have to rewrite all addresses just to become compatible with 0.12. This scheme however has a few flaws and that's why it's being replaced with the scheme contained within the new type system in use by 0.12 - cty. I will explain more details about differences below.
* There are risks associated with relative addressing:
  - **Outdated addresses:** Most providers undergo active development where many developers of different backgrounds (skills and knowledge of the codebase) change and refactor schemas many times over a long period. While the risk of forgetting to update addresses exists in the current (absolute) scheme, the outcome is less likely to lead to confusion as wrong absolute address will lead to a field which doesn't exist, rather than "parent field called X" which may exist, but it may not be what the developer intended after refactoring.
   - It makes it easier to build very complex and **deeply nested schemas**, which tend to be hard to understand generally and then present a challenge in ongoing maintenance. We realize that deeply nested schemas are sometimes still necessary, but we'd like to better understand these cases and help developers to avoid them. #248 #62 #57 represent some examples of how we may address this problem going forward.

---

The (current) dotted addressing scheme

 - stores addresses as `[]string`, or even `string` which makes it difficult to trace where address is being passed or mutated
 - has no consistent story about which layer of code can/should check whether the address exists or is even valid, leaving most addresses actually unchecked until the very last minute.
 - has no consistent story about how to walk through "steps" in the address and map it back to the schema or data (or both)
 - makes it practically impossible to address items within a set
 - can produce ambiguous addresses as it doesn't differentiate between keys of maps/objects and indexes of sets or lists (everything is separated by dots)

The new scheme provides us much more solid groundwork to construct addresses in much safer and unambiguous way.

---

I'm assuming you're working with deeply nested schemas and your feedback would therefore be valuable at any of the mentioned issues #248 #62 #57 . Alternatively I invite you to open a new issue if you feel that your problem can't be addressed by any of those.

PR #262, opened 2019-11-26
Binary acceptance test driver

### LINKED ISSUE #13: Allow provider discovery as part of testing framework
Acceptance tests of some providers depend on other providers, commonly on `template`, `tls`, or K8S depends on GCP (as it uses GCP to create some storage to test K8S storage APIs).

This has previously been achieved by importing providers as Go packages into the providers which depend on them. Each such provider brings in its own dependency tree which has impact on the dependency tree of the provider being tested. Adding a new provider is also a lot more involved process than it should/can be - one has to add the provider as a dependency and raise a big PR, before actually using it.

It would therefore be great if the testing framework just had the capability of discovering providers, like the CLI does today. That way developers don't need to worry about what providers do they depend on as these will just be downloaded automatically in the form of binaries, as part of each test run.

We might not want to download the same binary (provider of a particular version) more than once during 1 test run. Providers may have hundreds of MB. It would therefore be great if providers can be cached during each test run and downloaded only once.

kmoe:
This PR adds a new optional test driver for Terraform provider acceptance tests that runs the Terraform CLI.

To minimise work for provider developers at this stage, the API follows as closely as possible the functionality of the legacy test driver, minimising changes needed to acceptance test code.

Since we are targeting a 1.x.x release of the provider SDK, no breaking changes will be made, and the legacy test driver will only be removed in version 2.0.0.

## Usage

The new test driver must be enabled by initialising the test helper in the [`TestMain()`](https://golang.org/pkg/testing/#hdr-Main) function in all provider packages that run acceptance tests. Most providers have only one package.

Import the new `"github.com/hashicorp/terraform-plugin-sdk/acctest"` package and add the following `TestMain` function to `provider_test.go`:

```
func TestMain(m *testing.M) {
	acctest.UseBinaryDriver("provider_name", Provider)
	resource.TestMain(m)
}
```

Some providers already have a `TestMain()` defined, usually for the purpose of enabling test sweepers (as described in [Extending Terraform: Sweepers](https://www.terraform.io/docs/extend/testing/acceptance-tests/sweepers.html)). These additional occurrences should be removed.

### Known issue: `TypeSet` attribute checks

A limitation of the binary acceptance test driver is that attribute checks addressing `TypeSet` elements will fail.

These attribute checks are rarely used in the legacy test framework, as the hashcode used for addressing Set elements in the flatmap representation must be calculated during the test. Please see https://github.com/hashicorp/terraform-plugin-sdk/issues/196.

A follow-up improvement to the SDK type system will fix this issue and provide an idiomatic way to check complex type attributes.

### Opt-out per `TestCase`

Initialising the binary test helper using `acctest.UseBinaryDriver()` causes all tests to be run using the new binary driver. Until SDK version 2.0.0, the `DisableBinaryDriver` boolean property can be used to use the legacy test driver for individual `TestCase`s. 

### Multi-provider testing

Providers that require resources from other providers during acceptance tests have two options:
  - _Standard provider discovery mechanism._ During acceptance test runs, Terraform CLI will attempt to locate providers on the local filesystem, and will download any missing providers in the usual way. **This is the default behaviour.**
  - _Binaries in the provider repo._ Relying on the Terraform CLI provider discovery mechanism can be avoided by including auxiliary provider binaries in `terraform.d/plugins/$GOOS_$GOARCH/$provider_name`, relative to the provider source code directory root. The environment variable `TF_ACC_PROVIDER_ROOT_DIR` must be set to the path of the source code directory root to use this feature. Providers can be included in binary or zip format.

It is no longer necessary to import other Terraform providers as Go modules: these imports should be removed.

Further examples will be added to this section as I raise PRs to enable the binary test driver in HashiCorp-owned providers.

Closes #13

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,8 @@ require (
 	github.com/hashicorp/hcl/v2 v2.0.0
 	github.com/hashicorp/logutils v1.0.0
 	github.com/hashicorp/terraform-config-inspect v0.0.0-20191115094559-17f92b0546e8
+	github.com/hashicorp/terraform-json v0.4.0
+	github.com/hashicorp/terraform-plugin-test v1.2.0
 	github.com/hashicorp/terraform-svchost v0.0.0-20191011084731-65d371908596
 	github.com/hashicorp/yamux v0.0.0-20181012175058-2f1d1f20f75d // indirect
 	github.com/keybase/go-crypto v0.0.0-20161004153544-93f5b35093ba
@@ -42,7 +44,7 @@ require (
 	github.com/posener/complete v1.2.1 // indirect
 	github.com/spf13/afero v1.2.2
 	github.com/vmihailenco/msgpack v4.0.1+incompatible // indirect
-	github.com/zclconf/go-cty v1.1.0
+	github.com/zclconf/go-cty v1.2.1
 	github.com/zclconf/go-cty-yaml v1.0.1
 	golang.org/x/crypto v0.0.0-20190820162420-60c769a6c586
 	golang.org/x/net v0.0.0-20191009170851-d66e71096ffb
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -20,8 +20,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/logutils"
-	"github.com/mitchellh/colorstring"
-
+	"github.com/hashicorp/terraform-plugin-sdk/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/logging"
 	"github.com/hashicorp/terraform-plugin-sdk/internal/addrs"
 	"github.com/hashicorp/terraform-plugin-sdk/internal/command/format"
@@ -32,6 +31,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/internal/states"
 	"github.com/hashicorp/terraform-plugin-sdk/internal/tfdiags"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/mitchellh/colorstring"
 )
 
 // flagSweep is a flag available when running tests on the command line. It
@@ -108,6 +108,9 @@ func TestMain(m *testing.M) {
 			os.Exit(1)
 		}
 	} else {
+		if acctest.TestHelper != nil {
+			defer acctest.TestHelper.Close()
+		}
 		os.Exit(m.Run())
 	}
 }
@@ -562,6 +565,12 @@ func Test(t TestT, c TestCase) {
 			t.Fatal(err)
 		}
 		providers[name] = p
+	}
+
+	if acctest.TestHelper != nil {
+		// inject providers for ImportStateVerify
+		RunNewTest(t.(*testing.T), c, providers)
+		return
 	}
 
 	providerResolver, err := testProviderResolver(c)
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -328,6 +328,11 @@ type TestCase struct {
 	// IDRefreshIgnore is a list of configuration keys that will be ignored.
 	IDRefreshName   string
 	IDRefreshIgnore []string
+
+	// DisableBinaryDriver forces this test case to run using the legacy test
+	// driver, even if the binary test driver has been enabled.
+	// This property will be removed in version 2.0.0 of the SDK.
+	DisableBinaryDriver bool
 }
 
 // TestStep is a single apply sequence of a test, done within the
@@ -567,7 +572,7 @@ func Test(t TestT, c TestCase) {
 		providers[name] = p
 	}
 
-	if acctest.TestHelper != nil {
+	if acctest.TestHelper != nil && c.DisableBinaryDriver == false {
 		// inject providers for ImportStateVerify
 		RunNewTest(t.(*testing.T), c, providers)
 		return
--- a/helper/resource/testing_new.go
+++ b/helper/resource/testing_new.go
@@ -1,0 +1,179 @@
+package resource
+
+import (
+	"fmt"
+	"log"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/hashicorp/terraform-plugin-sdk/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	tftest "github.com/hashicorp/terraform-plugin-test"
+)
+
+func getState(t *testing.T, wd *tftest.WorkingDir) *terraform.State {
+	jsonState := wd.RequireState(t)
+	state, err := shimStateFromJson(jsonState)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return state
+}
+
+func RunNewTest(t *testing.T, c TestCase, providers map[string]terraform.ResourceProvider) {
+	spewConf := spew.NewDefaultConfig()
+	spewConf.SortKeys = true
+	wd := acctest.TestHelper.RequireNewWorkingDir(t)
+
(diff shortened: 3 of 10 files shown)


plugin/client.go
```diff
@@ -25,7 +25,7 @@ func ClientConfig(m discovery.PluginMeta) *plugin.ClientConfig {
 		Managed:          true,
 		Logger:           logger,
 		AllowedProtocols: []plugin.Protocol{plugin.ProtocolGRPC},
-		AutoMTLS:         true,
+		AutoMTLS:         false,
```
appilon: I'm not sure this code (entire file) ever executes from the SDK's perspective, it was removed in the version2 branch.

helper/resource/testing_new.go
```diff
@@ -0,0 +1,615 @@
...
+			}
+			appliedCfg = step.Config
+			continue
+		}
+
+		t.Fatal("Unsupported test mode")
+	}
+	// If we never checked an id-only refresh, it is a failure.
+	// TODO KEM: why is this here? does this ever happen?
+	// if idRefresh {
```
appilon: Do we know if any of "the big providers" even use this? What value does it provide, I'm not sure it's a modern practice I would be okay if we made providers "delete" those tests or rewrite them?

appilon:
One minor aspect that needs to be covered is how to providers that call into our `resource.TestMain()` can get setup with the new framework. It is used to run sweepers (maybe this is what you were referring to on the todo list)

plugin/client.go
radeksimko: I would hope it doesn't, because it's the client gRPC part - i.e. the part core uses to talk to plugins.

kmoe: Thanks both, you're quite right. This change isn't needed in the SDK and I've raised a PR against Core for the necessary flag: https://github.com/hashicorp/terraform/pull/23563

helper/resource/testing.go
```diff
@@ -559,6 +559,12 @@ func Test(t TestT, c TestCase) {
 		providers[name] = p
 	}
 
+	if acctest.TestHelper != nil {
+		// inject providers for ImportStateVerify
+		RunLegacyTest(t.(*testing.T), c, providers)
```
appilon: the providers map is something I managed to cut as cruft in the version2 branch, wondering if we can get away without it? Can't remember if I authored this comment but seems like its lingering around for ImportStateVerify

helper/resource/testing_new.go
```diff
@@ -0,0 +1,615 @@
...
+	"github.com/davecgh/go-spew/spew"
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/hashicorp/terraform-plugin-sdk/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/internal/addrs"
+	"github.com/hashicorp/terraform-plugin-sdk/internal/tfdiags"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func shimAttributeValues(flatmap map[string]string, currentKey string, value interface{}) {
```
appilon: In general a second set of eyes on the shimming/recursion would be good, likely from the Core team? I just winged this (unless you thoroughly also have investigated this already and made fixes and/or found it good to go)

helper/resource/testing_new.go
```diff
@@ -0,0 +1,615 @@
...
+		// the state is empty
+		return state, nil
+	}
+
+	if err := shimStateModule(state, jsonState.Values.RootModule); err != nil {
+		return nil, err
+	}
+
+	// shimming of lists and maps might be incorrect
+	for key, output := range jsonState.Values.Outputs {
```
appilon: I really had no clue if this part "worked", would be best to test against a provider that makes checks against outputes, this is common in datasource tests, one that checks against a list or map would be a great candidate (I wasn't concerned much about the string case)

helper/resource/testing_new.go
```diff
@@ -0,0 +1,615 @@
...
+				t.Fatalf(
+					"[ERROR] Test: ID-only test failed: %s", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func testIDRefresh(c TestCase, t *testing.T, wd *tftest.WorkingDir, step TestStep, r *terraform.ResourceState) error {
```
appilon: amazing that you reverse engineered whatever this was? I have never written an "IDRefresh" test myself and I don't think its covered in the extend docs.

helper/resource/testing.go
radeksimko: I suppose we can keep this in when merging into `master` (as opt-in) and then remove it in `version2` branch. This would align with the risk profile you set by removing it in `version2` I expect?

appilon: I agree we should remove the conditional check that acts as an "opt-in" in `version2`. Unfortunately it's the map passed to `RunLegacyTest` called `providers` that I cut from `version2` and once this is merged to master I will need to rebase `version2` and restore it. Unless of course we trace `providers` through `RunLegacyTest` and find a way to remove it.

It's not a big deal I'm sure we will get the merge conflicts solved one way or another.

helper/resource/testing_new.go
```diff
@@ -0,0 +1,188 @@
...
+func getState(t *testing.T, wd *tftest.WorkingDir) *terraform.State {
+	jsonState := wd.RequireState(t)
+	state, err := shimStateFromJson(jsonState)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return state
+}
+
+func RunLegacyTest(t *testing.T, c TestCase, providers map[string]terraform.ResourceProvider) {
```
radeksimko: Why is this function called `Legacy` when it's the one that calls binary Terraform? 🤔  Shouldn't it be called something like `RunNewTest`?

appilon: This was an artifact from me if I recall. I think my mindset was there would be a future where we have a NEW testing framework/API, and this function is called to run tests defined in our "old" API (the current one, resource.TestCase etc).

appilon: Probably too forward thinking haha.

kmoe: I'll change this to `RunNewTest` to match `testing_new.go`, which I think is helpful naming while the SDK contains both the legacy and new test frameworks.

When the legacy test framework is removed, we can change these names as part of the major version bump.

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/hashicorp/errwrap v1.0.0
 	github.com/hashicorp/go-cleanhttp v0.5.1
-	github.com/hashicorp/go-getter v1.4.0
+	github.com/hashicorp/go-getter v1.4.1
```
radeksimko: Was there a strong reason for bumping this? The reason I'm asking is because this seems to require Go 1.13 as described in https://github.com/hashicorp/go-getter/pull/233

kmoe: No, I think v1.4.0 will work fine. Will revert this.

helper/resource/testing_new.go
kmoe: Copying some data here from a conversation elsewhere:

There are 806 occurrences of ID refresh tests in `terraform-providers` providers:
```
$ grep -r --exclude-dir="vendor" --include="*_test.go" "IDRefresh" * | wc -l
806
```

and more than half of these are in "big" providers:

```
$ grep -r --exclude-dir="vendor" --include="*_test.go" "IDRefresh" * | grep -e aws -e azure -e google -e kubernetes -e oracle | wc -l
454
```

helper/resource/testing_new_config.go
```diff
@@ -0,0 +1,106 @@
...
+				if v, ok := m.Resources[c.IDRefreshName]; ok {
+					idRefreshCheck = v
+				}
+
+				break
+			}
+		}
+
+		// If we have an instance to check for refreshes, do it
+		// immediately. We do it in the middle of another test
```
appilon: I'm assuming there are file races happening?

PR #272, opened 2019-12-08
Makefile: Exclude vendor/ directory from gofmt

LorbusChris:


appilon:
Hello @LorbusChris! What situation are you running into where this needed? We have not committed any `vendor/` folder to this project as it is intended to be consumed as a library.

LorbusChris:
oh right. nevermind then. I ran into this when I had to make some changes to this codebase so I could import it alongside `terraform` in another project. (Unfortunately there are some duplicated types between here and terraform, which breaks things due to gob and proto panicking when trying to initialize them both at the same time. I'll raise that as a separate issue.

PR #275, opened 2019-12-10
Context Aware CRUD funcs

appilon:
This PR allows a developer to switch a CRUD(E) (Yes including the not popular exists func) to one that receives a `context.Context` from GRPC. The context has a timeout added to it right before being passed to the CRUD func. `InternalValidate` will error if both a matching CRUD and Context CRUD func is set (ie, developer passed `Read` and `ReadContext`).

Overall this is not yet well tested and there are some known caveats.

1. Technically coded as is, this is a "breaking change". I have added a context onto a few `schema.Resource` methods. This all works because from `grpc_provider.go` we can call a Resource's `Apply/Refresh/DataApplyRead` directly (Did not break `schema.Provider` or the `terraform.ResourceProvider` interface). We have entered a grey area where we are breaking an "API surface" that was not really intended to be exposed outside Terraform internals. If we want to be dogmatic about the definition of a breaking change I believe this could still land in V1 with some copy paste/changes.
2. A provider developer needs to recognize that implementing a context aware CRUD method likely makes their provider incompatible with Terraform 0.11 (which goes through a different codepath `terraform.ResourceProvider` which does not pass along any context). With this in mind we can:
    * Just hope the provider developer adheres to this with clear documentation
    * Enforce it with the technique detailed in an RFC I wrote a few weeks ago
    * Land this in the `version2` branch instead, I think there is benefit to landing this in V1 if possible though.
    * We should make sure its not possible to enter the changed `Resource` paths with a `nil` context, but the developer has specified and expecting to use a context aware CRUD func.

### DIFF
--- a/helper/schema/provider.go
+++ b/helper/schema/provider.go
@@ -291,7 +291,7 @@ func (p *Provider) Apply(
 		return nil, fmt.Errorf("unknown resource type: %s", info.Type)
 	}
 
-	return r.Apply(s, d, p.meta)
+	return r.Apply(nil, s, d, p.meta)
 }
 
 // Diff implementation of terraform.ResourceProvider interface.
@@ -330,7 +330,7 @@ func (p *Provider) Refresh(
 		return nil, fmt.Errorf("unknown resource type: %s", info.Type)
 	}
 
-	return r.Refresh(s, p.meta)
+	return r.Refresh(nil, s, p.meta)
 }
 
 // Resources implementation of terraform.ResourceProvider interface.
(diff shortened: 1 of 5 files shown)


appilon:
It is possible to do this without any breaking changes, it would have just involved some copy paste/duplicate implementations of the 4 affected methods on `Apply/Refresh/RefreshWithoutUpgrade/ReadDataApply`.

Anyone USING these contexts (defining even just one resource that has one of these context aware funcs) has made their provider incompatible with protocol4/tf0.11 so for that reason as well its not worth landing this in V1, going to close and PR against version2

PR #276, opened 2019-12-11
Context aware SDK with timeouts

appilon:
This change would allow developers to switch CRUD funcs and CustomizeDiff funcs to ones that receives a `context.Context`, configured with the appropriate timeouts (in the case of CRUD). This PR makes some breaking changes to a few `schema.Resource` methods by adding the `context.Context` as the first argument. This likely won't cause problems in the wild because a `schema.Resource's` `Apply/Refresh/etc` methods were not really meant to be public (they are not useful to anyone).

This PR also forces cancellation through the new context API, removing the previous `StopContext` feature (this is a breaking change).

This is a PR against our working `version2` branch. That branch I have taken the liberty of privatizing/internalizing and pruning dead code/code never intended for public use (cutting the internalized version for Core, and removing support for protocol 4). The PR represents the first enhancement w/ deliberate breaking change. PRs like this should go through a full review process and not just push directly to that branch.


appilon:
Ignore the first commit of this PR, that is the last commit of the `version2` branch I forgot to push... not sure how to get the PR to sync 🤔

helper/schema/shims_test.go
```diff
@@ -147,7 +147,7 @@ func TestShimResourceApply_create(t *testing.T) {
 		},
 	}
 
-	actual, err := r.Apply(s, d, nil)
+	actual, err := r.Apply(nil, s, d, nil)
```
paultyng: can use `context.Background` here (or `context.TODO` depending on the scenario)

helper/schema/provider.go
```diff
@@ -284,7 +284,7 @@ func (p *Provider) Apply(
 		return nil, fmt.Errorf("unknown resource type: %s", info.Type)
 	}
 
-	return r.Apply(s, d, p.meta)
+	return r.Apply(nil, s, d, p.meta)
```
paultyng: If this is dead code, can we remove it?

paultyng:
We may need the ability to opt out of timeouts, I'm not sure, we should check in with the Azure team and see how it goes with their [implementation](https://github.com/terraform-providers/terraform-provider-azurerm/blob/d413ab8cb31a4afee8ac2b4949176db9f2797c25/azurerm/internal/features/custom_timeouts.go#L8)

helper/schema/resource.go
```diff
@@ -696,6 +780,23 @@ func (r *Resource) InternalValidate(topSchemaMap schemaMap, writable bool) error
...
+	if r.ReadContext != nil && r.Read != nil {
+		return fmt.Errorf("only ReadContext or Read should be implemented, not both.")
+	}
+	if r.UpdateContext != nil && r.Update != nil {
+		return fmt.Errorf("only UpdateContext or Update should be implemented, not both.")
+	}
+	if r.DeleteContext != nil && r.Delete != nil {
+		return fmt.Errorf("only DeleteContext or Delete should be implemented, not both.")
+	}
+
```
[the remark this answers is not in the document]
appilon: Technically not needed, no issue intermixing them. If intermixing usage even works with tail-calling

```
func readFunc(d *schema.ResourceData, meta interface{}) error { ... }
...
Read: readFunc,
CreateContext: func (ctx context.Context, d *schema.ResourceData, meta interface{}) {
    return readFunc(d, meta)
},
...
```

paultyng:
What impact will this have on `StopContext`? Should it be removed?

helper/schema/shims_test.go
appilon: Ah yes, forgot about this file, I meant to use `context.Background()`

helper/schema/provider.go
appilon: It is dead from protocol 5's point of view but deleting the func will mean `schema.Provider` won't implement `terraform.ResourceProvider` and somewhere in `plugin.Serve` we are expecting an implementation of `terraform.ResourceProvider` I think it gets cast to `schema.Provider` right away though, I think the interface was only needed in V1.... I will investigate and at the very least this should have a `panic("should never be called")` good catch.

appilon:
@paultyng Yes this as its coded doesn't feel just right, as every context has a timeout set even if none was configured (as I believe the timeouts all have some hardcoded 20 min duration if nothing was was?) An opt out would make sense

helper/schema/resource.go
```diff
@@ -275,7 +299,15 @@ func (r *Resource) Apply(
 	if d.Destroy || d.RequiresNew() {
 		if s.ID != "" {
 			// Destroy the resource since it is created
-			if err := r.Delete(data, meta); err != nil {
+			var err error
+			if r.DeleteContext != nil {
+				ctx, cancel := context.WithTimeout(ctx, data.Timeout(TimeoutDelete))
+				defer cancel()
```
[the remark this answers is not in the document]
appilon: Yeah I forgot about that pattern, it's hard on the eyes but I agree safer!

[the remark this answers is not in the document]
appilon: realizing for our application I don't need to defer this cancel... I can simply cancel after each CRUD func? Right?

appilon:
Seems we might have a flaky test:
`TestWaitForState_inconsistent_negative` Seems to involve randomness and a heuristic or two.. re-running test to ensure it was a one in a hundred thing

helper/schema/resource_test.go
```diff
@@ -44,7 +45,7 @@ func TestResourceApply_create(t *testing.T) {
 		},
 	}
 
-	actual, err := r.Apply(s, d, nil)
+	actual, err := r.Contextify().Apply(context.Background(), s, d, nil)
```
appilon: throughout this file I was "lazy" and instead of rewriting all the CRUD func assignments to the context aware version I just invoke `Contextify()` on each resource before calling `Apply/Refresh/etc`.

helper/schema/provider.go
```diff
@@ -240,6 +192,8 @@ func (p *Provider) ValidateResource(
 }
 
 // Configure implementation of terraform.ResourceProvider interface.
+// This method should likely have Context pulled through it
+// However it will involve breaking the terraform.ResourceProvider interface
```
paultyng: I think this needs `Context` as it corresponds directly to a protocol rpc, we should just see what we can do about the interface.

helper/schema/resource.go
```diff
@@ -174,6 +183,51 @@ type Resource struct {
 	Timeouts *ResourceTimeout
 }
 
+// Contextify upgrades all set non-context funcs (CRUD, Exists, CustomizeDiff) to the
+// context aware counterparts.
+func (r *Resource) Contextify() *Resource {
```
paultyng: I think this method is a bit leaky. It should probably not be public as nobody outside should need to call it, and it has a lot of branching complexity.

You'd probably be better off just creating like an `r.create`, `r.read` or something (maybe `readInternal`), etc. Then your calls elsewhere are still a single func without their own branching logic but you don't have a public func that mutates schema.

Each function then could apply the timeout and bonus, they could easily `defer` the cancel in a tight scope.

```go
func (r *Resource) deleteInternal(ctx context.Context, ...) ... {
  if r.DeleteContext == nil {
    return r.Delete(...)
  }

  ctx, cancel := context.WithTimeout(ctx, data.Timeout(TimeoutDelete))
  defer cancel()
  return r.DeleteContext(ctx, ...)
}
```

Alternatively since the CRUD func signatures are so similar you could have something like:

```go
func shimContext(ctx context.Context, ctxFunc func(context.Context, *ResourceData, interface{}) error, legacyFunc func(*ResourceData, interface{}) error, timeout, d *ResourceData, meta interface{}) error {
  if ctxFunc == nil && legacyFunc == nil {
    // error at least one needs to be set
  }
 
  if ctxFunc == nil {
    return legacyFunc(...)
  }
  
  ctx, cancel := context.WithTimeout(ctx, timeout)
  defer cancel()
  return ctxFunc(...)
}

err := shimContext(ctx, r.CreateContext, r.Create, TimeoutCreate, d, meta)
...
```

Either way though its more functional and not mutating and the branching and responsibilities are kept smaller.

appilon: I agree it felt awkward as a public method. I do like the tighter `readInternal` suggestion, this works well for the CRUD functions but the `CustomizeDiffFunc` gets passed around by reference in a few spots (the codepaths there are a bit messier). I won't have the same opportunity to conditionally evaluate if a "context aware CDF" is set, it was easier to simply switch to the new function signature and upgrade through `contextify()`. I have now privatized this method and during runtime it is only called once, during `Resource.InternalValidate()` this also makes validation easier as I don't need double checks (`if r.Create || r.CreateContext`).

appilon: the others calls to `contextify()` are only called in tests, this was to avoid rewriting a zillion function signatures.

paultyng:
Is there anyway now that you have been in this code a bit to remove those dead functions with panics?

appilon:
@paultyng As discussed it should be possible, but I'll address it in a follow up PR (maybe I can sneak it in at the end)

helper/schema/resource.go
```diff
@@ -236,10 +259,56 @@ type StateUpgrader struct {
 type StateUpgradeFunc func(rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error)
 
 // See Resource documentation.
-type CustomizeDiffFunc func(*ResourceDiff, interface{}) error
+type CustomizeDiffFunc func(context.Context, *ResourceDiff, interface{}) error
+
+func (r *Resource) create(ctx context.Context, d *ResourceData, meta interface{}) error {
+	if r.CreateContext != nil {
```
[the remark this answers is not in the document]
appilon: this change has been made

helper/schema/resource.go
```diff
@@ -195,21 +216,36 @@ func (r *Resource) ShimInstanceStateFromValue(state cty.Value) (*terraform.Insta
...
-// See Resource documentation.
+// Deprecated: Please use the context aware equivalent DeleteContextFunc.
 type DeleteFunc func(*ResourceData, interface{}) error
 
-// See Resource documentation.
+// Deprecated: Please use the context aware equivalent ExistsContextFunc.
 type ExistsFunc func(*ResourceData, interface{}) (bool, error)
 
+// See Resource documentation.
+type CreateContextFunc func(context.Context, *ResourceData, interface{}) error
```
bflad: If we are forcing a new function signature, can we please consider two things:

* The ability to passthrough Terraform warning and error diagnostics -- this might be implicitly handled with `error` (sorry cannot remember), but if not, this is a good opportunity to inject the concept into here (protocol 4 can just drop warnings and raise errors as `error` if some conversion is necessary)
* Having one of the input parameters types or the output result type (e.g. `CreateError` that thinly wraps `error`) differentiate between Create/Read/Update/Delete functions for static analysis -- since each of these types currently share the same underlying function signature, its not easily possible to determine which type of action function it is to write linting tooling.

appilon: Hey Brian thanks for the feedback!
 * Yes there are plans to accomplish this, it would likely arrive in a follow up PR, but we can have developers returns a new error type (without changing the return type) and in `grpc_provider.go` try and cast to this enhanced type to give warnings/diags. We also discussed it being a pointer type (pointer, slice, map) attached to `ctx` so the developer can optionally fill it in and again we check it has been filled in `grpc_provider.go`.

* I'm not as up to speed on the challenges static analysis is facing. Even though all these funcs are types, I'm guessing because they are all the same signature your tooling is blocked. Without introducing something to the signature that differentiates them I'm not how this can be fixed? I see changing the signature in that way as less likely, as it might become hard to keep everything working with the shims. We will of course consider this though if the demand is there, it just might not arrive for a few iterations.

helper/customdiff/testing_test.go
```diff
@@ -1,11 +1,12 @@
 package customdiff
 
 import (
+	"context"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
```
paultyng: i think this needs a newline here, from `goimports`

PR #291, opened 2020-01-10
Change comment on `IDRfreshName` to avoid confusion

magodo:
The original wording in comment will mislead user to consider that even
not setting the `IDRefreshName`, the test still goes on, which is not
the case.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -314,13 +314,12 @@ type TestCase struct {
 	// same state. Each step can have its own check to verify correctness.
 	Steps []TestStep
 
-	// The settings below control the "ID-only refresh test." This is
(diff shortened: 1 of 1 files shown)


appilon:
We have recently pushed V2's history to `master`, V2 is now the mainline of development, I am retargeting your PR to `v1-maint`. Apologies for any inconvenience, for more information on V2 of the SDK, please see our [discuss post](https://discuss.hashicorp.com/t/a-note-on-breaking-changes-in-the-plugin-sdk/7869)

paddycarver:
I'm going to close this out, as the test _should_ be conducted even if the values are not set. If this is not happening, please open a bug with information on how to reproduce, as what is documented is the intended behavior.

PR #294, opened 2020-01-14
validation: add StringIsEmpty, StringIsWhitespace, StringIsNotEmpty, StringIsNotWhitespace, StringIsBase64, UUID, UUIDOrEmpty

katbyte:


helper/validation/strings.go
```diff
@@ -0,0 +1,37 @@
+package validation
+
+import (
+	"encoding/base64"
+	"fmt"
+	"strings"
+)
+
+// StringIsNotEmpty is a ValidateFunc that ensures a string is not empty or consisting entirely of whitespace characters
+func StringIsNotEmpty(i interface{}, k string) ([]string, []error) {
```
appilon: This is already covered (sort of) by `NoZeroValues`, with the distinction that the type can be a string, or int, or float. I wouldn't be bothered if we accepted this too though as I would like to see the opposite added `StringIsEmpty`

helper/validation/uuid.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/hashicorp/go-uuid"
+)
+
+// UUIDRegExp is a Regular Expression that can be used to validate UUIDs
+var UUIDRegExp = regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
```
appilon: nice 👍

helper/validation/uuid.go
```diff
@@ -0,0 +1,41 @@
...
+
+	if _, err := uuid.ParseUUID(v); err != nil {
+		errors = append(errors, fmt.Errorf("expected %q to be a valid UUID, got %v", k, v))
+	}
+
+	return warnings, errors
+}
+
+// UUID is a ValidateFunc that ensures a string is empty or can be parsed as UUID
+func UUIDOrEmpty(i interface{}, k string) (warnings []string, errors []error) {
```
appilon: I would prefer encourage composition with `validation.Any`.

```
validation.Any(validation.UUID, validation.StringIsEmpty)
```
I mentioned earlier it might be nice to add `StringIsEmpty` as well

helper/validation/strings.go
katbyte: afaik `NoZeroValues` will not trigger on `property = "        "`

appilon: You have the added Trim rule too which is good, we COULD probably add that to `NoZeroValues` :man-shrugging: . I think I like this method if we add the `StringIsEmpty` complement

helper/validation/uuid.go
katbyte: ooo i like that

helper/validation/strings.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if v != "" {
+		return nil, []error{fmt.Errorf("expected %q to be an empty string: got %v", k, v)}
+	}
+
+	return nil, nil
+}
+
+// StringIsEmpty is a ValidateFunc that ensures a string is composed of entirly Wh
```
appilon:
```suggestion
// StringIsEmpty is a ValidateFunc that ensures a string is composed of entirely whitespace
```

PR #297, opened 2020-01-15
helper/validation: UUID -> IsUUID

radeksimko:
Just a (nitpicky) patch to align the function name with others and clarify the purpose of the function a bit.

I would be also 👌  with `IsUUID` or `ValidUUID`, but the function name should IMO communicate that it's positive validation.

### DIFF
--- a/helper/validation/uuid.go
+++ b/helper/validation/uuid.go
@@ -2,16 +2,12 @@ package validation
 
 import (
 	"fmt"
-	"regexp"
 
 	"github.com/hashicorp/go-uuid"
 )
 
-// uuidRegExp is a Regular Expression that can be used to validate UUIDs
-var uuidRegExp = regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
-
-// IsValidUUID is a ValidateFunc that ensures a string can be parsed as UUID
-func IsValidUUID(i interface{}, k string) (warnings []string, errors []error) {
+// IsUUID is a ValidateFunc that ensures a string can be parsed as UUID
+func IsUUID(i interface{}, k string) (warnings []string, errors []error) {
 	v, ok := i.(string)
 	if !ok {
 		errors = append(errors, fmt.Errorf("expected type of %q to be string", k))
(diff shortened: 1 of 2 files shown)


katbyte:
I personally would say `IsUUID` is ideal as the package name already conveys "validating" and we would end up with `valid` in there twice: `validation.IsUUID` vs `validation.IsValidUUID`

radeksimko:
No problem - I will change it @katbyte

helper/validation/uuid.go
```diff
@@ -7,11 +7,11 @@ import (
 	"github.com/hashicorp/go-uuid"
 )
 
-// UUIDRegExp is a Regular Expression that can be used to validate UUIDs
-var UUIDRegExp = regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
+// uuidRegExp is a Regular Expression that can be used to validate UUIDs
+var uuidRegExp = regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
```
appilon: I think this was intentionally exported in case a developer ever wanted a regex pattern for UUID, even though we don't use it internally

radeksimko: Well, in my mind it's an implementation detail in the context of the `validation` package which is supposed to provide validation _functions_ and anything exported is a liability for potential stability, so that's why I'm thinking that this gives us some headroom to move this around more easily without having to break anything/anyone.

appilon: Ah, I should have made it more clear, that this regex actually ISN'T used in the implementation (the related function just uses go's native parser). I believe Katie just threw in the Regex if it provided value to developers looking to use the regex. I gave her the unofficial "yeah why not throw it in" so that is my bad. We might want to just delete it for now and publish it in a future release should there be demand for it (also maybe it doesn't belong in validation 🤔 )

radeksimko: > also maybe it doesn't belong in validation

That's my thinking. It would be very difficult to maintain any stability guarantees, because it's essentially a constant, so every time you change it you break it.

appilon: Okay, let's just delete it for now and bring it up again if it's needed later. cc @katbyte

PR #300, opened 2020-01-17
helper/acctest: fix RandIntRange

alkar:
As per the original ticket (#171).


appilon:
Hello @alkar thank you for your contribution. As mentioned in other issues, strictly speaking this is a subtle "breaking change". As such it will be merged into our V2 branch. Sorry about the delay..

Issue #277, opened 2019-12-12
Split out RPC types duplicated between here and terraform into common module


LorbusChris:
### SDK version
```
{
  "Path": "github.com/hashicorp/terraform-plugin-sdk",
  "Version": "v1.4.0"
}
```

### Description of Problem
Importing and using the `terraform-plugin-sdk` into a project alongside `terraform` causes panics in `gob` and `proto` when the RPC types that are duplicated between those two repos are attempted to be initialized for a second time (namely `tfdiags.rpcFriendlyDiag` and all proto-generated types from `tfplugin5`):
```
panic: gob: registering duplicate types for "*tfdiags.rpcFriendlyDiag": *tfdiags.rpcFriendlyDiag != *tfdiags.rpcFriendlyDiag
goroutine 1 [running]:
encoding/gob.RegisterName(0x84498da, 0x18, 0x94d9180, 0x0)
	/usr/local/go/src/encoding/gob/type.go:820 +0x55e
encoding/gob.Register(0x94d9180, 0x0)
	/usr/local/go/src/encoding/gob/type.go:874 +0x125
github.com/hashicorp/terraform-plugin-sdk/internal/tfdiags.init.0()
	/go/src/github.com/openshift/installer/vendor/github.com/hashicorp/terraform-plugin-sdk/internal/tfdiags/rpc_friendly.go:58 +0x36 
```
```
panic: proto: duplicate enum registered: tfplugin5.Diagnostic_Severity

goroutine 1 [running]:
github.com/golang/protobuf/proto.RegisterEnum(...)
        /go/pkg/mod/github.com/golang/protobuf@v1.3.2/proto/properties.go:458
github.com/hashicorp/terraform-plugin-sdk/internal/tfplugin5.init.0()
```
While I've been made aware that this is generally not a supported use-case, for us over at [OpenShift Installer](https://github.com/openshift/installer) this is a regression from previous behaviour after migrating to the sdk, and makes it impossible for us to import `terraform-plugin-sdk` there as-is.

I do however see ~~two~~ one possible option for a solution here:
1. One would be the deduplication of those types between here and the `terraform` repo, which is probably not a thing that can be done in the short-term.

[...]

PR #308, opened 2020-01-29
Make tfdiags and tfplugin5 public

LorbusChris:
This is an attempt at solving https://github.com/hashicorp/terraform/issues/23725 (via Proposal 2a from that issue)
and a follow-up to https://github.com/hashicorp/terraform/pull/23894
and https://github.com/hashicorp/terraform-plugin-sdk/issues/277

If the canonical location of these types were this repo, they could be imported into the main `terraform` repo from here, too. (as a follow up, Proposal 2b from https://github.com/hashicorp/terraform/issues/23725)

/cc @apparentlymart 

I do think splitting out the shared types into a shared common package is still most desirable (Proposal 1 from https://github.com/hashicorp/terraform/issues/23725), but it would require maintainers from this org to take initiative on that.


LorbusChris:
Ping. Maybe this should include _all_ the types currently duplicated between here and terraform?

paultyng:
We are moving further away from shared type representations between the SDK with core, as the purposes and use cases of the code diverge (SDK for developer facing consumption of the wire protocol, core is an implementation of the HCL language), so we are starting to fork things like [cty](https://github.com/hashicorp/go-cty), diags, paths, etc. We should have a cleaner wrapper of the wire protocol coming soon as well as the ability to "mux" individual resources to SDK/protocol implementation versions. As such, I don't think this PR's intent makes much sense in the context of our longer term goals, going to close. Regarding canonical types for the protobuf though, its best to regenerate your own implementations, protobuf compatibility is not Go compatibility, and the versioning would be very difficult to maintain in Go module's versioning scheme.

PR #322, opened 2020-02-13
Context aware StateChangeConf

### LINKED ISSUE #73: Implement graceful shutdown for helper/resource waiter
For many resources it's currently impossible for Terraform to gracefully stop the execution mid-flight as `helper/resource.StateChangeConf` is not cancellable from above the resource/provider.

Mitchell has done the initial work in https://github.com/hashicorp/terraform/pull/9536 (also includes changes in Azure provider + 1 resource). The goal is to make similar changes to (most) other providers - at least the ones we know leverage `helper/resource.StateChangeConf` and pass the `StopContext` down to the `resource.StateChangeConf` in the resource code.

### Terraform Version

`v0.8.5-dev (aa3eda76425a0b936192c6e95e45758e2727ba4b)`

### Affected Resource(s)

 - any resource using `resource.StateChangeConf`, but mainly ones that have long timeouts, e.g. AWS RDS, AWS Elasticsearch Domain

```hcl
resource "aws_rds_cluster" "default" {
  cluster_identifier = "aurora-cluster-demo"
  availability_zones = ["eu-west-2a","eu-west-2b"]
  database_name = "mydb"
  master_username = "foo"
  master_password = "barbarbar"
  backup_retention_period = 5
  preferred_backup_window = "07:00-09:00"
}
```

### Expected Behavior

```
aws_rds_cluster.default: Creating...
...
aws_rds_cluster.default: Still creating... (10s elapsed)
^CInterrupt received. Gracefully shutting down...
Apply failed. Interruption received, partial state saved.
```

### Actual Behavior

```
aws_rds_cluster.default: Creating...
...
aws_rds_cluster.default: Still creating... (10s elapsed)
^CInterrupt received. Gracefully shutting down...
aws_rds_cluster.default: Still creating... (20s elapsed)
aws_rds_cluster.default: Still creating... (30s elapsed)
aws_rds_cluster.default: Still creating... (40s elapsed)
aws_rds_cluster.default: Still creating... (50s elapsed)
aws_rds_cluster.default: Creation complete

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

### Steps to Reproduce

1. `terraform apply`
2. `Ctrl+C` (once)

### References

- https://github.com/hashicorp/terraform/pull/9536

appilon:
Creates a context aware `resource.RetryContext` and `resource.StateChangeConf.WaitForStateContext`.

Cancellation/deadline of the passed in context is respected. If `WaitForStateContext` returns due to context cancellation/deadline the resulting `context.DeadlineExceeded` or `context.Canceled` error is returned but the LastResult is not returned. If the passed in timeout is exceeded and we are in the graceful shutdown period, context cancelation will exit the grace period early, however the returned error will still be of type `*resource.TimeoutError`.

Context was not pulled through to the `Refresh` functions, these functions are often (if not always?) closed over within a CRUD method and if the context is needed within that scope it is in scope (much like `meta`).

### Proposed Usage
```
func create(ctx context.Context, d *schema.ResourceData, meta interface{}) error {
    // StateChangeConf
    conf := &StateChangeConf{
		Pending:      []string{"pending", "incomplete"},
		Target:       []string{"running"},
		Refresh:      func() (interface{}, string, error) {
		    // ...
	        },
		Timeout:      10 * time.Second,
    }
    state, err := conf.WaitForStateContext(ctx)

    // Retry
    resource.RetryContext(ctx, 10 * time.Second, func() *RetryError {
        // ...
    })
}
```

Closes #73

### DIFF
--- a/helper/resource/state.go
+++ b/helper/resource/state.go
@@ -94,8 +94,6 @@ func (conf *StateChangeConf) WaitForStateContext(ctx context.Context) (interface
 
 			// wait and watch for cancellation
 			select {
-			case <-ctx.Done():
-				return
 			case <-cancelCh:
 				return
 			case <-time.After(wait):
@@ -218,7 +216,9 @@ func (conf *StateChangeConf) WaitForStateContext(ctx context.Context) (interface
 
 			// still waiting, store the last result
 			lastResult = r
-
+		case <-ctx.Done():
+			// this is a simple approach but swallows lastResult.Error
+			return lastResult.Result, ctx.Err()
 		case <-timeout:
 			log.Printf("[WARN] WaitForState timeout after %s", conf.Timeout)
 			log.Printf("[WARN] WaitForState starting %s refresh grace period", refreshGracePeriod)
@@ -247,6 +247,9 @@ func (conf *StateChangeConf) WaitForStateContext(ctx context.Context) (interface
 					// target state not reached, save the result for the
 					// TimeoutError and wait for the channel to close
 					lastResult = r
+				case <-ctx.Done():
+					log.Println("[WARN] Context cancelation detected, abandoning grace period")
+					break forSelect
 				case <-timeout:
 					log.Println("[ERROR] WaitForState exceeded refresh grace period")
 					break forSelect
--- a/helper/resource/state.go
+++ b/helper/resource/state.go
@@ -83,7 +83,11 @@ func (conf *StateChangeConf) WaitForStateContext(ctx context.Context) (interface
 	go func() {
 		defer close(resCh)
 
-		time.Sleep(conf.Delay)
+		select {
+		case <-time.After(conf.Delay):
+		case <-cancelCh:
+			return
+		}
 
 		// start with 0 delay for the first loop
 		var wait time.Duration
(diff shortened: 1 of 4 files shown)


helper/resource/state.go
```diff
@@ -213,7 +216,9 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 
 			// still waiting, store the last result
 			lastResult = r
-
+		case <-ctx.Done():
+			// this is a simple approach but swallows lastResult.Error
+			return lastResult.Result, ctx.Err()
```
paultyng: considering `ctx.Err` is non-nil here, result should be `nil`

helper/resource/state.go
```diff
@@ -242,6 +247,9 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 					// target state not reached, save the result for the
 					// TimeoutError and wait for the channel to close
 					lastResult = r
+				case <-ctx.Done():
+					log.Println("[WARN] Context cancelation detected, abandoning grace period")
```
paultyng: I think this should be an `[ERROR]` not a `[WARN]` here as its going to return a `TimeoutError`

helper/resource/state.go
appilon: Yeah wasn't sure if we should return what we got at the very least? Would it not be beneficial to give the developer options? Not a strong opinion happy to keep it simple with `nil` if that's a better practice.

PR #325, opened 2020-02-17
schema: discard function fields for json marshalling

sylr:
Useful for debugging.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -111,7 +111,7 @@ type Schema struct {
 	// This allows comparison based on something other than primitive, list
 	// or map equality - for example SSH public keys may be considered
 	// equivalent regardless of trailing whitespace.
-	DiffSuppressFunc SchemaDiffSuppressFunc
+	DiffSuppressFunc SchemaDiffSuppressFunc `json:"-"`
 
 	// If this is non-nil, then this will be a default value that is used
 	// when this item is not set in the configuration.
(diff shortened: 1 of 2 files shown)


paultyng:
These types are not meant for marshaling, if you want to marshal in your own codebase you can copy the types to a type special cased for that purpose. The method we use currently for extracting schema to JSON is to execute the [`terraform providers schema` command](https://www.terraform.io/docs/commands/providers/schema.html).

sylr:
Like I said, it was for debugging:

```go
	j, err := json.MarshalIndent(result, "", "  ")
	log.Printf("[DEBUG] before sortSecurityRules sorting: %v (err: %v)", string(j), err)
	sortSecurityRules(result)
	j2, err := json.MarshalIndent(result, "", "  ")
	log.Printf("[DEBUG] after sortSecurityRules sorting: %v (err: %v)", string(j2), err)
```

paultyng:
I know what you are getting at, but dynamic schema itself is problematic for Terraform, if you are generating your schema to the point you need to debug it, you are probably going to run in to additional problems down the line (state migrations, etc). 

If you really want debugging of schema in your code, its more tedious, but you could wrap it in an anonymous struct to override like:

```go
	x := (struct {
		Type schema.ValueType
		ConfigMode schema.SchemaConfigMode
		Optional bool
		Required bool
		DiffSuppressFunc schema.SchemaDiffSuppressFunc `json:"-"`
		Default     interface{}
		DefaultFunc schema.SchemaDefaultFunc `json:"-"`
		Description string
		InputDefault string
		Computed  bool
		ForceNew  bool
		StateFunc schema.SchemaStateFunc `json:"-"`
		Elem interface{}
		MaxItems int
		MinItems int
		PromoteSingle bool
		Set schema.SchemaSetFunc `json:"-"`
		ComputedWhen []string
		ConflictsWith []string
		ExactlyOneOf  []string
		AtLeastOneOf  []string
		Deprecated string
		Removed string
		ValidateFunc schema.SchemaValidateFunc `json:"-"`
		Sensitive bool
	})(result)
```

sylr:
Well, that's the thing, I'm not debugging my schema but someone else's.

PR #327, opened 2020-02-18
Mandatory Binary Driver

appilon:
V2 of the SDK removes the internalized version of terraform core. Acceptance tests can now only be ran via the binary driver. Test cases can no longer opt out.


appilon:
@kmoe Its all gone already! Have a look at the version2 branch, let me know if i missed something

PR #332, opened 2020-02-20
Go Module V2

appilon:
After this is merged, we can start tagging this branch with alphas/betas/release candidates.


helper/schema/field_reader_diff_test.go
```diff
@@ -109,7 +109,7 @@ func TestDiffFieldReader_NestedSetUpdate(t *testing.T) {
 	}
 }
 
-// https://github.com/hashicorp/terraform-plugin-sdk/issues/914
+// https://github.com/hashicorp/terraform-plugin-sdk/v2/issues/914
```
paultyng: this is wrong

helper/schema/schema_test.go
```diff
@@ -1949,7 +1949,7 @@ func TestSchemaMap_Diff(t *testing.T) {
 		},
 
 		{
-			Name: "https://github.com/hashicorp/terraform-plugin-sdk/issues/824",
+			Name: "https://github.com/hashicorp/terraform-plugin-sdk/v2/issues/824",
```
paultyng: shouldn't be updated

helper/schema/shims_test.go
```diff
@@ -2234,7 +2234,7 @@ func TestShimSchemaMap_Diff(t *testing.T) {
 		},
 
 		{
-			Name: "https://github.com/hashicorp/terraform-plugin-sdk/issues/824",
+			Name: "https://github.com/hashicorp/terraform-plugin-sdk/v2/issues/824",
```
paultyng: shouldn't be updated

.go-version
```diff
@@ -1 +1 @@
-1.12.9
+1.12.17
```
paddycarver: Given we're on Go 1.14 now, should we bump the version accordingly?

appilon: I think in general I held off go1.13 personally (for all my development) for some weird transitive dependency things... I can't remember them at all so sure let's move to go1.14!

PR #333, opened 2020-02-20
Remove deprecated validation funcs

appilon:


internal/addrs/self.go
```diff
@@ -1,14 +0,0 @@
-package addrs
```
paddycarver: Why is this file getting removed in this PR?

appilon: I just wanted to sneak it in... its dead code, I know not the best habit, it somehow was missed in my formal cleanup PR.

PR #335, opened 2020-02-21
helper/resource: Return error with TestCheckResourceAttrPair() when comparing self

### LINKED ISSUE #334: helper/resource.TestCheckResourceAttrPair() Does Not Return Error If Comparing Self
### SDK version

```
v1.7.0
```

```go
// Usually less obvious with variable usage
resource.TestCheckResourceAttrPair("example_thing.test", "attr", "example_thing.test", "attr"),
```

N/A -- can be covered via unit testing

### Expected Behavior

The acceptance testing function should return an error saying the same resource state values are being checked against each other, which will always pass. It is more likely the developer intended to check either different attributes in the same resource (seems like an odd case, but should be still allowed) or different resources (most likely; e.g. checking data source attribute versus resource attribute).

### Actual Behavior

No errors returned by `TestCheckResourceAttrPair()` and acceptance testing passes.

### Steps to Reproduce

```
TF_ACC=1 go test ./... -count=1 -run='ExampleThing_' -v
```

### References

- https://github.com/terraform-providers/terraform-provider-aws/pull/10932#discussion_r382574335

bflad:
Closes #334

Also consolidates the unit testing for the function.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -873,6 +873,14 @@ func TestCheckModuleResourceAttrPair(mpFirst []string, nameFirst string, keyFirs
 }
 
 func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst string, keyFirst string, isSecond *terraform.InstanceState, nameSecond string, keySecond string) error {
+	if nameFirst == nameSecond && keyFirst == keySecond {
+		return fmt.Errorf(
+			"comparing self: resource %s attribute %s",
+			nameFirst,
+			keyFirst,
+		)
+	}
+
 	vFirst, okFirst := isFirst.Attributes[keyFirst]
 	vSecond, okSecond := isSecond.Attributes[keySecond]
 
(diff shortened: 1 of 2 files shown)


helper/resource/testing.go
```diff
@@ -1232,6 +1232,14 @@ func TestCheckModuleResourceAttrPair(mpFirst []string, nameFirst string, keyFirs
 }
 
 func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst string, keyFirst string, isSecond *terraform.InstanceState, nameSecond string, keySecond string) error {
+	if nameFirst == nameSecond && keyFirst == keySecond {
```
appilon: While I like the enhancement, I'm inclined to think in V1 maybe this should just be a warning like "you may be comparing an attribute with itself". We can make it an explicit error in V2.... tough one kind of a grey area on if this is a "breaking change" or a bugfix. I could be convinced either way.

appilon: But yeah, would the warning ever get noticed in review/during development vs explicit error, it certainly is valuable in avoiding accidental merging of true == true when a PR is showing all tests passing.

appilon:
@bflad I am going to take the liberty of rebasing with version2 and merging it there as its safer.

PR #344, opened 2020-03-06
TF_SCHEMA_PANIC_ON_ERROR on by default

appilon:
opt out with Go parsable boolean (False, false, FALSE, 0). In the case where a value is set but is not a parsable boolean, a warning is logged but true is the default to remain backwards compatible with the previous implementation (where opt in involved setting the env var to anything).


appilon:
@paddycarver We never had this environment variable documented... but I will make sure its documented in the 2.0.0 Changelog!

PR #350, opened 2020-03-11
Deprecate GetOkExists

### LINKED ISSUE #283: docs for GetOkExists are misleading
### Terraform Version
```
$ terraform -v
Terraform v0.12.18
```

n/a

### Debug Output
n/a

### Crash Output
n/a

### Expected Behavior
Documentation matches code behavior

### Actual Behavior
The documentation for `GetOkExists` in the `github.com/hashicorp/terraform/helper/schema` package starts the same as the documentation for `GetOk`:

https://github.com/hashicorp/terraform/blob/f1237f816c5d00c59648d3170c7b8f4afe6d982d/helper/schema/resource_data.go#L111-L125

Notably:
> GetOkExists returns the data for a given key and whether or not the key has been set to a non-zero value.

However, the implementation for `GetOkExists` (and the rest of its line comments!) indicate that it does _not_ check if the key has been set to a non-zero value, and that indeed that's why it exists as a function in the first place!

### Steps to Reproduce
n/a

### Additional Context

I ran in to this debugging an issue in a plugin where I was unable to set a value of `0` https://github.com/kevholditch/terraform-provider-kong/issues/93. It turned out that the field was being read with `GetOk` and thus, when it was `0`, it was being turned by the plugin in to `nil` as if it hadn't been set.

When digging in the code, I was somewhat confused until I realized the line comments were not in line with the code.

I didn't just open this as a PR because my research also indicated that `GetOk` and `GetOkExists` have a somewhat complicated history, and it's possible there's some subtlety going on here that I do not understand.

### References
https://github.com/hashicorp/terraform/pull/15723
https://github.com/hashicorp/terraform/issues/4936#issuecomment-197908792

appilon:
Taking a harder stance and really discouraging the use of this method. It will be deprecated but I would not remove it in V2.

closes #283

### DIFF
--- a/helper/schema/resource_data.go
+++ b/helper/schema/resource_data.go
@@ -108,16 +108,11 @@ func (d *ResourceData) GetOk(key string) (interface{}, bool) {
 	return r.Value, exists
 }
 
-// GetOkExists returns the data for a given key and whether or not the key
-// has been set to a non-zero value. This is only useful for determining
-// if boolean attributes have been set, if they are Optional but do not
-// have a Default value.
+// GetOkExists can check if TypeBool attributes that are Optional with
+// no Default value have been set.
 //
-// This is nearly the same function as GetOk, yet it does not check
-// for the zero value of the attribute's type. This allows for attributes
-// without a default, to fully check for a literal assignment, regardless
-// of the zero-value for that type.
-// This should only be used if absolutely required/needed.
+// Deprecated: usage is discouraged due to undefined behaviors and may be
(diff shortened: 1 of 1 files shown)


mingfang:
I need GetOkExists. What is the replacement once you remove it?

appilon:
We won't remove this without a suitable replacement. It's deprecated to turn people away from it as much as possible. If version 2 of the SDK were released today this method would still exist since we have not come up with a replacement.

PR #351, opened 2020-03-11
Enable GRPC reflection

appilon:
Updates to go-plugin 1.2.0 which enables grpc reflection


paultyng:
If hashicorp/go-plugin#137 lands first, we can just pull that in instead. We can slow this down in the meantime to give the conversation there some time.

paultyng:
I submitted an upstream PR: hashicorp/go-plugin#138

appilon:
@paultyng rewrote PR with go-plugin@1.2.0

PR #353, opened 2020-03-12
Update to tfplugin 5.2, map new protocol fields (v1)

paultyng:
This is the v1 backport of #349


paultyng:
cc @appilon @paddycarver , the v1 version of that other PR, not a high priority to get it in to v1, but would be useful for docs/lang server potentially.

PR #355, opened 2020-03-13
auto-configure providers during binary acctests

### LINKED ISSUE #352: Binary Test Driver: Unconfigured Provider Global
Some providers such as the DNS provider keep a global reference to the `*schema.Provider` for passing along `provider.Meta()` in test checks [example](https://github.com/terraform-providers/terraform-provider-dns/blob/master/dns/resource_dns_a_record_set_test.go#L107).

The issue is that when using the binary test driver, this instance of the provider has never had `ConfigureFunc` called, thus has no client/meta object configured, since the actual plugin being tested is running in a separate process. `provider.Meta()` returns nil resulting in a type assertion panic.

Globals in general might be anti-pattern, however this might be a widespread pattern we need to find a way to support (should investigate, perhaps it's only a DNS provider thing). I think it's possible we might just need to provide the developer a means to get a configured client (meta).

UPDATE:
Seems like we have precedent for configuring the global instance in a `PreCheck` function, see [aws](https://github.com/terraform-providers/terraform-provider-aws/blob/0286dba6e6f55c52863bdfa6662282a250a32a74/aws/provider_test.go#L67) this should do the trick

kmoe:
Fixes #352.

This runs `Configure()` on providers prior to the binary acctest run, so they can access `provider.Meta()` during tests.

The one nuance to be aware of here is that `PreCheck()` is called after the auto-configure, in case providers are already using this for custom configuration.

### Note

Once this is approved I'll generate CHANGELOG and raise another PR to the `version2` branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ DEPRECATED:
 
 * helper/schema: `ResourceData.GetOkExists` will not be removed in the next major version unless a suitable replacement or alternative can be prescribed [GH-350]
 
+BUG FIXES:
+
+* Binary acceptance test driver: auto-configure providers [GH-355]
+
 # 1.8.0 (March 11, 2020)
 
 FEATURES:
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -556,11 +556,6 @@ func Test(t TestT, c TestCase) {
 		return
 	}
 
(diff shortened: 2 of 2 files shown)


kmoe:
I'll just raise another PR to `version2` manually - no need for git magic here as it's a simple change.

PR #357, opened 2020-03-16
v2: binary acctesting - auto-Configure providers

kmoe:
This is a port of https://github.com/hashicorp/terraform-plugin-sdk/pull/355 to v2.

It also moves `helper/schema/diagnostic_test.go` to its own package, `internal/diagutils`.


kmoe:
This needs a bit more thought, as with `ConfigureContextFunc` we now have to pass `ctx` during the auto-configure.

`ConfigureFunc` is deprecated but things will work normally in `helper/schema/provider.go` if `ConfigureFunc` is set and `ConfigureContextFunc` is not. Should we switch in `testing.go` as well and auto-configure the provider even if only the deprecated `ConfigureFunc` is set, or use acceptance tests to force the new implementation? Maybe a warning?

helper/schema/provider_test.go
```diff
@@ -180,7 +181,7 @@ func TestProviderConfigure(t *testing.T) {
 		c := terraform.NewResourceConfigRaw(tc.Config)
 		diags := tc.P.Configure(context.Background(), c)
 		if diags.HasError() != tc.Err {
-			t.Fatalf("%d: %s", i, errorDiags(diags))
+			t.Fatalf("%d: %s", i, diagutils.ErrorDiags(diags))
```
appilon, line 184: Please delete `diagnostic_test.go` I think you will need to port a few lines to `diagutils.Warnings`

appilon: Oh you did move the file sorry.... weird is `warnDiags` not used at all?

helper/schema/schema_test.go
```diff
@@ -5197,19 +5198,19 @@ func TestSchemaMap_Validate(t *testing.T) {
...
-				for _, e := range errorDiags(diags).Errors() {
+				for _, e := range diagutils.ErrorDiags(diags).Errors() {
 					t.Errorf("%q: err: %s", tn, e)
 				}
 
 				t.FailNow()
 			}
 
-			ws := warningDiags(diags).Warnings()
+			ws := diagutils.WarningDiags(diags).Warnings()
```
appilon, line 5208: ah my eyes are slow today, ignore my other comments..

PR #359, opened 2020-03-16
Type Block need handle attribute marked as removed contionally.

magodo:
In some provider, e.g. AzureRM, TypeSet is use quite prevalent to
hold a set of sub-resources (e.g. security_rule in azurerm_network_security_group).
The reason for using TypeSet instead of TypeList might because service
return a collection of info with arbitrary order, or other reasons.

In this case: Given a slice of nested blocks, if one of the nested block (B1)
attributes is optional and has no default value, and user didn't specify a value
for it. Then if another nested block (B2) changed, which triggers some diff,
then B1 will also be replaced. That is because the optional attribute
contribute a diff of "zero value" -> `null`, which changed the hash of
the block.

This fix is to carefully handle this case. We keep attribute marked
as `NewRemoved` after a `diffString` only when all the attributes are
marked as such. Otherwise, as long as one attribute is not marked to be
removed, that means this block will be kept. Then we will manipulate the
attributes in this block, which being marked as removed just because
user didn't specify a value and the original value is the zero value.

This keeps consistent as other Resource types (e.g. List Resource).
(Though those type just remove the attribute from diff set, while for
Set we need to return the complete state as we will not depend on the
old state during diff apply).


magodo:
This should fix terraform-providers/terraform-provider-azurerm#5819.

appilon:
We have recently pushed V2's history to `master`, V2 is now the mainline of development, I am retargeting your PR to `v1-maint`. Apologies for any inconvenience, for more information on V2 of the SDK, please see our [discuss post](https://discuss.hashicorp.com/t/a-note-on-breaking-changes-in-the-plugin-sdk/7869)

PR #362, opened 2020-03-17
Fix HasChange when using nested schema.Set

### LINKED ISSUE #361: HasChange wrongly returns TRUE on nested shema.Set structures
### SDK version

```
v1.7.0
```

When investigating [Terraform AWS provider issue #11059](https://github.com/terraform-providers/terraform-provider-aws/issues/11059) I came across some cases where nested schema.Set structures as the one [mentioned in the 2nd part of this entry](https://github.com/terraform-providers/terraform-provider-aws/issues/11059#issuecomment-599631533) produce unnecessary calls later down in the code

```go
...
```

```hcl
...
```

### Expected Behavior

HasChange should return false as entries are perfectly equivalent

### Actual Behavior

[This test](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_security_group.go#L695) returns true because
of reflect.DeepEqual which returns false when traversing Func(s)

### Steps to Reproduce

See aforementioned issue

### References

- [Terraform AWS provider issue #11059](https://github.com/terraform-providers/terraform-provider-aws/issues/11059)
- [The code I have implemented which prevents this](https://github.com/hashicorp/terraform-plugin-sdk/compare/master...obourdon:master)

obourdon:
This fixes #361

### DIFF
--- a/helper/schema/set.go
+++ b/helper/schema/set.go
@@ -150,13 +150,46 @@ func (s *Set) Union(other *Set) *Set {
 	return result
 }
 
+func checkSetMapEqual(m1, m2 map[string]interface{}) bool {
+	if (m1 == nil) != (m2 == nil) {
+		return false
+	}
+	if len(m1) != len(m2) {
+		return false
+	}
+	for k := range m1 {
+		v1 := m1[k]
+		v2, ok := m2[k]
+		if !ok {
+			return false
+		}
+		switch v1.(type) {
+		case map[string]interface{}:
+			same := checkSetMapEqual(v1.(map[string]interface{}), v2.(map[string]interface{}))
+			if !same {
+				return false
+			}
+		case *Set:
+			same := v1.(*Set).Equal(v2)
+			if !same {
+				return false
+			}
+		default:
+			same := reflect.DeepEqual(v1, v2)
+			if !same {
+				return false
+			}
+		}
+	}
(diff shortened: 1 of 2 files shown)


paultyng:
@obourdon this seems like a reasonable fix, do you mind adding some testing around the issue to demonstrate the problems (and prevent regression)?

Is it possible the AWS code should be modified to use `HashEqual` instead? Currently the way the SDK does most set handling is via the hash, this is changing a bit in 0.12 but will probably remain this way in this version of the SDK for the time being.

obourdon:
@paultyng many thanks for the comments above. Will definitely look into putting some tests and also have a look at your suggestion about AWS HashEqual. I just fixed the formatting as I did not see that the CI had failed. Many thanks again

paultyng:
The more I thought about this, I think this approach to `Equal` is your best bet, as `HasChange` has different mean than the set hashing which is mostly about matching up values for diffing. Will have someone take a peek, but we'll need some test coverage to merge.

obourdon:
@paultyng test has been added and I have verified that reverting commit 
48bd90d fails and succeed otherwise running:
```
go test ./... -run 'TestSet.*' -count=1
```

obourdon:
@appilon no pb at all, just let me know if you need me for anything.

Issue #71, opened 2019-09-26
Using ConflictsWith with Lists


jtopjian:
I'm trying to use `ConflictsWith` on a `TypeList` and it isn't working. I think I've narrowed down the issue, but I'm not entirely sure.

The `openstack_compute_instance_v2` resource has a [`network`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_compute_instance_v2.go#L107) attribute of      `TypeList`. The list defines several nested attributes that can all be used to create a network.

Some of the attributes can't be used together. For example, [`port`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_compute_instance_v2.go#L126) and         [`floating_ip`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_compute_instance_v2.go#L144). So I made the following changes:

```hcl
"port": &schema.Schema{
        Type:     schema.TypeString,
        Optional: true,
        ForceNew: true,
        Computed: true,
        ConflictsWith: []string{"network.floating_ip"},
},
"floating_ip": &schema.Schema{
        Type:     schema.TypeString,
        Optional: true,
        ForceNew: true,
        Computed: true,
        ConflictsWith: []string{"network.port"},
},
```

Then, given something simple like:

```hcl
resource "openstack_compute_instance_v2" "instance_1" {
  name = "instance_1"
  network {
    floating_ip = "foo"
    port = "bar"
  }
}
```

I would expect an error to be throw, however, one isn't. _But_, if I change the schema to:

```hcl
"port": &schema.Schema{
        Type:     schema.TypeString,
        Optional: true,
        ForceNew: true,
        Computed: true,
        ConflictsWith: []string{"network.0.floating_ip"},
},
"floating_ip": &schema.Schema{
        Type:     schema.TypeString,
        Optional: true,
        ForceNew: true,
        Computed: true,
        ConflictsWith: []string{"network.0.port"},
},
```

Then I get a validation error. Of course, that is only applicable to the first defined network.

Throwing some debug output into `helper/schema/schema.go` and `terraform/resource.go`, I _think_ the following is happening:

[...]

PR #364, opened 2020-03-18
tests/helper/schema: Add unit testing covering existing behavior of ConflictsWith and list index/map key syntax

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-sdk/issues/71
Reference: https://github.com/bflad/tfproviderlint/issues/104

The `TestSchemaMap_InternalValidate` tests also include TODOs for potential schema validation improvements.

### DIFF
--- a/helper/schema/schema_test.go
+++ b/helper/schema/schema_test.go
@@ -3595,6 +3595,244 @@ func TestSchemaMap_InternalValidate(t *testing.T) {
 			true,
 		},
 
+		"ConflictsWith list index syntax with self reference": {
+			map[string]*Schema{
+				"config_block_attr": &Schema{
+					Type:     TypeList,
+					Optional: true,
+					Elem: &Resource{
+						Schema: map[string]*Schema{
+							"nested_attr": &Schema{
+								Type:          TypeString,
+								Optional:      true,
+								ConflictsWith: []string{"config_block_attr.0.nested_attr"},
+							},
+						},
+					},
+				},
+			},
+			false, // TODO: This should return an error as it will always error during runtime
+		},
+
+		"ConflictsWith list index syntax with list configuration block existing attribute": {
+			map[string]*Schema{
+				"config_block_attr": &Schema{
+					Type:     TypeList,
+					Optional: true,
+					Elem: &Resource{
+						Schema: map[string]*Schema{
+							"nested_attr": &Schema{
+								Type:     TypeString,
+								Optional: true,
+							},
+						},
+					},
+				},
+				"test": &Schema{
+					Type:          TypeBool,
+					Optional:      true,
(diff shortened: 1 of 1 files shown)


helper/schema/schema_test.go
```diff
@@ -3595,6 +3595,244 @@ func TestSchemaMap_InternalValidate(t *testing.T) {
...
+							"nested_attr": &Schema{
+								Type:          TypeString,
+								Optional:      true,
+								ConflictsWith: []string{"config_block_attr.0.nested_attr"},
+							},
+						},
+					},
+				},
+			},
+			false, // TODO: This should return an error as it will always error during runtime
```
appilon, line 3614: Can you let me know more details on this TODO? If the test is supposed to error but currently doesn't doesn't that mean something is broken?

appilon:
We have recently pushed V2's history onto master as the new mainline of development, I am retargeting your PR to `v1-maint`.

helper/schema/schema_test.go
bflad: Hi @appilon 👋 Yes, that TODO is meant to signify something is broken in the sense that `InternalValidate()` does not fail when you self-reference in this manner, however the schema is invalid and will never work when the attribute is declared in a configuration.

Copying this configuration block schema into a real attribute of a real resource (I just randomly chose `aws_devicefarm_project`):

```go
func resourceAwsDevicefarmProject() *schema.Resource {
	return &schema.Resource{
		Create: resourceAwsDevicefarmProjectCreate,
		Read:   resourceAwsDevicefarmProjectRead,
		Update: resourceAwsDevicefarmProjectUpdate,
		Delete: resourceAwsDevicefarmProjectDelete,

			"arn": {
				Type:     schema.TypeString,
				Computed: true,
			},

			"name": {
				Type:     schema.TypeString,
				Required: true,
			},

			"config_block_attr": {
				Type:     schema.TypeList,
				Optional: true,
				Elem: &schema.Resource{
						"nested_attr": {
							Type:          schema.TypeString,
							Optional:      true,
							ConflictsWith: []string{"config_block_attr.0.nested_attr"},
						},
					},
				},
			},
		},
	}
}
```

And adding the configuration to the test configuration:

```go
func testAccDeviceFarmProjectConfig(rInt int) string {
	return fmt.Sprintf(`
resource "aws_devicefarm_project" "foo" {
  name = "tf-testproject-%d"

  config_block_attr {
    nested_attr = "test"
  }
}
`, rInt)
}
```

Will always return a runtime error due to the self-reference:

```
=== CONT  TestAccAWSDeviceFarmProject_basic
    TestAccAWSDeviceFarmProject_basic: testing.go:669: Step 0 error: config is invalid: "config_block_attr.0.nested_attr": conflicts with config_block_attr.0.nested_attr
```

PR #368, opened 2020-03-20
Diagnostics Proposal

appilon:
This PR implements diagnostics for validation, CRUD, and ConfigureContextFunc. There are breaking changes to some functions along the callstack but they are functions mostly considered "internal" despite being publicly exported.

## Usage
```
  "bar": {
    Type: schema.TypeMap,
    Required: true,
    ValidateDiagFunc: func(v interface{}, path cty.Path) diag.Diagnostics {
      // please build up a slice of diagnostics
      // the new signature passes you the cty.Path up to the attribute being validated
      // the SDK will automatically set the attribute path of all diagnostics to at least this point
      // therefore for primitive types, developers can leave the AttributePath unset
      // in the case of TypeMap, to have precise diagnostics, please provider either a relative path
      // or an absolute path to the key that has a diagnostic. 
      var diags diag.Diagnostics

      // set a relative path
      diags = append(diags, diag.Diagnostic{
        Severity: diag.Error,
        Summary: "error in map key",
        AttributePath: cty.Path{ cty.IndexStep{ Key: cty.StringVal("key_name") } },
      })

      // create the absolute path yourself with the passed in path
      diags = append(diags, diag.Diagnostic{
        Severity: diag.Error,
        Summary: "error in map key",
        AttributePath: append(path, cty.IndexStep{ Key: cty.StringVal("key_name")),
      })

      return diags
    }
  },
}
```
For diagnostics support in the other areas of the SDK (such as the context aware CRUD) the usage is similar except the developer is responsible for constructing and setting the path entirely (validation is unique in that the SDK drills down the schema)

### TypeSet
Unfortunately `schema.TypeSet` cannot be represented as an attribute path with the current protocol. In the case of validation the path is truncated at that attribute, which will provide some context to the user but it is not perfect, err on the side of verbose wording for diagnostics that are a child of `TypeSet`. 

We are still investigating having guardrails around this in the CRUD API so developers don't create invalid paths in general (relative to their schema definition). The situation isn't so bad though as Terraform will just render blank where the line and file info would have been (not great but also nonfatal)


diag/diagnostic.go
```diff
@@ -0,0 +1,87 @@
...
+		if d.Severity == Error {
+			return true
+		}
+	}
+	return false
+}
+
+func (diags Diagnostics) Error() string {
+	return multierror.ListFormatFunc(diags.Errors())
+}
```
paultyng: I really don't think this should implement the `Error` interface, people will do bad things with it.

diag/diagnostic.go
```diff
@@ -0,0 +1,89 @@
...
+func (diags Diagnostics) HasErrors() bool {
+	for _, d := range diags {
+		if d.Severity == Error {
+			return true
+		}
+	}
+	return false
+}
+
+func (diags Diagnostics) Errors() []error {
```
paultyng, line 47: Can remove this as well if you remove the `Err` method, only used in tests.

diag/diagnostic.go
```diff
@@ -0,0 +1,89 @@
...
+	var errs []error
+	for _, d := range diags {
+		if d.Severity == Error {
+			errs = append(errs, error(d))
+		}
+	}
+	return errs
+}
+
+func (diags Diagnostics) Warnings() []string {
```
paultyng, line 57: Only used in tests, can remove.

diag/diagnostic.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+	return diags
+}
+
+func (diags Diagnostics) Err() error {
+	return errors.New(multierror.ListFormatFunc(diags.Errors()))
+}
+
+func (diags Diagnostics) HasErrors() bool {
```
paultyng, line 38: This should be `HasError` its only checking for at least one.

helper/schema/diagnostic.go
```diff
@@ -0,0 +1,45 @@
+package schema
```
paultyng, line 1: can you rename this file `diagnostic_test.go`?

diag/path.go
```diff
@@ -0,0 +1,16 @@
+package diag
+
+import "github.com/zclconf/go-cty/cty"
+
+func JoinPath(parent cty.Path, child cty.Path) cty.Path {
```
kmoe, line 5: Does this need to be public? Also, I wonder if we can get away with putting this sort of pure cty functionality in another (internal?) package, as it isn't `diag`-specific.

appilon: does not need to be public. It's somewhat niche logic honestly I can just encapsulate it in place where it used I don't suspect it will be used again.

internal/plugin/convert/diagnostics.go
```diff
@@ -50,38 +40,57 @@ func AppendProtoDiag(diags []*proto.Diagnostic, d interface{}) []*proto.Diagnost
...
 	}
 
 	return diags
 }
 
+func DiagsToProto(diags diag.Diagnostics) []*proto.Diagnostic {
+	var ds []*proto.Diagnostic
+	for _, d := range diags {
+		if err := d.Validate(); err != nil {
+			panic(fmt.Errorf("Invalid diagnostic: %s. This is always a bug in the provider implementation", err))
```
paultyng, line 71: why is this is an `Errorf` inside the panic? Should this func just return an error?

diag/diagnostic.go
```diff
@@ -0,0 +1,58 @@
...
+		if d.Severity == sev {
+			validSev = true
+			break
+		}
+	}
+	if !validSev {
+		return fmt.Errorf("invalid severity: %v", d.Severity)
+	}
+	if d.Summary == "" {
+		return fmt.Errorf("empty detail")
```
paultyng, line 39:
```suggestion
		return fmt.Errorf("empty summary")
```

helper/schema/resource.go
```diff
@@ -348,31 +368,31 @@ func (r *Resource) Apply(
...
-	err = nil
 	if data.Id() == "" {
 		// We're creating, it is a new resource.
 		data.MarkNewResource()
-		err = r.create(ctx, data, meta)
+		diags = append(diags, r.create(ctx, data, meta)...)
 	} else {
 		if !r.updateFuncSet() {
-			return s, fmt.Errorf("doesn't support update")
+			return s, append(diags, diag.FromErr(fmt.Errorf("doesn't support update")))
```
paultyng, line 390: not a big deal, but if this isn't actually formatting, can you just construct the struct manually here? the implementation of `Errorf` is non-trivial at least and pretty unnecessary here: https://golang.org/src/fmt/errors.go?s=624:674#L7

diag/diagnostic.go
appilon: good catch

internal/plugin/convert/diagnostics.go
appilon: I wanted to add the line that this should be considered a bug in the provider implementation. wanted the errors coming from the diagnostic validation to be terse

paultyng:
Temporarily converted to draft while we work out the `cty` stuff, don't want to accidentally merge this.

Issue #63, opened 2019-09-26
Cleanup, sync and commit actions for Providers


mongrelion:
Hi!

The Provider API could use a `TeardownFunc` method just like it offers a `ConfigureFunc` method.
Basically, some use cases need to make a final call to some API after the whole Terraform run is done, say, for example, in cases where the client used to communicate with the API does not support token-based authentication but instead user/password combination, and must make a call in the end to kill the session (some systems start to show issues if you leave too many connections open as the timeout is quite long and the developer may not have control whatsoever over that value).
There is another use case for this that I can think of at the moment and that is the Cobbler provider: by the end of creating a bunch of systems you need to make a call to their `sync` method, which you can call from when you create a resource from within Terraform but this has lead to odd behaviour as Cobbler is not thread safe and making a call to this `sync` method several times in a short period of time will make things look funky (https://github.com/cobbler/cobbler/issues/1570 & https://github.com/hashicorp/terraform/pull/5969)

Something like this is what I have in mind:

``` go
schema.Provider{
  Schema: ...,
  ResourcesMap: ...,
  ConfigureFunc: func (d *schema.ResourceData) ({}interface, error) {},
  TeardownFunc: func (meta {}interface) (error) {},
}
```

And the teardown method would receive the value previously returned by the `ConfigureFunc`.

What do you guys think?

paultyng:
The protocol has a `Close` which we plan to expose in the future. That being said, this will strictly be for cleanup, session logout, etc, not for commit type operations. Terraform's state model doesn't have any concept of a "commit" or what resources are committed, and as far as I know the plans to add this are not on the roadmap. If you need to support a commit model, there are potential resource design changes you could make to make it work, and I'm happy to discuss those on specific providers, but without a significant change to Terraform upstream, I don't think this will be added anytime soon.

jwatte:
That is unfortunate. I, too, have the need for stashing a bunch of changes, and then committing them at the end (which may fail if the whole batch isn't overall consistent.)

The workaround of "one giant synthetic resource that sources everything else" is imperfect at best.

AndrewAlexMac:
I do a lot of work involving DNS records, and it's currently very easy for Terraform to temporarily break these during a large plan because there's no guarantee on how the changes will be scheduled, particularly when these records relate to resources the plan doesn't know about.

[...]

PR #377, opened 2020-04-03
Add customizable Stop to provider

paultyng:


helper/schema/provider.go
```diff
@@ -60,11 +60,20 @@ type Provider struct {
 	// provider doesn't need to be configured, this can be omitted.
 	ConfigureContextFunc ConfigureContextFunc
 
+	// StopFunc is called from the gRPC server Stop call. This does not
```
appilon, line 63: Only nitpick is the wording here, I find the phrase `Terraform graph` a bit ambiguous, I would suggest something like
```
StopFunc is called from the gRPC server Stop call. This is an opportunity to cleanup things outside of Terraform's resource lifecycle and management (ie. session files, etc).
```
I'm no wordsmith though, what is here is already pretty good.

paultyng: Want to write this up as a suggestion and I can merge it?

paultyng:
Related to #63

appilon:
Unfortunately after discussion we want to offer a proper shutdown hook but it can't be done through gRPC stop message. Closing until a shutdown hook makes it's way into the protocol or through `go-plugin`

PR #378, opened 2020-04-03
Binary testing: Fix cleanup of temporary directories

kmoe:
Fix a bug in which temporary directories created during a binary test run were not cleaned up at the end of the run.


appilon:
@kmoe friendly reminder to recreate the codechange (don't bring along changelog) to V2

PR #379, opened 2020-04-06
Deprecate Exists

appilon:
Removes context aware Exists function added in the context work and deprecates exists.

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -103,11 +103,20 @@ type Resource struct {
 	Read   ReadFunc
 	Update UpdateFunc
 	Delete DeleteFunc
+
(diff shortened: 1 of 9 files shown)


helper/schema/resource.go
```diff
@@ -505,7 +493,7 @@ func (r *Resource) RefreshWithoutUpgrade(
 		}
 	}
 
-	if r.existsFuncSet() {
+	if r.Exists != nil {
 		// Make a copy of data so that if it is modified it doesn't
```
paultyng, line 497: I wonder if we should warn here?

appilon: good idea

PR #380, opened 2020-04-07
Adding PairedWith Validation Check

mbfrahry:
Adding a new validation check for a use case where attributes must be specified together but they don't fit in the List/Set model.

```
go test ./helper/schema/ -run TestValidatePairedWithAttributes
ok  	github.com/hashicorp/terraform-plugin-sdk/helper/schema	1.670s
```

Also, I'm open to a different name. `ContigentOn` seems like a viable option as well.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -223,9 +223,13 @@ type Schema struct {
 	//
 	// AtLeastOneOf is a set of schema keys that, when set, at least one of
 	// the keys in that list must be specified.
+	//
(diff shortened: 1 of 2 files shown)


bflad:
This looks related as well: https://github.com/hashicorp/terraform-plugin-sdk/pull/342

mbfrahry:
Oh good find! Thanks @bflad.

appilon:
Closed in favor of #342

PR #385, opened 2020-04-13
Add package docs for binary test driver

paultyng:


acctest/doc.go
```diff
@@ -0,0 +1,31 @@
+package acctest
+
+// Package acctest provides the ability to opt-in to the new binary test driver. The binary
+// test driver allows you to run your acceptance tests with a binary of Terraform instead of
+// a source code version included with the testing code. This allows for a number of important
+// enhancements, but most notably a more realistic testing experience and matrix testing
+// against multiple versions of Terraform CLI. This also allows the SDK to be completely
+// separated, at a dependency level, from the Terraform CLI.
```
appilon, line 8:
```suggestion
// separated, at a dependency level, from the Terraform CLI, as long as it is >= 0.12.0
```
The binary driver needs to be API compatible with those json commands.

acctest/doc.go
```diff
@@ -0,0 +1,31 @@
...
+//
+// In v2 of the SDK, the binary test driver will be mandatory.
+//
+// After importing this package, you can add code similar to the following:
+//
+//   func TestMain(m *testing.M) {
+//     acctest.UseBinaryDriver("provider_name", Provider)
+//     resource.TestMain(m)
+//   }
+//
```
appilon, line 22:
```suggestion
// Where `Provider` is the function that returns the instance of a configured `terraform.ResourceProvider`
```

acctest/doc.go
```diff
@@ -0,0 +1,31 @@
...
+// Some providers already have a TestMain defined, usually for the purpose of enabling test
+// sweepers. These additional occurrences should be removed.
+//
+// Initialising the binary test helper using UseBinaryDriver causes all tests to be run using
+// the new binary driver. Until SDK v2, the DisableBinaryDriver boolean property can be used
+// to use the legacy test driver for an individual TestCase.
+//
+// It is no longer necessary to import other Terraform providers as Go modules: these
```
appilon, line 30: There is something about Import tests that still forwards along a map of `map[string]terraform.ResourceProvider`. We should see if it can be removed entirely it might have been an artifact from the initial shimming/PoC.

acctest/doc.go
```diff
@@ -0,0 +1,31 @@
+package acctest
+
+// Package acctest provides the ability to opt-in to the new binary test driver. The binary
```
kmoe, line 3:
```suggestion
// Package acctest provides the ability to opt in to the new binary test driver. The binary
```

appilon:
@paultyng please cherry pick/recreate and PR to v2

PR #388, opened 2020-04-15
Use HashiCorp Cty fork

appilon:
Temporary solution to using the hashicorp cty fork is to internalize HCL until `internal/configs/configschema` no longer depends on HCL (this made it impossible to substitute `zclconf/go-cty` with `hashicorp/go-cty`). The internalized version of HCL has been updated to depend on `hashicorp/go-cty`.


appilon:
Likely closing in favor of #391

appilon:
Closed in favor of #391

PR #397, opened 2020-04-20
remove resource.TestT

appilon:
The interface and associated tests are not useful for V2

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -450,7 +450,7 @@ type TestStep struct {
 // Set to a file mask in sprintf format where %s is test name
 const envLogPathMask = "TF_LOG_PATH_MASK"
 
-func logOutput(t TestT) (logOutput io.Writer, err error) {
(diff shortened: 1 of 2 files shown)


paultyng:
I think you may want to keep this so you don't need to import the `testing` package in live code?

appilon:
Probably, except we currently do in `RunNewTest` which probably was an oversight in V1 (so testing is exposed in the V1 API as it is). We type asserted to `testing.T` to satisfy a testing interface in `terraform-plugin-test`.

We will have to harmonize the interfaces if we don't want to directly import `testing` in V2

appilon:
@paultyng We should make them both the same I like that idea but then we are exposing an interface owned by mitchell to the surface. I can try type asserting to `terraform-plugin-test`'s interface and seeing if that works.

appilon:
#396 opts to simulate fatal on mockT with panic/recover, cherry-picking that to V2 will remove the need for this PR.

cleaning up the interface usage and casting in V2 can be done in another PR

PR #399, opened 2020-04-21
Unexpose testing pkg

appilon:
This PR scrubs the SDK surface of uses of Go's `testing` package, `testing.T` and `testing.M` were exposed to the SDK surface as an oversight in V1. It instead propagates the interface previously known as `resource.TestT` with a few added signatures so it can be compatible with a testing interface present in `terraform-plugin-test` (the driver library for acceptance tests). The interface had to be moved to its own package (now `helper/resource/testing`) to avoid circular imports.

Ideally we could remove any imports of Go's`testing` in non `_test.go` files from the SDK. However the acceptance test framework checks if `go test` was called with verbose enabled. This reference could not be removed unless we remove that requirement.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -574,7 +574,7 @@ func Test(t TestT, c TestCase) {
 		t.Fatal("Please configure the acctest binary driver")
 	}
 
-	RunNewTest(t.(*testing.T), c, providers)
+	RunNewTest(t, c, providers)
 }
 
 // testProviderConfig takes the list of Providers in a TestCase and returns a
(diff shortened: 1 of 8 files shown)


helper/resource/testing.go
```diff
@@ -537,7 +538,7 @@ func Test(t TestT, c TestCase) {
 	log.SetOutput(logWriter)
 
 	// We require verbose mode so that the user knows what is going on.
-	if !testing.Verbose() && !c.IsUnitTest {
+	if !gotesting.Verbose() && !c.IsUnitTest {
```
paddycarver, line 541: Could we wrap this in the `testing` package we created? So we don't need to import both of them here, at least? Which avoids the confusing import name clash?

appilon: yes no problem

PR #403, opened 2020-04-23
Port provider_meta functionality from core.

paddycarver:
Port the functionality added in hashicorp/terraform#22583 that allows
providers to receive arbitrary, provider-definable module-scoped
information from configs.

### DIFF
--- a/helper/schema/resource_data.go
+++ b/helper/schema/resource_data.go
@@ -7,6 +7,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/go-cty/cty/gocty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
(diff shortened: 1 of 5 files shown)


helper/schema/resource_data.go
```diff
@@ -506,3 +509,10 @@ func (d *ResourceData) get(addr []string, source getSource) getResult {
 		Schema:         schema,
 	}
 }
+
+func (d *ResourceData) GetProviderMeta(dst interface{}) error {
```
appilon, line 513: Any reason this function didn't return `interface{}, error` I'm not as familiar with the underlying value of `interface{}` I'm guessing it's pointer-like otherwise this wouldn't even work (function get's copy of value, we update its value, but the developer calling this function never gets the value)

paddycarver: When I wrote this, months ago, I had two options: I could either do `GetProviderMeta() (cty.Value, error)`, or I could do `GetProviderMeta(interface{}) error`. The first would return the cty.Value that came over the protocol, and the user would be responsible for turning that into a Go value. The second would automatically turn the value that came over the wire into a Go value, much like `json.Unmarshal` does, and like the underlying `gocty.FromCtyValue` does.

I figured there wasn't a whole lot the user would reasonably want to do with the cty.Value except unmarshal it into a struct, so I opted for the second approach, so every provider that uses this wouldn't have to write that wrapper themselves.

PR #405, opened 2020-04-23
Port provider_meta functionality from core.

paddycarver:
Port the functionality added in hashicorp/terraform#22583 that allows
providers to receive arbitrary, provider-definable module-scoped
information from configs.

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -330,6 +330,10 @@ func (r *Resource) Apply(
 		return s, append(diags, diag.FromErr(err))
 	}
 
+	if s != nil && data != nil {
+		data.providerMeta = s.ProviderMeta
+	}
+
 	// Instance Diff shoould have the timeout info, need to copy it over to the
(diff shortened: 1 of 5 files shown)


helper/schema/resource.go
```diff
@@ -330,6 +330,10 @@ func (r *Resource) Apply(
 		return s, append(diags, diag.FromErr(err))
 	}
 
+	if s != nil && data != nil {
```
appilon, line 333: Assuming its okay for these to be nil in certain situations? No need for logging warnings or panicking?

appilon: I suppose in the context of provider schema being an entirely optional feature sanity checking nil like this is safe and we don't really need to think about why something might be nil, the rest of the SDK should concern itself with that, not this feature.

paddycarver: In most cases, this will be nil. This will only be non-nil when a provider_meta block is set for this provider.

PR #406, opened 2020-04-23
Remove Verbose check

paultyng:


paultyng:
This currently requires mitchellh/go-testing-interface @ v1.14.1 but it could be relaxed to v1.0.3. There were some complications in the publishing, so waiting for that to be fixed so we can relax the constraint here to not force go 1.14.

helper/resource/testing.go
```diff
@@ -538,11 +540,6 @@ func Test(t testing.T, c TestCase) {
 	}
 	log.SetOutput(logWriter)
 
-	// We require verbose mode so that the user knows what is going on.
-	if !testing.Verbose() && !c.IsUnitTest {
```
appilon, line 542: I need to see is `IsUnitTest` is used anywhere now, worried too much was cut in V2 and there are no safeguards around accidentally running acceptance tests that cost money

helper/resource/testing_test.go
```diff
@@ -71,9 +74,8 @@ func TestTest_factoryError(t *testing.T) {
 		})
 	}()
 
-	fatalStr := fmt.Sprint(mt.FatalArgs)
-	if !strings.Contains(fatalStr, resourceFactoryError.Error()) {
-		t.Fatalf("test should've failed with %s, failed with %s", resourceFactoryError, fatalStr)
```
appilon, line 76: I hate this test in general I think it's pointless and actually prefer delete it, but aside from that I wanted to make sure the code Fatal'd at the place we expected it to, and not for another reason.

helper/resource/testing_test.go
```diff
@@ -156,67 +158,19 @@ func TestComposeTestCheckFunc(t *testing.T) {
...
-	t.f = true
-
-	panic("mockT.FailNow")
-}
-
-func (t *mockT) Fatal(args ...interface{}) {
-	t.FatalCalled = true
-	t.FatalArgs = args
-	t.f = true
+	testinginterface.RuntimeT
```
appilon, line 161: Ah I see now the data captured in the mock is removed. Honestly the 2 unit tests that use the mock are somewhat superfluous. I think I would prefer expose mitchell's lib on the SDK surface and just delete those 2 tests and `mockT` altogether.

helper/resource/testing_test.go
appilon: I see mockT no longer captures the information it once did.

helper/resource/testing.go
appilon: nvm this is fine

helper/resource/testing_test.go
paultyng: Considering its a `_test` package I just went for minimal change here. Its not exposed externally really and we did have testing to check if parallel was invoked, so I needed a wrapper to capture that.

paultyng:
@appilon the upstream issue is resolved

PR #407, opened 2020-04-24
helper/schema: Additional validation for schema attribute references

bflad:
Rebased #365
Reference: #71

Builds on testing from https://github.com/hashicorp/terraform-plugin-sdk/pull/364 (targeting v1). This PR currently targets v2.


bflad:
@appilon sure thing, I'll submit another PR after rebasing this one so the difference is super easy to replicate. 👍

bflad:
v1-maint pull request with cherry-picked commits 👍 https://github.com/hashicorp/terraform-plugin-sdk/pull/416

PR #414, opened 2020-04-28
Remove Removed and prune related code

paultyng:


helper/resource/testing_new_import_state.go
```diff
@@ -94,23 +92,6 @@ func testStepNewImportState(t testing.T, c TestCase, wd *tftest.WorkingDir, step
 					r.Primary.ID)
 			}
 
-			// We'll try our best to find the schema for this resource type
```
appilon, line 97: Not sure on this deletion, this was copy pasted directly from the old implementation of the acctest framework.

paultyng: We can let @kmoe chime in on this, but I think it was just to avoid the diff mismatch in import tests.

paultyng:
@appilon not as far as I can tell. I simulated manually adding an unrecognized field to state, and its ignored silently. So I don't think it serves any purpose.

appilon:
If we can quickly confirm import tests are unaffected I'm fine with merging.

appilon:
Needs a rebase after recent PR being merged

paultyng:
Rebased

paultyng:
Here is a test run with import tests referencing this commit, it has no removed attributes though: https://github.com/paultyng/terraform-provider-unifi/runs/630973829

bflad:
Fan out relating to this:

* v1 deprecations: https://github.com/hashicorp/terraform-plugin-sdk/pull/424
* terraform-website update: https://github.com/hashicorp/terraform-website/pull/1229
* tfproviderlint check: https://github.com/bflad/tfproviderlint/issues/168

paultyng:
Thanks @bflad!

PR #417, opened 2020-04-29
TypeSet Check Heuristic

appilon:
This change can assist users when enabling the binary driver. Previously any checks that addressed through TypeSet hashes would error confusingly (just claim the resource does not exist) as the binary driver's shimmed state did not represent TypeSet addresses.

A small hack/flag needed to be added onto `terraform.State`, but the docs clearly state it's not for external use.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -1439,18 +1439,20 @@ func indexesIntoTypeSet(key string) bool {
 
 func checkIfIndexesIntoTypeSet(key string, f TestCheckFunc) TestCheckFunc {
 	return func(s *terraform.State) error {
-		if s.IsBinaryDrivenTest && indexesIntoTypeSet(key) {
-			return fmt.Errorf("Test check address %q likely indexes into TypeSet\nThis is not possible in V1 of the SDK while using the binary driver\nPlease disable the driver for this TestCase with DisableBinaryDriver: true", key)
+		err := f(s)
+		if err != nil && s.IsBinaryDrivenTest && indexesIntoTypeSet(key) {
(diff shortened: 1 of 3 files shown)


kmoe:
This is a great UX improvement and fills the gap between binary testing in v1 and the fix for #196 (which I still hope I can land on v1 as well as v2). 

If we're concerned about false positives, could we instead run the heuristic and log the warning only if the attribute check fails?

helper/resource/testing.go
```diff
@@ -1423,3 +1424,36 @@ func detailedErrorMessage(err error) string {
 		return err.Error()
 	}
 }
+
+// indexesIntoTypeSet is a heuristic to try and identify if a flatmap style
+// string address uses a precalculated TypeSet hash, which are always 10 digits
+// long and are a number
+func indexesIntoTypeSet(key string) bool {
+	for _, part := range strings.Split(key, ".") {
+		if len(part) == 10 {
```
bflad, line 1433: `TypeSet` hashes can be any positive integer from `0` to `MaxInt` as `SchemaSetFunc` requires an `int` return type: https://github.com/hashicorp/terraform-plugin-sdk/blob/7952a20c9d3c4c46fb0fa855557b8c556283dd23/helper/schema/schema.go#L310-L312

Here's an example test that include varying length hashes: https://github.com/terraform-providers/terraform-provider-aws/blob/7350b304548ea79dff6a626cfd2041057069509f/aws/resource_aws_lb_listener_rule_test.go#L664-L686

To increase the accuracy of this it might be worth going through each part and instead comparing if its an integer then its integer value being greater than 10 or 100 (or something no one would be reasonably testing a `TypeList` with).

appilon: Ah thank you for the info there, I agree a value greater than 100 is likely a better catch in this case

PR #418, opened 2020-04-29
helper/schema: Allow self references with AllowOneOf, ExactlyOneOf, and RequiredWith in InternalValidate

bflad:
Reference: #407
Reference: #416

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -750,28 +750,28 @@ func (m schemaMap) internalValidate(topSchemaMap schemaMap, attrsOnly bool) erro
 		}
 
 		if len(v.ConflictsWith) > 0 {
-			err := checkKeysAgainstSchemaFlags(k, v.ConflictsWith, topSchemaMap, v)
+			err := checkKeysAgainstSchemaFlags(k, v.ConflictsWith, topSchemaMap, v, false)
 			if err != nil {
 				return fmt.Errorf("ConflictsWith: %+v", err)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


helper/schema/schema.go
```diff
@@ -948,6 +954,11 @@ func checkKeysAgainstSchemaFlags(k string, keys []string, topSchemaMap schemaMap
 			return fmt.Errorf("%s cannot contain Computed(When) attribute (%s)", k, key)
 		}
 	}
+
+	if requireSelfReference && !foundSelfReference {
```
appilon, line 958: `AtLeastOneOf` and `ExactlyOneOf` were implemented with an idempotent line that adds itself and removes duplicate, it allows the self reference to be optional. I would change this PR to have a flag for `allowSelfRef` and ditch the requirement for it, but rather disallow for `ConflictsWith` (and possibly `RequiredWith`)

appilon: Unless we remove that line in `validateRequiredWith` I would actually ONLY have this error for `ConflictsWith`, just to be consistent, I will approve either final implementation.

paultyng:
Per Alex's comment, may be worth referencing back to the RFC, there was some discussion about the requirement of the self reference, I think we landed on not requiring it for a reason, though I can't recall the reason 😂

bflad:
I'll update this to be allow vs require and add `RequiresWhen` unit testing as well to ensure that new attribute is covered since its not currently (with allowed self-references for it). 👍

bflad:
Implementation updated and submitted v1 backport: #423

PR #436, opened 2020-05-01
Remove helper/encryption package.

paddycarver:
As we no longer support encrypting state
(https://www.terraform.io/docs/extend/best-practices/sensitive-state.html#don-39-t-encrypt-state)
let's get rid of the helper/encryption package that we supported for
specifically that use case.


appilon:
@paddycarver With this gone we can tree shake internal stuff a bit, mind if I push to your branch? Or it can come in a follow up.

paddycarver:
I'm happy with either approach.

PR #437, opened 2020-05-01
Deprecate helper/encryption.

paddycarver:
In line with
https://www.terraform.io/docs/extend/best-practices/sensitive-state.html#don-39-t-encrypt-state,
mark all the functions in helper/encryption as deprecated and up for
removal in v2.


greut:
Sadly, we make a good use of this. Terraform creates some Vault tokens, which are ciphered in GPG and then given as is to another system via the JSON output.

PR #439, opened 2020-05-03
use cmp.Equal for Set equality

kmoe:


kmoe:
Actually this is a bad idea - `go-cmp` should only be used for tests.

PR #449, opened 2020-05-06
Update diag convenience methods

paultyng:


helper/schema/provider.go
```diff
@@ -243,12 +243,9 @@ func (p *Provider) ValidateResource(
 //
 // This won't be called at all if no provider configuration is given.
 func (p *Provider) Configure(ctx context.Context, c *terraform.ResourceConfig) diag.Diagnostics {
-
-	var diags diag.Diagnostics
-
 	// No configuration
 	if p.ConfigureFunc == nil && p.ConfigureContextFunc == nil {
-		return diags
+		return nil
```
paultyng, line 248: I swapped some of these back, I think returning `nil` when possible is more clear that its not a potential error case. Preserving the go idioms if possible is still useful. When someone does have warnings, the top level `diags` makes sense, but otherwise, I think it may be more effort and compromises on clarity.

helper/schema/resource.go
```diff
@@ -380,7 +374,7 @@ func (r *Resource) Apply(
 		// Reset the data to be stateless since we just destroyed
 		data, err = schemaMap(r.Schema).Data(nil, d)
 		if err != nil {
-			return nil, append(diags, diag.FromErr(err))
+			return nil, diag.FromErr(err)
```
paultyng, line 377:
```suggestion
			return nil, append(diags, diag.FromErr(err)...)
```

diag/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+import "fmt"
+
+// FromErr will convert an error into a Diagnostics. This returns Diagnostics
+// as the most common use case in Go will be handling a single error
+// returned from a function.
+//
+//   if err != nil {
+//     return diag.FromErr(err)
+//   }
+func FromErr(err error) Diagnostics {
```
appilon, line 12: Overall I do like this pattern, however I would still like a convenience to create a singular `Diagnostic` from an error. would it be crazy to keep the helpers for creating on a single diag, but then have a helper to quickly convert to the collection?
```
return diag.FromErr(err).ToDiags()
```

appilon: It's valuable in scenarios where we have gathered warnings, and encounter a function that just returns a go error
```
if err != nil {
    return append(diags, diag.FromErr(err)
}
```

paultyng: I think the thing is, that you are burdening the 80% case for the 20% case, and you could easily rewrite that second one as:

```go
if err != nil {
    return append(diags, diag.FromErr(err)...)
}
```

appilon: I wish the package was called `diags` with this model, but that would collide with calling your slice `diags`.

PR #438, opened 2020-05-01
Remove unnecessary self-contained convenience packages

paultyng:
The commits are discrete and can be cherry-picked.

We should make sure we also add deprecations to v1.

PR #453, opened 2020-05-13
Deprecated packages that were removed in v2

paultyng:
See #438

### DIFF
--- a/helper/hashcode/hashcode.go
+++ b/helper/hashcode/hashcode.go
@@ -8,6 +8,10 @@ import (
 
 // String hashes a string to a unique hashcode.
 //
+// Deprecated: This will be removed in v2 without replacement. If you need
+// its functionality, you can copy it, use crc32 directory, or reference the
+// v1 package.
+//
 // crc32 returns a uint32, but for our use we need
 // and non negative integer. Here we cast to an integer
 // and invert it if the result is negative.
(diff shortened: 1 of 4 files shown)


helper/hashcode/hashcode.go
```diff
@@ -8,6 +8,10 @@ import (
 
 // String hashes a string to a unique hashcode.
 //
+// Deprecated: This will be removed in v2 without replacement. If you need
+// its functionality, you can copy it, use crc32 directory, or reference the
```
appilon, line 12:
```suggestion
// its functionality, you can copy it, import crc32 directly, or reference the
```

helper/hashcode/hashcode.go
```diff
@@ -24,6 +28,10 @@ func String(s string) int {
 }
 
 // Strings hashes a list of strings to a unique hashcode.
+//
+// Deprecated: This will be removed in v2 without replacement. If you need
+// its functionality, you can copy it, use crc32 directory, or reference the
```
appilon, line 33:
```suggestion
// its functionality, you can copy it, import crc32 directly, or reference the
```

helper/hashcode/hashcode.go
appilon: You can leave the word "use" if you prefer, but I think you meant "directly"

Issue #455, opened 2020-05-14
Better handling of schema panic on error


paultyng:
This is a follow-on to #451 

We may be able to plumb a `testing.T` down to the `ResourceData` (perhaps via `context.Context`) and a gRPC interceptor or something. If it were available, instead of a `panic` the SDK could simply `t.Errorf` but continue execution.

Considering time and cost of acc test runs, this would probably be a more user friendly approach.

bflad:
Unofficially since terraform-plugin-framework v1.0.0, there are no more plans for major refactoring of `helper/schema`, so closing this out. Framework based providers do not have this issue.

PR #462, opened 2020-05-26
Remove TF_SCHEMA_PANIC_ON_ERR

### LINKED ISSUE #451: Enable TF_SCHEMA_PANIC_ON_ERROR Only During Acceptance Testing
### SDK version

```
v2.0.0-rc.1
```

### Use-cases

For a long time, the acceptance testing framework has accepted the `TF_SCHEMA_PANIC_ON_ERROR` environment variable as a way to tease out state setting and drift detection issues in resources where the `(*helper/schema.ResourceData).Set()` call is not error checked, e.g.

```go
func ExampleThingRead(d *schema.ResourceData, meta interface{}) error {
  // ...
  d.Set("attr", someInvalidValue)
  // ...
}
```

In the upcoming SDK v2, the value has been flipped on by default, which is great for catching provider bugs, but unfortunately at the potential expense of very poor user experience. When a resource panics, it makes the resource practically unusable (since this affects refresh) and potentially can leave the user's Terraform state in an errant state (depending on how Terraform CLI handles the panic). Considering that it can be fairly easy for larger providers to miss failing acceptance testing during merges and before releases, this can cause a lot of issues to be submitted and an immediate need to cut a bug fix release. Asking users to run with `TF_SCHEMA_PANIC_ON_ERROR` disabled could be problematic in its own right.

### Proposal

Consider guarding the `TF_SCHEMA_PANIC_ON_ERROR` behavior to only be enabled by default in the acceptance testing framework. For example, acceptance testing today requires `TF_ACC=1` for `helper/resource.Test()` / `helper/resource.ParallelTest()` so that could potentially be used.

### References

- https://github.com/hashicorp/terraform-plugin-sdk/pull/344

appilon:
Schema errors will panic if TF_ACC is set, panicking is no longer on by default
but cannot be opted out of when running acceptance tests. This does not address #455 .

Technically a breaking change `schema.PanicOnErr` is no longer an exported var

Closes #451

### DIFF
--- a/helper/resource/testing_test.go
+++ b/helper/resource/testing_test.go
@@ -18,11 +18,6 @@ import (
 )
 
 func init() {
-	// TODO: Remove when we remove the guard on id checks
-	if err := os.Setenv("TF_ACC_IDONLY", "1"); err != nil {
-		panic(err)
-	}
-
 	if err := os.Setenv(testEnvVar, "1"); err != nil {
 		panic(err)
 	}
--- a/helper/schema/schema_test.go
+++ b/helper/schema/schema_test.go
@@ -8209,48 +8209,26 @@ func TestValidateAtLeastOneOfAttributes(t *testing.T) {
 	}
 }
 
-func Test_panicOnErrDefaultTrue(t *testing.T) {
-	oldEnv := os.Getenv(PanicOnErr)
+func TestPanicOnErrorDefaultsFalse(t *testing.T) {
+	oldEnv := os.Getenv("TF_ACC")
 
-	os.Setenv(PanicOnErr, "")
-	if !schemaMap(nil).panicOnError() {
-		t.Fatalf("Empty %s should default to true", PanicOnErr)
-	}
-
-	os.Setenv(PanicOnErr, oldEnv)
-}
-
-func Test_panicOnErrParsableTrue(t *testing.T) {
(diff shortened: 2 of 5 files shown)


helper/resource/testing_test.go
```diff
@@ -18,11 +18,6 @@ import (
 )
 
 func init() {
-	// TODO: Remove when we remove the guard on id checks
```
appilon, line 21: This is not referenced anywhere else

helper/schema/schema_test.go
```diff
@@ -8209,48 +8209,26 @@ func TestValidateAtLeastOneOfAttributes(t *testing.T) {
...
-	}
-
-	os.Setenv(PanicOnErr, oldEnv)
-}
-
-func Test_panicOnErrParsableFalse(t *testing.T) {
-	oldEnv := os.Getenv(PanicOnErr)
-
-	os.Setenv(PanicOnErr, "false")
+	os.Setenv("TF_ACC", "")
```
paultyng, line 8215: oh god this is just asking for a race condition problem 😭

appilon: Agreed, these unit tests aren't super high value so we could ditch the tests if you find this too "anti pattern"

appilon:
@paddycarver no problem will do.

PR #471, opened 2020-06-09
Bring test runs back into a single process.

paddycarver:
This is a long saga of code that stretches back over months to make tests debuggable.

Historically, tests ran by starting up the Go test framework, which would spin up a shim of Terraform (importing half of Terraform in the process), which would drive the provider code, then the test framework would inspect the state. This had some drawbacks, though; the shim of Terraform wasn't _quite_ Terraform, and could behave in different ways sometimes. As the plugin SDK got separated from the Terraform codebase, this drift potential became worse.

So we came up with the Binary Acceptance Testing driver, which would start up the Go test framework, which would _shell out_ to a Terraform binary, which would start up a provider process to drive, and then the test framework would inspect the state back in the test process. This solved the issue it was addressing--we were now testing against production versions of Terraform, guaranteeing the tests would match what users would see--but it created some problems of its own. Our original model ran in a single process, and as a happy accident of that, it meant that `go test` could report coverage information and delve could be used to debug tests. Our binary acceptance testing model split everything across multiple processes, which broke coverage information (`go test` couldn't see the provider code being run anymore) and delve (delve also couldn't see the provider code). We didn't test, but suspect, that other tooling like race detection would also break.

This led to coordinating with the Terraform core team to come up with a new process. We would still start up the Go test framework, it would still shell out to the Terraform binary, but instead of Terraform spinning up a new process for the provider, a provider server would be started in the same process as the Go test framework, and Terraform would be told to reconnect to it. This kept us using a production Terraform binary, so we don't need to worry about our shim drifting, but also kept our provider and test code in the same process, where it can be reached by delve or any of the other standard testing tooling available.

This PR is adding support for that third approach. The pieces needed in go-plugin and terraform-plugin-test have been merged and released. The pieces in Terraform that are required were included in 0.12.26 and 0.13.0-beta.1. This is the final piece of the puzzle, the code that drives the tests.

As a result of this reworking, the `acctest` package that was called from `TestMain` in providers is no longer needed, and is being removed as part of v2's breaking changes.

Also, as a side effect from this work, providers can be debugged even outside of testing. To support that, the `plugin.DebugServe` helper was added, though a more friendly API may be needed. Providers can include that in their `main` package, with the recommendation being to switch between `plugin.DebugServe` and `plugin.Serve` based on a flag, with the default being `plugin.Serve`. This would allow production versions of binaries to have debuggers attached to them, by starting the binary in debug mode with the debugger attached, then having Terraform reconnect and drive it.

There are a few differences under debug mode. Terraform performs several graph walks per command, and usually starts a provider process at the beginning of each graph walk and kills the process at the end of the graph walk. Because these graph walks have no hooks, it's impossible to support this when the server is started outside of Terraform's control. So when in debug mode, Terraform will not care about the provider lifecycle at all; it will not start the provider, nor will it kill the provider process after graph walks. This means global mutable state will be longer lived than it would be in non-debug mode. Terraform init also will not do anything with providers that are in debug mode; it will not attempt to fetch binaries, match versions, or find binaries on the file system. The information provided out of band when the provider is in debug mode is all Terraform needs to operate, so it will not bother trying to find any other information.

The commits on this PR will be squashed, rewritten, and force-pushed.


paddycarver:
I feel like I should add tests for this, but am drawing a blank on how to test it. Also, this feels like it needs more documentation than it has, but beyond the `plugin` package's few new exported types, I didn't really touch the API surface that much. Does most the documentation for this living on terraform.io feel acceptable?

PR #473, opened 2020-06-11
Fix test logging output.

paddycarver:
In version 1.0.0-1.6.0, our in-process non-binary test driver didn't
call go-plugin, and called log.SetOutput to control what got logged, to
avoid flooding test output with debug logs.

From version 1.6.0, the in-process non-binary test driver had the same
behavior, but we added an out-of-process binary test driver, which used
go-plugin, but had the provider in a separate process that sent its
stdout and stderr through Terraform. Terraform then, not the test
framework, was responsible for filtering the output.

With the introduction of our in-process binary test driver, however,
we're using go-plugin and Terraform isn't filtering log output for us.
In theory, the log.SetOutput call that was working for the in-process
non-binary test driver should've worked here, as well, and it sort of
did. The issue was that go-plugin calls log.SetOutput and sets it to
os.Stderr in plugin.Serve, which overrides our filter.

To get this behavior back, after every call to plugin.Serve, we override
the go-plugin log.SetOutput by calling our own again.

This also removes the helper/resource.logOutput function, replacing its
usage with helper/logging.SetOutput, which does the same thing. The only
difference should be that filtering on test name is no longer supported.
If we want to support that, we would need to plumb a testing.T through
to runProviderCommand, which involves updating a lot of code for
marginal gain; as far as I know, nobody actually uses the test name
filtering for logging, preferring instead to log everything and only run
the test they want logs for.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -5,17 +5,13 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"io"
-	"io/ioutil"
 	"log"
 	"os"
 	"regexp"
 	"strconv"
 	"strings"
-	"syscall"
 
 	"github.com/hashicorp/go-multierror"
-	"github.com/hashicorp/logutils"
 	tftest "github.com/hashicorp/terraform-plugin-test"
 	testing "github.com/mitchellh/go-testing-interface"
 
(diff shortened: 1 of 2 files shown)


helper/resource/testing.go
```diff
@@ -451,49 +447,6 @@ type TestStep struct {
 	providers map[string]*schema.Provider
 }
 
-// Set to a file mask in sprintf format where %s is test name
-const envLogPathMask = "TF_LOG_PATH_MASK"
```
appilon, line 455: I believe `TF_LOG_PATH_MASK` may be used in our acctests in TeamCity, is this accounted for in the new implementation?

appilon: I am investigating if this is something we can live without

PR #474, opened 2020-06-11
Add helper/schema.Provider.UserAgent helper.

paddycarver:
Add a helper function to generate the suggested User-Agent header for
provider requests.


paddycarver:
Changed the logic to account for empty provider name or version. @paultyng, still happy with it?

PR #479, opened 2020-06-22
Better description of SchemaDiffSuppressFund

abitrolly:
I found that the notion of `old` and `new` arguments for `SchemaDiffSuppressFund` is confusing when the resource is changed outside of Terraform.

For example, I had changed GitLab remote branch to "master".
```
default_branch                                   = "master" -> null
```
Now Terraform want to change the value back to `null`. However, in this context, the `null` is the old value, but for SchemaDiffSuppressFund the `old` argument contains the string "master".

The suggested description could be improved. I tried to do my best from my current understanding of Terraform and lack of proper terminology.

https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#diffsuppressfunc

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -257,12 +257,12 @@ const (
 	SchemaConfigModeBlock
 )
 
-// SchemaDiffSuppressFunc is a function which can be used to determine
-// whether a detected diff on a schema element is "valid" or not, and
-// suppress it from the plan if necessary.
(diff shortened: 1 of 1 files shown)


abitrolly:
Although it is not explicitly mentioned, the diff is between `current` and `planned` value of remote resource, and the `current` value in the above case with `default_branch` is newer than the old `planned` value from Terraform config.

PR #488, opened 2020-07-07
Nil check error in `diag.FromErr`

magodo:
This allows following use cases:

```go
func foo() error {
    //...
}

func bar() diag.diagnostics {
    //...
    return diag.FromErr(foo())
}
```


appilon:
I may have approved a little too quickly, it is something we are aware of but we recently moved away from a similar pattern in #199 as it can lead developers to subtle bugs.

PR #496, opened 2020-07-09
Mark method as helper

paultyng:
Need to mark all SDK internal helpers so they don't show up in test messaging over user test code:

Example bad log:

```
=== RUN   TestAccResourceID
    TestAccResourceID: testing_new_config.go:54: Check 1/2 error: base64 string length is 0; want 6
--- FAIL: TestAccResourceID (2.15s)
```


paultyng:
Closing this as I think it will be covered in the `reattachmultiple` work

Issue #475, opened 2020-06-14
Does the suffix for name_prefix need to be that long?


edward2a:
Hi all,

I had some instances where name_prefix ended up being something very limiting (not entirely your fault, limits on the cloud platform too) and having to use a random_id instead, as a result of the (unnecessarily long IMHO) UniqueId suffix generated by tf.

Do you really need to set such a long unique Id?
I, personally, don't need or see the point for a timestamp, plus an index, plus 'terraform' word.

You could shorten this by using shorter versions of the values you propose, and a smaller max value for the index (4294967295 updates would probably be a hefty amount of years beyond the life expectancy of this tech :P).

Sample:
```go
package main

import (
  "fmt"
  "strings"
  "time"
)

func main() {
  var old_idx uint32

  old_str := "terraform-"
  old_t := strings.Replace(time.Now().UTC().Format("20060102150405.0000"), ".", "", 1)
  old_idx = 4294967295

  var new_idx uint32
  new_str := "tf"
  new_t   := time.Now().UTC().Unix()
  new_idx = 16777215

  fmt.Printf("Old suffix is %s%s%08x\n", old_str, old_t, old_idx)
  fmt.Printf("New suffix is %s%v%06x\n", new_str, new_t, new_idx)
}
```

Output:
```
Old suffix is terraform-202006141048156188ffffffff
New suffix is tf1592131695ffffff
```

Readability counts I agree... but there is a boundary between it and practicality :)

Thank you,
Ed

paddycarver:
Hi @edward2a! What SDK function is this relating to? `name_prefix` is a provider feature, not something built into the SDK, as far as I'm aware, so I'd love some more information about what specific code you're talking about here.

edward2a:
Hi @paddycarver , this comes specifically from the following validation:

https://github.com/terraform-providers/terraform-provider-aws/blob/ddd0d885149817dfca29f5847978c432bd0444d9/aws/validators.go#L2307

In this case, the name_prefix is for this AWS resource is limited to 6 chars.

Bear in mind that name_prefix feature is used in many providers which depend on having a unique name either due to platform limitations of for create_before_destroy.

paddycarver:
[...]

PR #508, opened 2020-07-21
helper/resource: Use base 36 to generate incremental ids

bholzer:
This is a copy of the PR at https://github.com/hashicorp/terraform/pull/16997

The current implementation makes the `name_prefix` of some resources in AWS pretty much useless. This is intended to solve that problem.

### **What changes**
- Instead of using a human readable representation of a date (such as "201712280947212674") which doesn't fully use base 10 (months go only up to 12, days up to 31, minutes and seconds up to 60), use a timestamp which is an always incrementing value, which does not skip steps and is therefore more compact.
- Rather than using base 10 to represent the timestamp, use base 36 ([a-z0-9]) this gives a wider range of characters to express the number making the suffix even more compact. I went with 36 instead of the classic 64 or 62 ([a-zA-Z0-9]) because some systems do not care about the case or do not deal with 'special characters'.
- Instead of using a base16 incremental count, use base36 (same reasoning as above)
- Instead of having a fixed length which requires padding and leads to waste, concatenate the timestamp to the id and let the suffix grow.

This leads to these changes:

- `20171228094721267400000004` becomes `p1o0ix4`
- `20171228094721267400010000` becomes `p1o0ixcra`

### **Notes**
- The length of the suffix is not fixed
- The timestamp and the id have to be be converted independently to keep the same "prefix"
- The creation order of the resources across multiple runs is kept as the suffixes can be sorted alphabetically (note that `9 < a`).
- The creation order of the resources within the same run is not going to be easily accessible (as we're not padding with 0), for example `p1o0ix4` is the 4th object created, `p1o0ix5` is the 5th one and `p1o0ix40` is the 144th, ordering alphabetically will show the order p1o0ix4, p1o0ix40, p1o0ix5
- The milliseconds have been dropped from the timestamp. They could be added easily (we could even use nanoseconds, but it feels like a waste of space.
- The order of the resources will be less accessible when the timestamp grows by one order of magnitude. Today we got the timestamp as `p1o0ix`, the next move will be at `1000000` which is `2176782336` in base 10 or on the 24th of December 2038 (the following one being in the year 4453).
- If the above point is a concern, adding milliseconds change the timestamp into `jbqb2lgo` which moves the next order of magnitude change to `100000000` which is `2821109907456` base 10 which is the 25th of May 2059. I'm not sure that the minor inconvenience of not being able to sort warrants the extra length, but for two characters it isn't the end of the world either.

### DIFF
--- a/helper/resource/id.go
+++ b/helper/resource/id.go
@@ -25,21 +25,23 @@ const UniqueIDSuffixLength = 26
 
 // Helper for a resource to generate a unique identifier w/ given prefix
 //
-// After the prefix, the ID consists of an incrementing 26 digit value (to match
-// previous timestamp output).  After the prefix, the ID consists of a timestamp
-// and an incrementing 8 hex digit value The timestamp means that multiple IDs
-// created with the same prefix will sort in the order of their creation, even
-// across multiple terraform executions, as long as the clock is not turned back
-// between calls, and as long as any given terraform execution generates fewer
-// than 4 billion IDs.
+// After the prefix, the ID consists of an incrementing base 36 value.
+// The value is made of the current timestamp in base 36 followed by an
+// incrementing base 36 counter.
+// This means that the identifier can grow in size depending on the number of
+// items handled by terraform.
+// Because the first part is using the timestamp, it is always possible to sort
+// the identifiers of multiple resources alphabetically (with numbers first)
+// to get the list of resources created from the oldest to the newest.
+// The timestamp means that multiple IDs created with the same prefix will sort
+// in the order of their creation, even across multiple terraform executions, as
+// long as the clock is not turned back between calls.
 func PrefixedUniqueId(prefix string) string {
-	// Be precise to 4 digits of fractional seconds, but remove the dot before the
-	// fractional seconds.
-	timestamp := strings.Replace(
-		time.Now().UTC().Format("20060102150405.0000"), ".", "", 1)
+	timestamp := strconv.FormatUint(uint64(time.Now().Unix()), 36)
 
 	idMutex.Lock()
 	defer idMutex.Unlock()
 	idCounter++
-	return fmt.Sprintf("%s%s%08x", prefix, timestamp, idCounter)
+	id := strconv.FormatUint(uint64(idCounter), 36)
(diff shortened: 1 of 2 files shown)


paddycarver:
I think, per #475, we may want to address this by adding a new helper function in `helper/schema`, instead of changing the existing function in `helper/resource`, and then deprecating the `helper/resource` version.

dturnbu:
> I think, per #475, we may want to address this by adding a new helper function in `helper/schema`, instead of changing the existing function in `helper/resource`, and then deprecating the `helper/resource` version.

Given that the suffix added by PrefixedUniqueId is really just junk data meant to ensure a unique name, is it really necessary to go down the path of designing and implementing new helper functions rather than just approving and merging the changes in this PR? 

I'm all for doing things properly, but this change doesn't dig us further into a hole, instead it simply and elegantly solves a problem that people are facing right now.

bflad:
> Given that the suffix added by PrefixedUniqueId is really just junk data meant to ensure a unique name

While the unique identifier information may seem like "junk data", it is data with a specific, consistent format that has been present in Terraform for many versions. For better or worse, downstream code can ([and does](https://github.com/terraform-providers/terraform-provider-aws/blob/eca69aed77eb2f3392d221ec4188d15a5eea6cca/aws/internal/naming/naming.go#L12-L16)) depend on its format, so changing the format should be considered a breaking change in the API and either saved for a major version or I would also throw my hat for this to be adjusted in the manner proposed by https://github.com/hashicorp/terraform-plugin-sdk/pull/508#issuecomment-664133781.

bflad:
Hi @bholzer 👋  Thank you for submitting this. Per https://github.com/hashicorp/terraform-plugin-sdk/pull/508#issuecomment-664133781, https://github.com/hashicorp/terraform-plugin-sdk/pull/508#issuecomment-682726243, and without comment or change of the pull request implementation, this change will not be accepted as-is and will be closed out. If interested, we could certainly look at a separate helper function that achieves the desired changes to not break compatibility with the existing functionality. Given this functionality does not require data or functionality that is internal to the SDK, providers can also implement their own helpers in this case.

PR #512, opened 2020-07-24
Add support for multiple provider factories

paultyng:
Modify runProviderCommand to use true factories and support multiple names and provider instances for reattaching.


helper/resource/plugin.go
```diff
@@ -77,12 +52,59 @@ func runProviderCommand(f func() error, wd *tftest.WorkingDir, opts *plugin.Serv
...
+		if diags.HasError() {
+			return fmt.Errorf("unable to configure provider %q: %w", providerName, diagutils.ErrorDiags(diags))
+		}
+
+		opts := &plugin.ServeOpts{
+			GRPCProviderFunc: func() proto.ProviderServer {
+				return grpcplugin.NewGRPCProviderServer(provider)
+			},
+			Logger: hclog.New(&hclog.LoggerOptions{
+				Name:   "plugintest",
```
paultyng, line 81: I'm not sure if this is output anywhere, maybe it should have the plugin name?

helper/resource/plugin.go
```diff
@@ -77,12 +52,59 @@ func runProviderCommand(f func() error, wd *tftest.WorkingDir, opts *plugin.Serv
...
+		// just foo. So let's fix that.
+		providerName = strings.TrimPrefix(providerName, "terraform-provider-")
+
+		provider, err := factory()
+		if err != nil {
+			return fmt.Errorf("unable to create provider %q from factory: %w", providerName, err)
+		}
+
+		// PT: should this actually be called here? does it not get called by TF itself already?
+		diags := provider.Configure(ctx, terraform.NewResourceConfigRaw(nil))
```
appilon, line 71: Have you tried it without this line? Or are you saying it was necessary to get it working

helper/resource/plugin.go
```diff
@@ -103,7 +125,8 @@ func runProviderCommand(f func() error, wd *tftest.WorkingDir, opts *plugin.Serv
 
 	// wait for the server to actually shut down; it may take a moment for
 	// it to clean up, or whatever.
-	<-closeCh
+	// TODO: add a timeout here?
```
appilon, line 128: Select and race a timeout might be good so acceptance tests don't hang. Make sure to log or error though because this should not happen

helper/resource/testing.go
```diff
@@ -481,32 +493,19 @@ func Test(t testing.T, c TestCase) {
...
+	// Copy any explicitly passed providers to factories, this is for backwards compatibility.
+	if len(c.Providers) > 0 {
+		c.ProviderFactories = map[string]func() (*schema.Provider, error){}
 
-	// Auto-configure all providers.
-	for _, p := range providers {
-		diags := p.Configure(context.Background(), terraform.NewResourceConfigRaw(nil))
-		if diags.HasError() {
-			t.Fatal("error configuring provider: %s", diagutils.ErrorDiags(diags))
+		t.Log("TestCase.Providers is deprecated, please use ProviderFactories")
```
appilon, line 500: I would avoid logging the deprecation. I think we can rely on the godoc deprecatiob

helper/resource/plugin.go
paddycarver: This could happen due to a provider implementation error, however.

paddycarver:
It does appear, as well, that this actually broke `go test` for the `helper/resource` package. Looking into that now.

paddycarver:
Turns out the test failures weren't actually exposing bugs in this, they were exposing bugs in the work this was built on. Moving where the factories produced providers just exposed the bugs, not caused them.

I've pushed a fix that fixes the errors.

helper/resource/plugin.go
paddycarver: (As I noted in a comment on the code, won't the individual test timeout be sufficient for this? Or would we like our own, more aggressive timeout?)

helper/resource/plugin.go
paultyng: I left it since it was already there, but I am unsure why it would be required, maybe @paddycarver knows?

helper/resource/plugin.go
paultyng: That's what I wasn't sure about, go test already times out, but its an aggregate timeout for all tests, not for an individual I believe. It may be useful to add an individual timeout on the `TestCase` or something? Unsure and people have some long tests, so didn't want to cause a need for additional modification to get it to work for those cases that exist.

helper/resource/plugin.go
paddycarver: It seems like it was added in this PR? I don't think it was in the codebase before this... I can't find any reference to it.

paultyng: It came from here: https://github.com/hashicorp/terraform-plugin-sdk/blob/d609e41bce23b1dfd993591a7c3e2d5a63128dbd/helper/resource/testing.go#L505-L511

But I think lets drop it for now.

helper/resource/testing.go
paultyng, line 500:
```suggestion
```

paddycarver:
Let's do it.

PR #513, opened 2020-07-27
Fix a Fatal to a Fatalf when using replacers.

paddycarver:
We were calling t.Fatal but using %s, which was resulting in weird
output. Fix to use Fatalf.


paddycarver:
This isn't needed after #512.

PR #514, opened 2020-07-29
Restore test log splitting.

paddycarver:
Restore the ability to split out logging by test name when using the
acceptance test framework.

Note: I'm not sure if this will have race conditions, given it's called
repeatedly.

### DIFF
--- a/helper/logging/logging.go
+++ b/helper/logging/logging.go
@@ -1,6 +1,7 @@
 package logging
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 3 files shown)


helper/logging/logging.go
```diff
@@ -9,19 +10,21 @@ import (
...
 )
 
 // These are the environmental variables that determine if we log, and if
 // we log whether or not the log should go to a file.
 const (
-	EnvLog     = "TF_LOG"      // Set to True
-	EnvLogFile = "TF_LOG_PATH" // Set to a file
+	EnvLog         = "TF_LOG"      // Set to True
+	EnvLogFile     = "TF_LOG_PATH" // Set to a file
+	EnvLogPathMask = "TF_LOG_PATH_MASK"
```
bflad, line 21: Should we include a comment here about its purpose for future travelers?

```suggestion
	// EnvLogPathMask splits test log files by name.
	EnvLogPathMask = "TF_LOG_PATH_MASK"
```

PR #516, opened 2020-07-30
Allow specifying external providers in a TestCase.

paddycarver:
Users used to specify third-party providers they relied on for
tests--usually utility providers like random--by importing the provider
and specifying it in resource.TestCase.Providers or
resource.TestCase.ProviderFactories. With binary testing, this is no
longer necessary, as the providers can be downloaded from the registry
using Terraform init. This allows for more up-to-date providers with
fewer Go dependencies, which is a nicer system all around.

The acceptance testing framework will run `terraform init`, which will
automatically notice these providers being used in the configuration
files and download them, just like in production.

However, there's a wrinkle; tests that import don't have the resources
set in the configuration, which means they're not downloaded during
init, which causes test failures.

To resolve this, this PR adds a new property to TestCase, which lets
developers list the third party providers their test relies on. This
list is then added to the list of provider blocks auto-generated by the
test framework, allowing them to be inited during import tests.

Arguably, we should have this new property take a map of blocks to
either their version, their source block, or both, so users can control
specifically which version of the provider they're using and where it's
downloaded from. I'm open to thoughts and opinions on that.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -307,7 +307,7 @@ type TestCase struct {
 	// will be automatically retrieved during init. Import doesn't use a
 	// config, however, so we allow manually specifying them here to be
 	// downloaded for import tests.
-	ExternalProviders []string
+	ExternalProviders map[string]ExternalProvider
 
 	// PreventPostDestroyRefresh can be set to true for cases where data sources
 	// are tested alongside real resources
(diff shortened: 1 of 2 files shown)


paddycarver:
Pushed a change to be a map of structs to support versioning and sources. I can't actually discriminate based on 0.13 or not, but providers that want to run tests against 0.12.26+ should be able to just leave the `Source` property blank and it won't show up in the config, which means it should be 0.12 compatible.

helper/resource/testing.go
```diff
@@ -533,13 +548,35 @@ func Test(t testing.T, c TestCase) {
...
 // config is provided, but the providers must be defined.
-func testProviderConfig(c TestCase) string {
+func testProviderConfig(c TestCase) (string, error) {
 	var lines []string
+	var requiredProviders []string
 	for p := range c.Providers {
 		lines = append(lines, fmt.Sprintf("provider %q {}\n", p))
 	}
+	for p, v := range c.ExternalProviders {
+		if _, ok := c.Providers[p]; ok {
```
paultyng, line 558: I think you need to check factories as well here?

paddycarver: Factories don't produce a provider block :man_shrugging:

PR #527, opened 2020-08-10
Backport reattach testing to v1.

paddycarver:
Backport the reattach-based binary acceptance testing mode introduced in
v2 to our v1 branch. It will only be enabled if TF_ACC_REATTACH is set
to 1 _and_ binary acceptance testing is used.


paddycarver:
In _theory_ this _should be_ a backwards-compatible change, but I'd love if reviewers evaluated with an eye to that.

kmoe:
Also, as the test/vet failures indicate, the addition of `Errorf` `%w` ends our support for Go 1.12. This is in accordance with our policy of supporting the last two Go versions (1.14 and 1.13), so you may want to update CircleCI to use a Go 1.13 container.

paddycarver:
I opted to keep 1.12 support for ease of upgrade, because the feature differentiation wasn't that important. Also, the Azure team helpfully verified that this works as expected in their test runs.

PR #544, opened 2020-08-21
add handling for step.Destroy

megan07:
Fixes a few broken tests I saw where `Step.Destroy = true`

A couple questions with this:
1) Can someone confirm my assumption that after doing a destroy, the plan-refresh-plan will always be non-empty? I think that's true, but wanted another opinion on if there would be a reason to continue with the plan-refresh-plan anyway?

2) Is there a place I can test this? (Other than my own provider - because I've tested it there and it seems to work fine)

### DIFF
--- a/helper/resource/testing_new_config.go
+++ b/helper/resource/testing_new_config.go
@@ -55,12 +55,26 @@ func testStepNewConfig(t testing.T, c TestCase, wd *tftest.WorkingDir, step Test
 		if err != nil {
 			return err
 		}
+		if step.Destroy {
+			err = runProviderCommand(t, func() error {
+				return wd.Destroy()
+			}, wd, c.ProviderFactories)
+			if err != nil {
+				return err
+			}
+		}
 		if step.Check != nil {
 			state.IsBinaryDrivenTest = true
 			if err := step.Check(state); err != nil {
 				t.Fatal(err)
 			}
 		}
+	}
+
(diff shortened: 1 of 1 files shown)


paddycarver:
It looks like PreventPostDestroyRefresh was introduced here: https://github.com/hashicorp/terraform/pull/6946. The original refresh behavior was added here: https://github.com/hashicorp/terraform/pull/1757 and the original empty plan checking happened here: https://github.com/hashicorp/terraform/pull/1443

I think I'm going to need to dig into how the context stuff works to figure out why this isn't failing.

paddycarver:
OK, here's my theory:

The `context.Plan`s that were happening before were doing a plan _of the destroy_; that is, it was the same context from start to end, and the context was set in destroy mode. So when it called `context.Plan`, the generated plan was for destroying any resources in state, which obviously there were none. Refresh would then refresh all the resources in state, which obviously there were none, and then plan would plan for the destruction of any resources in state, of which there remain none. This is how we could be running plans after destroying things and getting no errors; they were the only kind of plans (that I know of) that don't care about the config.

This can be emulated in binary testing by using the equivalent of `terraform plan -destroy`, which doesn't appear to exist yet in terraform-plugin-test, but should be easy enough to add.

Now that I understand the old behavior, I think I can emulate it.

paddycarver:
Closing this out as superseded by #563

PR #526, opened 2020-08-08
Capture plan and render for unexpected diffs

appilon:
This restores output similar to V1 of the SDK

Closes #454

PR #550, opened 2020-08-25
helper/resource/testing_new: update import state testing

anGie44:
Closes #532 

`testStepNewImportState` can now return the error (if any) of the underlying method that imports the resource, thus allowing for error handling w/in `runNewTest` when `step.ImportState` is set to `true`. Special error handling is needed here to account for provider acceptance tests that check for expected errors during `ImportState` test steps as illustrated in #532. 

Output of previously failing acceptance test:
```
--- PASS: TestAccAWSKinesisFirehoseDeliveryStream_basic (303.34s)
```

PR #522, opened 2020-08-03
Consider resource provider when locating expected state in ImportState acceptance test case

ewbankkit:
Closes https://github.com/hashicorp/terraform-plugin-sdk/issues/521.

The AWS provider has not quite upgraded to v2.0.0 of the SDK yet so I haven't been able to test the fix against my failing acceptance test case although the same fix against v1.14.0 succeeded.

I see no suitable unit tests to add a test case to.

PR #553, opened 2020-08-26
Backport fixes to reattach testing.

paddycarver:
Backport including step number in error messages, backport fix to use correct state when conducting import tests, and backport better diffs in error messages.

This backports #552, #526, and #524.


paddycarver:
Backported #522 as well.

paddycarver:
I pushed new commits backporting #556, #555, #550, #557, and #563, which I believe brings this inline with all the bug fixes on master.

If the code looks good, I'll merge it, and @mbfrahry has volunteered to run some tests against it, then we can cut a release.

appilon (approved):

PR #560, opened 2020-08-27
Add 'helper/resource' error helpers

ewbankkit:
The AWS Provider uses its [own versions](https://github.com/terraform-providers/terraform-provider-aws/blob/684357c96d9b6729db07daa3475ae161228babbc/aws/utils.go#L46-L54) of these helpers extensively and as the provider restructures to add more functionality in `internal` packages it makes sense to add them to the Plugin SDK.

The name `IsTimeoutErrorWithoutLastError` is just a suggestion (@bflad Any ideas?). The AWS Provider's name `isResourceTimeoutError` is misleading in general as what is being tested is whether it's a `TimeoutError` that doesn't wrap an error from the prior refresh function's invocation.

### DIFF
--- a/helper/resource/error.go
+++ b/helper/resource/error.go
@@ -83,3 +83,11 @@ func (e *TimeoutError) Error() string {
 	return fmt.Sprintf("timeout while waiting for %s%s",
 		expectedState, suffix)
 }
+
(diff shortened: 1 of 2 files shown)


ewbankkit:
After some consideration I think these make more sense to be located in the provider's own repository.

PR #563, opened 2020-09-01
Fix destroy testing.

paddycarver:
We weren't properly accounting for TestStep.Destroy in our new
acceptance testing driver. This commit restores the behavior of v1 with
regards to destroy testing.

It also restores other minor areas where we drifted from the previous
testing behaviors.

It fixes an undiscovered bug where TestStep.ExpectNonEmptyPlan wouldn't
fail if an empty plan was returned.


paddycarver:
I believe this should supersede #544.

paddycarver:
This relies on hashicorp/terraform-plugin-test#37, as well.

kmoe (approved):
hashicorp/terraform-plugin-test#37 is released in terraform-plugin-test v2.1.0.

LGTM pending manual test approval from @megan07.

PR #567, opened 2020-09-07
use wd.SetReattachInfo

kmoe:
Needs https://github.com/hashicorp/terraform-plugin-test/pull/38

Propagating the reattach info is now managed by `tftest.WorkingDir` instead of the shell environment.


paultyng (approved):
`WorkingDir` is specific to each `TestCase`? So there wouldn't ever be a race due to `t.Parallel` or something? Otherwise, LGTM

paddycarver:
Yeah, a new one is created for every `runNewTest`, which is invoked by `resource.Test`, so this should be unique per TestCase and should not be subject to parallelism.

paddycarver (approved):

PR #572, opened 2020-09-10
Drop support for Go 1.12.

paddycarver:
We only support supported versions of Go, so stop testing against Go
1.12 and start testing against Go 1.13, the oldest version of Go we
support.


appilon (changes requested):
Please bump .go-version file

paddycarver:
Updated! Good call, thank you.

appilon (approved):

PR #576, opened 2020-09-11
Fix spurious diff from timeouts.

paddycarver:
When testing and verifying that the imported state matches the applied
state, we're seeing some resources report that the number of elements in
the `timeouts` map don't match between the two states. Apparently, any
timeouts set in the config are added to state during _import_, but
aren't during a regular apply.

To avoid returning this incorrectly as a diff, we're just not going to
verify that.


appilon (approved):
This would only "backfire" if someone was doing test checks on the timeouts (why are these even in state?)

kmoe (approved):

PR #578, opened 2020-09-11
internal/plugintest pkg (formerly terraform-plugin-test)

kmoe:
Include terraform-plugin-test as an internal package, called internal/plugintest. After this is merged I will import relevant issues and PRs from the tftest repository and archive or transfer it.

Remove the `Require*` versions of the Terraform functions, as these are not necessary.

### DIFF
--- a/helper/resource/testing_new.go
+++ b/helper/resource/testing_new.go
@@ -19,8 +19,7 @@ func runPostTestDestroy(t testing.T, c TestCase, wd *plugintest.WorkingDir, fact
 	t.Helper()
 
 	err := runProviderCommand(t, func() error {
-		wd.RequireDestroy(t)
-		return nil
+		return wd.Destroy()
 	}, wd, factories)
 	if err != nil {
 		return err
@@ -44,8 +43,12 @@ func runNewTest(t testing.T, c TestCase, helper *plugintest.Helper) {
 
 	defer func() {
 		var statePreDestroy *terraform.State
-		err := runProviderCommand(t, func() error {
-			statePreDestroy = getState(t, wd)
+		var err error
+		err = runProviderCommand(t, func() error {
+			statePreDestroy, err = getState(t, wd)
+			if err != nil {
+				return err
+			}
 			return nil
 		}, wd, c.ProviderFactories)
 		if err != nil {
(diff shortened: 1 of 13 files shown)


internal/plugintest/plugin.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"os"
+)
+
+// RunningAsPlugin returns true if it detects the usual Terraform plugin
+// detection environment variables, suggesting that the current process is
+// being launched as a plugin server.
+func RunningAsPlugin() bool {
+	const cookieVar = "TF_PLUGIN_MAGIC_COOKIE"
```
paddycarver, line 11: Could we not reference these constants from the `plugin` package?

helper/resource/testing_new.go
```diff
@@ -143,10 +144,13 @@ func runNewTest(t testing.T, c TestCase, helper *tftest.Helper) {
...
 }
 
-func getState(t testing.T, wd *tftest.WorkingDir) *terraform.State {
+func getState(t testing.T, wd *plugintest.WorkingDir) *terraform.State {
 	t.Helper()
 
-	jsonState := wd.RequireState(t)
+	jsonState, err := wd.State()
+	if err != nil {
+		t.Fatalf("Error reading state: %s", err)
```
paddycarver, line 152: I'm ok with merging this as is and splitting this out as a separate task, but I think we should return this error instead of essentially panicking on it, for all the reasons we want to stop using Require: we sometimes want to add decorative text to that error, maybe about test step numbers, maybe about dangling resources, and `t.Fatal` robs us of that opportunity.

kmoe: Yes, will change these

paddycarver (approved):
LGTM, let's do it!

paddycarver:
Ugh, I messed up the branch by resolving conflicts, oops, one sec. Going to try to fix it.

paddycarver:
@kmoe I fixed, and it's merge-ready. I had to drop the `symlinkAuxiliaryProviders` helper in `internal/plugintest/helper.go`, because it wasn't compatible with go-getter v1.5.0, which we need for a security fix. I couldn't find any instance of us using it, and it was internal and unexported, so I considered it dead code. Please let me know if I've done a wrong thing. Otherwise, feel free to hit the merge button if my changes look good to you. :)

PR #579, opened 2020-09-11
Backport #576: fix importstateverify + timeouts.

paddycarver:
Resolve the issue with spurious test failures with ImportStateVerify from resources using timeouts on the v1-maint branch, as well.


appilon (approved):

appilon:
this already exists in V2, and this PR ports it to V1, so I am removing that label. I made the labels so periodically we can look at closed PRs and see if any outstanding things need to he ported forward to V2, or backward to V1. Consequently the system falls apart if we don't remove these labels from the "first" PR once it's been ported. I totally probably have not done that part all the time, so not sure if this system is pointless.

Issue #235, opened 2019-11-05
Test HCL configs isolated in files


radeksimko:
## Problem Statement

The test framework today expects test configuration in the form of a string. This is a convention nearly all official providers comply with.

This approach makes it difficult, or rather impossible to parse test configs outside the context of a test framework, which in turn prevents us from:

 - collecting any data about test infrastructure, e.g. to estimate how much it costs to run all acceptance tests
 - knowing how many and which tests cover specific resource or data source
 - formatting configs via `terraform fmt`
 - upgrading configs via `terraform 0.12upgrade`
 - and otherwise take advantage of any other tooling that users have available to them when dealing with raw HCL

It is technically possible to move these inlined strings into files, but we have no conventions in that area and no API surface helping with this.

## Extra Context

One of the motivating factors behind this design was (probably?) to allow developers to easily generate many different versions of the same config by leveraging common Go tooling such as `fmt.Sprintf` or `text/template`.

It is expected that this requirement isn't going away and we need to find a solution that allows for such templating, e.g. via new 0.12 `templatefile` function or through other means.

## Proposal

Come up with a convention for directory structure on how to structure configs and allow test framework to load valid files from there without providers having to deal with any low-level IO functions.

bflad:
My suggestion in this area would be to allow provider developers to choose their directory/file structure by passing in the relative path, similar to what is possible today with `file()` usage in test configurations. Maybe we can just recommend certain conventions and follow them in official providers. Passing in the path will allow additional use cases, such as `examples/` testing. These provider directories conventionally contain end-to-end multiple resource configurations for practitioners, typically split up by services/functionality and may contain multiple files. e.g.

- https://github.com/terraform-providers/terraform-provider-aws/tree/master/examples
- https://github.com/terraform-providers/terraform-provider-azurerm/tree/master/examples

## Proposals

These are not necessarily mutually exclusive proposals, but presented in a personal preferential order. They are shown with a very similar developer experience to the `Config` handling today.

### Proposal 1: (helper/resource.TestStep).ConfigDirectory

Example developer experience:

```go
// provider/resource_example_thing_test.go

func TestAccExampleThing_basic(t *testing.T) {
	resource.ParallelTest(t, resource.TestCase{
		/* ... */
		Steps: []resource.TestStep{
			{
				ConfigDirectory: "testdata/example_thing/basic", // conventional pathing, for large providers may be split into service directories, etc.
				Check: /* ... */,
			},
		},
	})
[...]

PR #587, opened 2020-09-20
Add support for TestCase.ConfigDir

paultyng:
This is some of the initial work to support live .tf file acc tests (#235).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/hashicorp/errwrap v1.0.0
 	github.com/hashicorp/go-cleanhttp v0.5.1
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
-	github.com/hashicorp/go-getter v1.4.2-0.20200106182914-9813cbd4eb02
+	github.com/hashicorp/go-getter v1.4.2-0.20200106182914-9813cbd4eb02 // indirect
 	github.com/hashicorp/go-hclog v0.9.2
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hashicorp/go-plugin v1.3.0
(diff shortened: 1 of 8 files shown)


paultyng:
Investigating adding this for automated tested of examples in the Unifi provider: https://github.com/paultyng/terraform-provider-unifi/pull/57

paultyng:
cc @bflad, this doesn't have the var support yet, but curious if this was kinda what you were thinking?

bflad:
@paultyng this aligns well with what I was thinking for this. 👍

bflad:
Actually on second thought, I thought we'd be doing per `TestStep` with the `ConfigDir` to align with the current ability to change configurations wholly between steps, but maybe there is good benefit to enforcing more conditional test configurations via variables?

paultyng:
I originally had gone that direction, but this still allows you to specify a config to write. My thought was that the general case would be variable changing step by step instead of configs though in the case of files on disk.

bflad:
I went through a few of the more "hairier" acceptance tests we have in the AWS Provider with regards to more complex configuration changes between steps and generally speaking, I think having this at the `TestCase` level would actually be better in a few ways. 👍  I think it will certainly be clearer for provider developers to understand what's being changed configuration-wise in a single test if its variable driven and may slightly encourage more of the "per attribute" testing that we tend to recommend in some of our larger providers.

bflad:
This is ready for rebase. 😄

In my local testing with it rebased, it looks like we can only use `override.tf` iff `TestCase.ConfigDir` is not declared, otherwise you get errors like the following:

```
    resource_aws_ssm_document_test.go:19: Step 1/2 error: Error running pre-apply refresh: 2020/10/16 20:41:39 [DEBUG] Using modified User-Agent: Terraform/0.12.29 HashiCorp-terraform-exec/0.10.0

        Error: Missing resource to override

          on ../../../../../../../../var/folders/w8/05f3x02n27x72g0mc2jy6_180000gp/T/plugintest282708980/work295587011/config685518503/override.tf line 2:
           2: resource "aws_ssm_document" "test" {

        There is no aws_ssm_document resource named "test". An override file can only
        override a resource block defined in a primary configuration file.
```

paultyng:
Oh yeah, probably needs like an opt-in override mode or something.

PR #591, opened 2020-09-24
Pass pre-destroy state to CheckDestroy as expected by the documentation

enk21:
Currently, the CheckDestroy function is being passed the state after the destroy occurs. In most cases, the resources will be empty so the function can't verify that the resources have been disposed / handled as expected.

I wrote up a discuss [post](https://discuss.hashicorp.com/t/checkdestroy-functionality/14615) here with some more details.

According to the documentation, the CheckDestroy method **‘receives the last known Terraform state as input’**.

The comment in the code says: **'CheckDestroy is called after the resource is finally destroyed to allow the tester to test that the resource is truly gone’**.

The documentation for the CheckDestroy type states: the **‘TestCheckFunc is the callback type used with acceptance tests to check the state of a resource. The state passed in is the latest state known, or in the case of being after a destroy, it is the last known state when it was created’**. 

This PR will pass the pre-destroy state to the runPostTestDestroy function, which will pass it to the CheckDestroy function where the user can verify that the destroy worked as expected.

### DIFF
--- a/helper/resource/testing_new.go
+++ b/helper/resource/testing_new.go
@@ -15,7 +15,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-func runPostTestDestroy(t testing.T, c TestCase, wd *tftest.WorkingDir, factories map[string]func() (*schema.Provider, error)) error {
+func runPostTestDestroy(t testing.T, c TestCase, wd *tftest.WorkingDir, factories map[string]func() (*schema.Provider, error), statePreDestroy *terraform.State) error {
 	t.Helper()
 
 	err := runProviderCommand(t, func() error {
@@ -27,16 +27,7 @@ func runPostTestDestroy(t testing.T, c TestCase, wd *tftest.WorkingDir, factorie
 	}
 
 	if c.CheckDestroy != nil {
-		var statePostDestroy *terraform.State
-		err := runProviderCommand(t, func() error {
-			statePostDestroy = getState(t, wd)
-			return nil
-		}, wd, factories)
-		if err != nil {
-			return err
-		}
-
-		if err := c.CheckDestroy(statePostDestroy); err != nil {
+		if err := c.CheckDestroy(statePreDestroy); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


bflad:
Please note that I can confirm this is a notable regression in the binary testing framework by attempting to run the AWS Provider acceptance testing with v2.0.3 (as compared to our current v2.0.1), since errors are now being returned due to #581 

New test failure (hidden prior to v2.0.3):

```
=== CONT  TestAccAWSSSMDocument_basic
TestAccAWSSSMDocument_basic: testing_new.go:68: Error running post-test destroy, there may be dangling resources: Default error in SSM Document Test
--- FAIL: TestAccAWSSSMDocument_basic (26.80s)
```

Some of our older style acceptance testing includes `CheckDestroy` logic that follows the below coding pattern (which presumably was ensuring that the pre-destroy state included at least one of the expected resource types and verified it was deleted in the API):

```go
func testAccCheckAWSSSMDocumentDestroy(s *terraform.State) error {
	conn := testAccProvider.Meta().(*AWSClient).ssmconn

	for _, rs := range s.RootModule().Resources {
		if rs.Type != "aws_ssm_document" {
			continue
		}

		out, err := conn.DescribeDocument(&ssm.DescribeDocumentInput{
			Name: aws.String(rs.Primary.Attributes["name"]),
		})

		if err != nil {
			// InvalidDocument means it's gone, this is good
			if wserr, ok := err.(awserr.Error); ok && wserr.Code() == ssm.ErrCodeInvalidDocument {
				return nil
			}
			return err
		}

		if out != nil {
			return fmt.Errorf("Expected AWS SSM Document to be gone, but was still found")
		}

		return nil
	}

	return fmt.Errorf("Default error in SSM Document Test")
}
```

In the old testing framework, that error would not trigger. Many of our newer `CheckDestroy` functions `return nil` outside the loop and are hiding this issue.

bflad:
The fix seems correct here as the legacy testing framework after all steps ran the final destroy and followup CheckDestroy using the state from the last step import/config apply:

https://github.com/hashicorp/terraform-plugin-sdk/blob/3f85e93e6fdec5464dbb40c77b58a54e4f8ed32f/helper/resource/testing.go#L677-L679

https://github.com/hashicorp/terraform-plugin-sdk/blob/3f85e93e6fdec5464dbb40c77b58a54e4f8ed32f/helper/resource/testing.go#L753-L764

I can confirm the above test is fixed with this:

```
--- PASS: TestAccAWSSSMDocument_basic (15.86s)
```

kmoe (approved):
LGTM, but please wait for additional review by @paddycarver, who last touched this code.

paddycarver (approved):

paddycarver:
This looks good to me, and reasonable. Thanks @enk21 and @bflad for the investigation and bugfix!

PR #592, opened 2020-09-25
testing: Provider handling of errors, Add ErrorCheck to TestCase

### LINKED ISSUE #568: Give providers option to handle testing errors (e.g., skip tests based on error)
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

We propose giving providers the option to be called back with errors encountered during acceptance testing.

At least for AWS, knowing which services and features are supported in different regions (e.g., `us-east-1`) and partitions (e.g., ISO, GovCloud, China, commercial) is difficult. New services and features may come online in one subsection of AWS but not others. The features and services may or may not ever be added to all areas of AWS. Having tests fail because a service or feature is unavailable is misleading (and leaves dangling resources) and should instead result in a skipped test.

Currently, the PreCheck field is helpful for many situations. However, often it is not possible to assess that a service or feature is unavailable or functions differently in a partition or region _until we get an error._ There are no cost-free API calls to determine the limitation.

### Design

**Design Objective:** Allow providers the option to handle errors.

**Design:** Add an optional field to `TestCase` so that the provider can be called back when an error occurs. The provider can handle the error by 1) skipping or 2) fataling the test, or 3) consuming the error. If the provider does not consume, skip or fatal the error, the provider can hand the error back to plugin SDK to handle as it normally would.

### Example

Route53 in the AWS GovCloud partitions does not support operations related to PublicDNS. There are no cost-free API calls to determine that GovCloud has this limitation. But, running a Route53 acceptance test in GovCloud results in this error. Using this PR, we can skip the test rather than reporting a failure.

How this enhancement might look from the AWS Provider:

```go
func testAccAWSRoute53RecordPublicDNSErrorCheck(err error, t *testing.T) {
	re := regexp.MustCompile(`Operations related to PublicDNS are not supported in this aws partition`)

	if re.MatchString(err.Error()) {
		t.Skipf("skipping test; this partition %s does not support PublicDNS operations", testAccGetPartition())
	}
}

func TestAccAWSRoute53Record_basic(t *testing.T) {
	var record1 route53.ResourceRecordSet
	resourceName := "aws_route53_record.default"

	resource.ParallelTest(t, resource.TestCase{
		PreCheck:      func() { testAccPreCheck(t) },
		ErrorCheck:    func(err error) { testAccAWSRoute53RecordPublicDNSErrorCheck(err, t) },
		IDRefreshName: resourceName,
		Providers:     testAccProviders,
		CheckDestroy:  testAccCheckRoute53RecordDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccRoute53RecordConfig,
				Check: resource.ComposeTestCheckFunc(
					testAccCheckRoute53RecordExists(resourceName, &record1),
				),
			},
		},
	})
}
```

### References

* terraform-providers/terraform-provider-aws#15022

YakDriver:
Closes #568 
Relates terraform-providers/terraform-provider-aws#14188

This PR includes

1. A new `TestCase` field called `ErrorCheck`, which takes an `ErrorCheckFunc` function. The function is called when an error occurs allowing providers the option of handling errors.

See #568 for design considerations and an example of using this PR.

In GovCloud, the output from the AWS provider acceptance test above with this PR:
```
    resource_aws_route53_record_test.go:113: skipping test; this partition aws-us-gov does not support
PublicDNS operations
--- SKIP: TestAccAWSRoute53Record_basic (6.64s)
```

On the standard/commercial partition, the output from the test with this PR (unaffected by this change):
```
--- PASS: TestAccAWSRoute53Record_basic (110.87s)
```


YakDriver:
Since this is intended to be a test-level functionality rather than step-level, I think it would make more sense for the `SkipTestOnError` field to be part of `TestCase` rather than `TestStep`. Thoughts?

bflad:
Super stoked about this!

I'm wondering if we should use a custom Go type for the field value, e.g.

```go
SkipOnError TestSkipOnErrorFunc
// ...
type TestSkipOnErrorFunc func() *regexp.Regexp
// ... or likely better ...
type TestSkipOnErrorFunc func(error) bool
```

Which would allow for providers to more easily inject custom logic? For example in the AWS Provider, we have this helper that we use in a lot of the `PreCheck` functions:

```go
// Check service API call error for reasons to skip acceptance testing
// These include missing API endpoints and unsupported API calls
func testAccPreCheckSkipError(err error) bool {
	// GovCloud has endpoints that respond with (no message provided after the error code):
	// AccessDeniedException:
	// Ignore these API endpoints that exist but are not officially enabled
	if isAWSErr(err, "AccessDeniedException", "") {
		return true
	}
	// Ignore missing API endpoints
	if isAWSErr(err, "RequestError", "send request failed") {
		return true
	}
	// Ignore unsupported API calls
	if isAWSErr(err, "UnknownOperationException", "") {
		return true
	}
	if isAWSErr(err, "UnsupportedOperation", "") {
		return true
	}
	return false
}
```

Being able to use something like that directly:

```go
resource.TestCase{
// ...
	SkipOnError: testAccPreCheckSkipError,
// ...
}
```

Or potentially with some additional helpers to still make the regular expression case easy, e.g.

```go
func SkipOnErrorRegexp(re *regexp.Regexp) TestSkipOnErrorFunc {
  return func(err error) bool {
    if err == nil {
      return false
    }

    return re.MatchString(err.Error())
  }
}
```

And:

```go
resource.TestCase{
// ...
	SkipOnError: acctest.SkipOnErrorRegexp(regexp.MustCompile(`Operations related to PublicDNS are not supported in this aws partition`)),
// ...
}
```

Sounds amazing!

YakDriver:
@bflad I made most of your suggested changes. The one variation is creating `acctest.SkipOnErrorContains()` rather than `acctest.SkipOnErrorRegexp()` to avoid all the `regexp.MustCompile()` boilerplate everywhere.

helper/acctest/skip_on_error.go
```diff
@@ -0,0 +1,20 @@
...
+// SkipOnErrorContains helps skip tests based on a portion of an error message
+// which is matched to an error.
+func SkipOnErrorContains(skipError string) resource.SkipOnErrorFunc {
+	return func(err error) bool {
+		if err == nil {
+			return false
+		}
+
+		re := regexp.MustCompile(skipError)
+		return re.MatchString(err.Error())
```
bflad, line 18: Since this typically will be ran after testing infrastructure is provisioned, a regular expression that does not compile will cause a panic and leave all the dangling infrastructure. This is typically why a lot of the other testing code has uses `regexp.MustCompile()` in the test functions themselves, since it forces a Go build error.

This should either use `strings.Contains()` for safety or move the regexp compilation back into the providers.

helper/acctest/skip_on_error_test.go
```diff
@@ -0,0 +1,25 @@
+package acctest
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestSkipOnError(t *testing.T) {
```
bflad, line 8: Nit: In Go testing, typically unit testing functions are named to match their code equivalent, e.g.

```suggestion
func TestSkipOnErrorContains(t *testing.T) {
```

helper/acctest/skip_on_error_test.go
```diff
@@ -0,0 +1,25 @@
...
+	testCases := []struct {
```
bflad, line 9: These unit test cases should include a negative match. 👍

bflad (changes requested):

helper/acctest/skip_on_error.go
YakDriver: I added helpers and accompanying tests for both `SkipOnErrorContains()` and `SkipOnErrorRegexp()` which are safe.

helper/acctest/skip_on_error_test.go
YakDriver: Of course!

helper/acctest/skip_on_error_test.go
YakDriver: Added!

bflad (approved):
Non-binding approval from me. 👍  Whenever the maintainers can take a look it would be greatly appreciated.

helper/resource/testing_new.go
```diff
@@ -117,6 +117,9 @@ func runNewTest(t testing.T, c TestCase, helper *tftest.Helper) {
 					t.Fatalf("Step %d/%d error running import, expected an error with pattern (%s), no match on: %s", i+1, len(c.Steps), step.ExpectError.String(), err)
 				}
 			} else {
+				if err != nil && c.ErrorCheck != nil {
+					c.ErrorCheck(err)
```
paultyng, line 121: should this have a way to say ignore the error? like a return of `(bool, error)` or something? so this could be an alternative `ExpectError` but let you use things like `errors.Is` / `errors.As` and flag it as ok for the test?

just throwing this out, not sure if it makes sense

YakDriver: I can't foresee a use for that which, of course, means someone will need that functionality within 14ms of merging. I'll make a change.

YakDriver:
@paultyng This is the implementation on the AWS Provider side:

```go
func testAccAWSRoute53RecordPublicDNSErrorCheck(err error, t *testing.T) error {
	re := regexp.MustCompile(`Operations related to PublicDNS are not supported in this aws partition`)

	if re.MatchString(err.Error()) {
		t.Skipf("skipping test; this partition %s does not support PublicDNS operations", testAccGetPartition())
	}
	return err
}

func TestAccAWSRoute53Record_basic(t *testing.T) {
	var record1 route53.ResourceRecordSet
	resourceName := "aws_route53_record.default"

	resource.ParallelTest(t, resource.TestCase{
		PreCheck:      func() { testAccPreCheck(t) },
		ErrorCheck:    func(err error) error { return testAccAWSRoute53RecordPublicDNSErrorCheck(err, t) },
		IDRefreshName: resourceName,
		Providers:     testAccProviders,
		CheckDestroy:  testAccCheckRoute53RecordDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccRoute53RecordConfig,
				Check: resource.ComposeTestCheckFunc(
					testAccCheckRoute53RecordExists(resourceName, &record1),
				),
			},
			{
				ResourceName:            resourceName,
				ImportState:             true,
				ImportStateVerify:       true,
				ImportStateVerifyIgnore: []string{"allow_overwrite", "weight"},
			},
		},
	})
}
```

helper/resource/testing_new.go
YakDriver: Updated

paultyng (approved):
This LGTM, will also need an @paddycarver or @kmoe sign off as well.

paddycarver:
We were _just_ talking about this! If possible, I'd like to wait until early November for some other things to fall in place, but I think we should be able to review and merge this that week, if that timing works out.

kmoe (approved):
Thanks for this. Looks good to me.

PR #596, opened 2020-09-26
Fix warnings around errors

abergmeier:
Use correct function and remove ending exclamation mark.

### DIFF
--- a/helper/resource/testing_new_config.go
+++ b/helper/resource/testing_new_config.go
@@ -1,6 +1,7 @@
 package resource
 
 import (
+	"errors"
 	"fmt"
 
 	tfjson "github.com/hashicorp/terraform-json"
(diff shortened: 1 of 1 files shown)


paddycarver:
Hey @abergmeier, does this fix a behavior or does something not work without this merged that works with it merged?

abergmeier:
It fixes warning you get from Go. So reducing the noise when coding. Wonder though, why your CI doesn't flag that.

paddycarver:
Hmm, as far as I know, Go doesn't issue warnings. Do you mean `golint`? Or `go vet`? What command are you running that's producing the warning?

abergmeier:
It is actually both. IIRC exclamation mark is `golint`, formatting function is `go vet`.

paddycarver (approved):

Issue #534, opened 2020-08-12
Add wrapper functions for validation helper without Diagnostics


pavel-github:
### SDK version

```
github.com/hashicorp/terraform-plugin-sdk/v2
v2.0.0
```

### Use-cases

The field `ValidateFunc` of Schema type is deprecated since v2. Now we should use `ValidateDiagFunc`, but all validation helper (except map helper) use `schema.SchemaValidateFunc` and not `schema.SchemaValidateDiagFunc`.

### Attempted Solutions

What we tried to do in our plugin was create function that wraps existing validation helpers and returns `diag.Diagnostics` instead of errors.

### Proposal

To help with migration from v1 to v2 we could create wrapper functions for every validation helper that return `schema.SchemaValidateDiagFunc` instead of `schema.SchemaValidateFunc`.

birchb1024:
Maybe something like this:
```
				ValidateDiagFunc : validateV1(validation.IsURLWithHTTPorHTTPS),
```

Would this sort of curry work?
```
func validateV1(fn schema.SchemaValidateFunc) (schema.SchemaValidateDiagFunc) {
	return func(v interface{}, path cty.Path) diag.Diagnostics {
		var diags diag.Diagnostics

		warnings, errors := fn(v, fmt.Sprintf("%#v", path))
		for _, w := range warnings {
			diags = append(diags, diag.Diagnostic{
				Severity:      diag.Warning,
				Summary:       w,
				Detail:        w,
				AttributePath: path,
			})
		}
		for _, err := range errors {
			diags = append(diags, diag.Diagnostic{
				Severity:      diag.Warning,
				Summary:       err.Error(),
				Detail:        err.Error(),
				AttributePath: path,
			})
		}
		return diags
	}
}

```

risicle:
Similarly, the `All()` and `Any()` functions only take `schema.SchemaValidateFunc`, which is pretty annoying.

paddycarver:
This was released with version 2.4.0 of the SDK.

PR #611, opened 2020-10-11
helper/validation: Wrapper function for legacy SchemaValidateFunc

gnarlex:
Adds a wrapper function for validation functions of type `SchemaValidateFunc` (https://github.com/hashicorp/terraform-plugin-sdk/issues/534). 📦

It follows a similar implementation as currently used in `schema.validateFunc` ([schema.go#L441-L456](https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/schema/schema.go#L441-L456))to handle legacy validation functions. This implementation parses the last step of given `cty.Path` to be used as the second parameter of the legacy validator.

This would help a lot of projects since no "diag" versions of the validation helpers are implemented yet.

### DIFF
--- a/helper/validation/meta.go
+++ b/helper/validation/meta.go
@@ -2,10 +2,11 @@ package validation
 
 import (
 	"fmt"
+	"reflect"
+
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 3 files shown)


remyleone:
https://discuss.hashicorp.com/t/validatefunc-deprecation-in-terraform-plugin-sdk-v2/12000

paddycarver (approved):
This looks great, thanks for the contribution!

PR #613, opened 2020-10-13
Relax validation of ID attribute

### LINKED ISSUE #607: Relax InternalValidate on the `ID` attribute
As long as the the explicit schema of an `ID` attribute conforms to some specific values (maybe only type of string, unsure), it should be allowed to be overridden. 

There are a few different use cases (document generation, LS, schema JSON), but mostly they stem from the fact that its always set to Optional + Computed even if its not actually optional for resource.

paultyng:
Fixes #607

Currently in the `InternalValidate` validation, we prohibit any explicit overriding of the `id` attribute for resources in favor of writing an explicit version. But this does not allow modifying the `Optional` bit, or setting a custom `Description` which is useful for docs generation and the language server. This allows implementers to override the `id` if they so choose as long as it still adheres to the rules specified for the built-in usage.

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -26,7 +26,6 @@ var ReservedResourceFields = []string{
 	"connection",
 	"count",
 	"depends_on",
-	"id",
 	"lifecycle",
 	"provider",
 	"provisioner",
@@ -649,6 +648,14 @@ func (r *Resource) InternalValidate(topSchemaMap schemaMap, writable bool) error
 			}
 		}
 
+		if f, ok := tsm["id"]; ok {
+			// if there is an explicit ID, validate it...
+			err := validateResourceID(f)
+			if err != nil {
+				return err
+			}
+		}
+
 		for k, f := range tsm {
 			if isReservedResourceFieldName(k, f) {
 				return fmt.Errorf("%s is a reserved field name", k)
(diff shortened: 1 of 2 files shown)


bflad (approved):
Conceptually looks good to me. 😄

radeksimko (commented):
I'd be curious to see how this affects the schemas of top N providers - whether any of them would fail this validation and why.

I remember tweaking the internal validation in the past myself which brought many surprises and lead to some reverting later. 😅 

--- 

I may have missed some recent updates around ID handling, but on what occasion would a resource want to make `id` optional? I'm aware that `id` has lost its special status in core, but aren't there still SDK parts which treat it as special? e.g. we haven't really gotten rid of `ResourceData.Id()` or `ResourceData.SetId()` - this was one of the reasons for the strict validation AFAICT. I guess it's fine if calling `SetId(val)` & `Set("id", val)` and `Id()` & `Get("id")` have the same implications? They [don't seem to though](https://github.com/hashicorp/terraform-plugin-sdk/blob/8dec4902fc67e60ecaa466261cbe892a09a5208e/helper/schema/resource_data.go#L204-L255).

The description use case makes sense to me though.

paultyng:
Resources IDs are _currently_ optional, this is letting you turn it off.

radeksimko:
Ah, that's fair then 👍

radeksimko (approved):

PR #620, opened 2020-10-20
helper/resource: Support Configuration Directories and Variables in Acceptance Testing

### LINKED ISSUE #235: Test HCL configs isolated in files

bflad:
Closes hashicorp/terraform-plugin-testing#72
Rebases and continues https://github.com/hashicorp/terraform-plugin-sdk/pull/587

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -333,6 +333,14 @@ type TestCase struct {
 	IDRefreshName   string
 	IDRefreshIgnore []string
 
+	// ConfigDir is a relative directory path for the base Terraform configuration of the test.
+	//
+	// When running Terraform operations for the test, Terraform will
+	// be executed with copies of the files of this directory as its
+	// working directory. This setting can be combined with
+	// TestStep.Config to override portions of the Terraform
+	// configuration or omitted in preference of only specifying the
+	// Terraform configuration via each TestStep.Config.
 	ConfigDir string
 }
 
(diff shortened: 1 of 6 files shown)


bendbennett:
Superseded by [Add Option to use Terraform Configuration in TestStep](https://github.com/hashicorp/terraform-plugin-testing/pull/153#top).

PR #622, opened 2020-10-22
helper/acctest: Generate 2048 bit long private keys in helper/acctest.RandTLSCert

### LINKED ISSUE #621: acctest.RandTLSCert generated invalid private keys (key with <2048 bits)
### SDK version

```
v2.0.3
```

```go
rCert, rKey, err := acctest.RandTLSCert("myorg")
	rInt := acctest.RandInt()
	if err != nil {
		t.Fatal(err)
	}
```

```hcl
-
```

### Expected Behavior
The acctest generate private keys with 1024 bits (hardcoded -->https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/acctest/random.go#L145). The generated certificate is invalid for most distributions (as a sample Google Chrome requires TLS certificated with a private key >= 2048 since 2014.)

We use the function within our test for the Hetzner Cloud Load Balancer Feature and we noticed that the Load Balancer instance on the other end will never come up because of the invalid cert.

### Actual Behavior
Short Solution: Increase the bits to 2048
Long Term: Make the bits configurable (this might be a breaking change)

LKaemmerling:
Generate 2048 bit long private keys in helper/acctest.RandTLSCert instead of 1024 bit long private keys.

This closes #621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ FEATURES
 * Ported TypeSet test check funcs essential for migrating to V2 of the SDK [GH-614]
 * Improved debug output for how to manually invoke the Terraform CLI [GH-615]
 
+ENHANCEMENTS
+* Generate 2048 bit long private keys with `helper/acctest.RandTLSCert` [GH-621]
+
 # 2.0.4 (October 06, 2020)
 
 BUG FIXES
--- a/helper/acctest/random.go
+++ b/helper/acctest/random.go
@@ -142,7 +142,7 @@ func RandIpAddress(s string) (string, error) {
 }
 
 func genPrivateKey() (*rsa.PrivateKey, string, error) {
-	privateKey, err := rsa.GenerateKey(crand.Reader, 1024)
+	privateKey, err := rsa.GenerateKey(crand.Reader, 2048)
 	if err != nil {
 		return nil, "", err
 	}


bflad:
Hi @LKaemmerling 👋 Thank you for submitting this change and apologies for the delayed response here. Unfortunately, we cannot merge a change like this as-is because it could represent a breaking change for systems or testing that may be explicitly expecting the lower bit count. Similar to the response in the associated issue, its not likely we would accept a different form of the change either, therefore I'm going to close this submission not because its security concerns aren't valid, but because these helpers are unfortunately a source of potential maintenance burden.

> At some point since this was raised, we began documenting the cryptographic specifics of the similar [`acctest.RandSSHKeyPair()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest#RandSSHKeyPair) and recommending that testing with different or stronger security requirements should use code created outside this codebase. Similar documentation would apply for [`acctest.RandTLSCert()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest#RandTLSCert). Security requirements are constantly evolving and it is untenable for this Go module to maintain all possible permutations required for cryptographic logic needed across the entire provider ecosystem without constant maintenance or inevitably needing to introduce breaking changes to the exported Go APIs. Future major versions of this code would likely remove helpers like these for that reason.

Thanks again though!

PR #639, opened 2020-11-06
Update go-plugin, and use hclog.Logger when serving a plugin

jbardin:
Update go-plugin, so that it no longer overrides the default output in the `log` package.

Go-plugin can transfer JSON structured hclog output to a client logger. Rather than requiring existing providers to update all logging calls, we inject a wrapper into the std log package output which translates the defacto log format into structured log output with the correct levels. This will allow terraform to correctly filter the output, rather than default to handling all logs as Debug. This allows the logger to remove redundant tags and timestamps in the log output, and fixes multi-line log output where the first line may have a tag other than `[DEBUG]`, but the remainder of the lines will be converted to `[DEBUG]`.

This PR intends to only configure the logging when running as a plugin. The same logger could also be adapted for testing, but that is to be left for another PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,9 +18,9 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.1
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/go-getter v1.5.0 // indirect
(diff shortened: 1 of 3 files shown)


paddycarver (approved):
Let's do it. Thanks @jbardin. I think we're going to want to follow up in the testing framework, but for the moment, we can ship this incremental improvement.

PR #648, opened 2020-11-27
Avoid panics when testing nested sets of varying depth

IronSavior:
I encountered a panic when I have a hierarchical attribute where the depth of the hierarchy isn't necessarily the same for all  objects. The panic is caused by an out of bounds slice index in cases where you specify an attribute path greater than or equal to the deepest structure in state when using `TestCheckTypeSetElemNestedAttrs()`.

example - Let's say there are groups that can have sub-groups. Your state contains 4 groups and three of them are in a chain and the other has no sub-groups:

```
group1 <- group 2 <- group3

group4
```

When a unit test tries to match `TestCheckTypeSetElemNestedAttrs("groups.*.groups.*.groups.*", map[string]string{"name": expectedName})`, it will panic because `groups.*.groups.*.groups.*` causes the check to index too far into group4.

### DIFF
--- a/helper/resource/testing_sets.go
+++ b/helper/resource/testing_sets.go
@@ -214,7 +214,7 @@ func testCheckTypeSetElemNestedAttrsInState(is *terraform.InstanceState, attrPar
 		// a Set/List item with nested attrs would have a flatmap address of
 		// at least length 3
 		// foo.0.name = "bar"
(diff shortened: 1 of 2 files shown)


paddycarver (approved):
I think I'm following, and this seems reasonable enough to me. Thanks!

PR #656, opened 2020-12-09
Fix large number issues in upgrade state

paultyng:
All usages of plain `json.Unmarshal` should probably be replaced by this, but I'm hesitant to do with without some deeper testing.


paddycarver:
Superseded by #662.

PR #662, opened 2020-12-18
Add a flag for using json.Number from state

paddycarver:
Add a flag to schema.Resource that will parse numbers in state to json.Number. This should have no impact on testing, which uses a string representation of the number, anyways, or on MigrateStateFuncs, which similarly use a string representation. When the flag is flipped, however, the StateUpgraders will need to be updated to use json.Number instead of float64 in their type assertions. This is why this is off by default.

Turning this on, however, will prevent errors about plans not matching applies when using sufficiently large integers with Terraform. See #655 for more details on the problem that can arise from keeping float64s.

Most providers, especially on resources without StateUpgraders, should toggle the flag on to avoid these kinds of problems before they start.


paddycarver:
@kmoe if you want to cut (or are okay with me cutting) a release of terraform-exec, we can use an actual released version of it in this PR.

paultyng:
Why make this opt in for state upgraders that are just broken until it's set?

kmoe (approved):
terraform-exec v0.12.0 is released.

I think an opt-in `UseJSONNumber` is appropriate as a quick fix for providers without a breaking change to the SDK.

paddycarver:
> Why make this opt in for state upgraders that are just broken until it's set?

They're only broken _in theory_. In practice, many of them work, and continue to work with float64 because they're not using large enough numbers to matter. Not using this flag would require a code change before their provider wouldn't panic at runtime, which feels like a breaking change to me, and that seems like overkill if we can avoid it.

paddycarver:
We're going to run with this as-is, to offer an immediate workaround for provider devs. In January, we'll look into turning it on by default for any provider that doesn't use StateUpgraders, as research indicates that may be safe and non-breaking.

Issue #661, opened 2020-12-17
Fix logging in the helper/resource test framework


paddycarver:
### SDK version

```
v2.3.0
```

### Description

We should update the helper/resource test framework to handle logs the same way that the main plugin.Serve function does, so we can have consistent behavior in tests and in prod now.

### References

* #639 
* #543

PR #663, opened 2020-12-18
Disable checkpoint during tests.

### LINKED ISSUE #640: CHECKPOINT_DISABLE should be set for integration test runs
### SDK version

```
{
  "Path": "github.com/hashicorp/terraform-plugin-sdk/v2",
  "Version": "v2.0.3"
}
```

None.

None. 

### Debug Output

```
=== RUN   TestAccARedactedResourceType
     cannot run Terraform provider tests: unexpected error: Get "https://checkpoint-api.hashicorp.com/v1/check/terraform?arch=amd64&os=linux&signature=&version=": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

### Expected Behavior

The Terraform binary should not reach out to Checkpoint during integration test runs.

### Actual Behavior

The Terraform binary should not reach out to Checkpoint during integration test runs and caused a spurious test failure because of network connectivity to Checkpoint or the Checkpoint service itself (it's not clear which is the root cause here).

This could be fixed by ensuring that the integration testing framework passes `CHECKPOINT_DISABLE=<some non-empty value>` via the environment to the Terraform binary being run.

### Steps to Reproduce

Run any integration test with traffic logging turned on.

paddycarver:
We don't need to reach out to checkpoint when running tests, so disable
that information. Fixes #640.

Also remove our logging override, as we've updated to a version of
go-plugin that doesn't require that anymore. See also #661.

### DIFF
--- a/helper/resource/plugin.go
+++ b/helper/resource/plugin.go
@@ -36,6 +36,9 @@ func runProviderCommand(t testing.T, f func() error, wd *plugintest.WorkingDir, 
 	// plugins.
 	os.Setenv("PLUGIN_PROTOCOL_VERSIONS", "5")
 
+	// Terraform doesn't need to reach out to Checkpoint during testing.
+	wd.Setenv("CHECKPOINT_DISABLE", "1")
+
 	// Terraform 0.12.X and 0.13.X+ treat namespaceless providers
 	// differently in terms of what namespace they default to. So we're
 	// going to set both variations, as we don't know which version of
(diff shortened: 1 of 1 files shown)


kmoe (approved):
Whether the SDK needs to reach out to Checkpoint during tests depends on the following environment variables:

- `TF_ACC_TERRAFORM_VERSION`
- `TF_ACC_TERRAFORM_PATH`

If neither of these is set, and if there is no version of `terraform` on the system `$PATH`, `plugintest` will fall back to installing the latest available version of Terraform via `tfinstall.LatestVersion`, with `forceCheckpoint` set to `true` (see https://github.com/hashicorp/terraform-plugin-sdk/blob/master/internal/plugintest/config.go#L42). This means the value of `CHECKPOINT_DISABLE` is ignored, so the latest version of Terraform can always be found.

PR #665, opened 2020-12-18
Drop logging override.

paddycarver:
Missed a logging override in #663. This should fix the last of the race
condition where logging can get written to stderr unexpectedly during
tests. We no longer need to do this because go-plugin was fixed and we
upgraded in #639. We don't want to overwrite that fix in tests.


paultyng (approved):

paddycarver:
Fixes #661. Fixes #543.

PR #666, opened 2020-12-18
Stop overriding log output in plugin.Serve in tests

paddycarver:
When testing, we don't want to override where log output gets sent--the
test framework takes care of that itself. go-plugin has disabled its use
of log.SetOutput for us, but #639 introduced a log.SetOutput in
plugin.Serve. This adds a plugin.ServeOpt property to disable that
log.SetOutput call so we can finally have tests where log lines don't
randomly show up in test output.


bflad (approved):
Looks good locally!

PR #671, opened 2021-01-04
Use better command for module version detection.

### LINKED ISSUE #641: Use "go list -m" to get user SDK version in issue template
Currently the following command is suggested to get version of SDK one uses:
```sh
go mod edit -json | jq '.Require[] | select(.Path=="github.com/hashicorp/terraform-plugin-sdk/v2")'
```

However, this requires `jq` binary and may not be accurate, if one use `replace` option in `go.mod`.

I think asking users to run the command below would be better, as it will show which version of SDK is used for building
```sh
go list -m github.com/hashicorp/terraform-plugin-sdk/v2
```

Sample output:
```console
$ go list -m github.com/hashicorp/terraform-plugin-sdk/v2
github.com/hashicorp/terraform-plugin-sdk/v2 v2.2.0
```

See https://golang.org/ref/mod#go-list-m for more details.

paddycarver:
Update the issue templates to show a command that retrieves the current
version of the module without relying on jq or any other tooling beyond
the `go` command.

Fixes #641.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,7 +8,7 @@ labels: bug
 <!---
 Inspect your go.mod as below to find the version, and paste the result between the ``` marks below.
 
-go list -m github.com/hashicorp/terraform-plugin-sdk/v2
+go list -m github.com/hashicorp/terraform-plugin-sdk/...
 
 If you are not running the latest version of the SDK, please try upgrading
 because your bug may have already been fixed.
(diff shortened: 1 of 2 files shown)


invidian (approved):
Cool!

bflad (approved):
Looks good to me, although you may be able to replace `/v2` with `/...` to make it version agnostic. 👍 

```console
$ go list -m github.com/hashicorp/terraform-plugin-sdk/...
github.com/hashicorp/terraform-plugin-sdk/v2 v2.4.0
```

paddycarver:
Magic! I like that a lot better, and have verified it works on v1. Let's do that!

kmoe (approved):

Issue #456, opened 2020-05-14
Diagnostic attribute path not surfaced in error message


jta:
### SDK version
```
{
  "Path": "github.com/hashicorp/terraform-plugin-sdk/v2",
  "Version": "v2.0.0-rc.1"
}
```

### Relevant provider source code

Resource with the following schema:
```go
        Schema: map[string]*schema.Schema{
            "nested": {
                Type:     schema.TypeList,
                Required: true,
                Elem: &schema.Resource{
                    Schema: map[string]*schema.Schema{
                        "attr": {
                            Type:     schema.TypeString,
                            Required: true,
                        },
                    },
                },
            },
       }, 
```

### Terraform Configuration Files

```hcl
resource "custom" "test" {
  // "nested" is required, but not provided
}
```

### Expected Behavior

I would expect the error message to tell me `nested` is missing, similar to the message i would have gotten if the type were `schema.TypeString`:

```
Error: Missing required argument

  on main.tf line 12, in resource "custom" "test":
  12: resource "custom" "test" {

The argument "nested" is required, but no definition was found.
```

### Actual Behavior

I do get an error, but there is no indication as to what attribute is missing:

```
Error: Required attribute is not set

  on main.tf line 12, in resource "custom" "test":
  12: resource "custom" "test" {
```

AFAICT, the diagnostic attribute path is correctly serialized to proto.Diagnostic, so this may be a case of terraform itself not surfacing the information, but for backward compatibility it would be useful to encode the path information in diagnostic detail field, even if it is potentially redundant.

appilon:
Hello @jta ,

Terraform unfortunately does not have normal "required" support for "block types" such as `TypeList`. The SDK though should be "faking" it by setting `MinItems` to 1 automatically. Could you please run [`terraform providers schema`](https://www.terraform.io/docs/commands/providers/schema.html) in an example configuration using this provider and attributes, and then post the results?

bflad:
Related: #535

@appilon here's the `terraform providers schema -json` output from that issue: https://github.com/hashicorp/terraform-plugin-sdk/issues/535#issuecomment-682750157

paddycarver:
As this should [now be fixed](https://github.com/hashicorp/terraform-plugin-sdk/issues/535#issuecomment-755263375) I'm going to close this issue out.

Feel free to open a new issue if this is still occurring.

PR #672, opened 2021-01-06
Surface the required block that's missing in the error.

### LINKED ISSUE #535: Required Configuration Blocks Not Providing Which Block Is Missing in Error
### SDK version

```
v2.0.1
```

```go
func resourceAwsEksNodeGroup() *schema.Resource {
	return &schema.Resource{
		/* ... */
			/* ... */
			"scaling_config": {
				Type:     schema.TypeList,
				Required: true,
				MaxItems: 1,
				Elem: &schema.Resource{
						"desired_size": {
							Type:         schema.TypeInt,
							Required:     true,
							ValidateFunc: validation.IntAtLeast(1),
						},
						"max_size": {
							Type:         schema.TypeInt,
							Required:     true,
							ValidateFunc: validation.IntAtLeast(1),
						},
						"min_size": {
							Type:         schema.TypeInt,
							Required:     true,
							ValidateFunc: validation.IntAtLeast(1),
						},
					},
				},
			},
		},
	}
}
```

```hcl
resource "aws_eks_node_group" "test" {
  # ... other configuration except scaling_config ...
}
```

### Debug Output

Please ask if you need this.

### Expected Behavior

The acceptance testing framework should output which attribute name is missing, when reporting the `Required attribute is not set` error.

### Actual Behavior

```
=== CONT  TestAccAWSEksNodeGroup_LaunchTemplate_Id
    TestAccAWSEksNodeGroup_LaunchTemplate_Id: resource_aws_eks_node_group_test.go:312: Step 1/3 error: terraform failed: exit status 1

        stderr:

        Error: Required attribute is not set

          on ../../../../../../../../var/folders/w8/05f3x02n27x72g0mc2jy6_180000gp/T/tftest038613874/work421440041/config093175149/terraform_plugin_test.tf line 171, in resource "aws_eks_node_group" "test":
         171: resource "aws_eks_node_group" "test" {


--- FAIL: TestAccAWSEksNodeGroup_LaunchTemplate_Id (2.09s)
```

### Steps to Reproduce

1. Find a resource with a `Required: true` attribute
1. Find a test for that resource and remove the required attribute from the configuration
1. Run acceptance test

paddycarver:
Our current diagnostic only surfaces that a required argument was not
set, but doesn't actually tell you which one. This updates the error
message to match the error returned by core, which includes the name of
the field that's missing.

Why do we return an error if core returns an error? Because core returns
the error only for required _attributes_, because as far as core is
concerned, there is no such thing as required blocks. Only attributes
can have flags, and "required" is a flag, so there can't be "required"
blocks.

But historically blocks have been able to be required, so the SDK fakes
it by doing provider-side validation on the block. But the error message
was less than helpful. This fixes the error message to be more
actionable.

Fixes #535.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -1447,7 +1447,8 @@ func (m schemaMap) validate(
 		if schema.Required {
 			return append(diags, diag.Diagnostic{
 				Severity:      diag.Error,
-				Summary:       "Required attribute is not set",
+				Summary:       "Missing required argument",
+				Detail:        fmt.Sprintf("The argument %q is required, but no definition was found.", k),
 				AttributePath: path,
 			})
 		}
@@ -1458,7 +1459,8 @@ func (m schemaMap) validate(
 		// This is a computed-only field
 		return append(diags, diag.Diagnostic{
 			Severity:      diag.Error,
-			Summary:       "Computed attribute cannot be set",
+			Summary:       "Computed attributes cannot be set",
+			Detail:        fmt.Sprintf("Computed attributes cannot be set, but a value was set for %q.", k),
 			AttributePath: path,
 		})
 	}


paddycarver:
Fixes #456.

kmoe (approved):

Issue #548, opened 2020-08-24
Problems using CustomizeDiff with calculated fields


yoink00:
### SDK version

```
{
  "Path": "github.com/hashicorp/terraform-plugin-sdk/v2",
  "Version": "v2.0.1"
}
```

### Relevant provider source code

```go
		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
			old, new := d.GetChange("network_interface")

			oldList := old.([]interface{})
			newList := new.([]interface{})

			// If the new list is shorter than the old list we
			// want to force new.
			if len(newList) < len(oldList) {
				d.ForceNew("network_interface")
				return nil
			}

			for i, oi := range oldList {
				oiB := oi.(map[string]interface{})
				niB := newList[i].(map[string]interface{})

				if oiB["type"] != niB["type"] {
					d.ForceNew(fmt.Sprintf("network_interface.%d.type", i))
				}
				if oiB["ip_address_family"] != niB["ip_address_family"] {
					d.ForceNew(fmt.Sprintf("network_interface.%d.ip_address_family", i))
				}
				if oiB["network"] != niB["network"] {
					d.ForceNew(fmt.Sprintf("network_interface.%d.network", i))
				}
			}

			return nil
		},
```

#### Initial Configuration

```hcl
resource "upcloud_server" "my-server" {
  zone     = "fi-hel1"
  hostname = "debian.example.com"
  plan     = "2xCPU-4GB"

  storage_devices {
    action = "create"
    size   = 10
    tier   = "maxiops"
  }

  network_interface {
    type = "utility"

  }

  network_interface {
    type = "private"

    network = upcloud_network.test_network_1.id
  }

}

resource "upcloud_network" "test_network_1" {
  name = "test_network_1"
  zone = "fi-hel1"

  ip_network {
    address            = "10.0.34.0/24"
    dhcp               = true
    dhcp_default_route = false
    family             = "IPv4"
    gateway            = "10.0.34.1"
  }
}
```

#### Updated Configuration

```hcl
resource "upcloud_server" "my-server" {
  zone     = "fi-hel1"
  hostname = "debian.example.com"
  plan     = "2xCPU-4GB"

  storage_devices {
    action = "create"
    size   = 10
    tier   = "maxiops"
  }

  network_interface {
    type = "utility"

  }

  network_interface {
    type = "private"

    network = upcloud_network.test_network_11.id
  }

}

[...]

PR #686, opened 2021-01-12
Don't crash when populating NewExtra.

paddycarver:
When populating NewExtra, we iterate over attributes and inspect their
NewExtra property. In the case of null attributes, this results in a
crash, as we're dereferencing a nil pointer.

I'm still not clear on why this is happening, where it's coming from, or
what it means. The current understanding is it has something to do with
usage of CustomizeDiff. I'm not sure why yet. But this commit at least
surfaces which attribute was null, by logging it, and prevents the
crash, even if it doesn't resolve the problem.

See #548, hashicorp/terraform-provider-google#7934

### DIFF
--- a/helper/schema/grpc_provider.go
+++ b/helper/schema/grpc_provider.go
@@ -796,6 +796,10 @@ func (s *GRPCProviderServer) PlanResourceChange(ctx context.Context, req *tfprot
 	newExtra := map[string]interface{}{}
 
 	for k, v := range diff.Attributes {
+		if v == nil {
+			log.Printf("[WARN] Field %q was null, not modifying its NewExtra", k)
+			continue
+		}
 		if v.NewExtra != nil {
 			newExtra[k] = v.NewExtra
 		}
--- a/helper/schema/grpc_provider.go
+++ b/helper/schema/grpc_provider.go
@@ -796,10 +796,7 @@ func (s *GRPCProviderServer) PlanResourceChange(ctx context.Context, req *tfprot
 	newExtra := map[string]interface{}{}
 
 	for k, v := range diff.Attributes {
-		if v == nil {
-			log.Printf("[WARN] Field %q was null, not modifying its NewExtra", k)
-			continue
-		}
+		log.Printf("[TRACE] tpg-7934: copying over attribute %q", k)
 		if v.NewExtra != nil {
 			newExtra[k] = v.NewExtra
 		}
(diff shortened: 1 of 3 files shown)


kmoe (approved):
This seems wise - my only concern is that by preventing the crash without understanding why it occurred, we might be allowing incorrect behaviour. Would it be better to log and still crash?

Approving anyway because based on https://github.com/hashicorp/terraform-provider-google/issues/7934, the risk of incorrect behaviour seems small when compared to the effect of the crash on users right now.

paddycarver:
@kmoe I successfully got to the bottom of the problem, and a customer that could reproduce it got me logs (see 99e4a7064bebab23118ff7726b37c544f5ecf33b) that caught the bug red-handed and confirmed that `helper/schema.Schema.finalizeDiff` was returning a nil `*terraform.ResourceAttrDiff`. That was being put directly into the `helper/schema.ResourceDiff`, which our `PlanResourceChange` implementation didn't expect. The properties of the `*terraform.ResourceAttrDiff` were being accessed without checking if the `*terraform.ResourceAttrDiff` was nil, and that caused the crash.

I could've updated our usage of `helper/schema.ResourceDiff` to check for a nil `*terraform.ResourceAttrDiff` before using it, but given that `helper/schema.ResourceDiff` is exported, I felt it may be used in an unknown number of places, and so we would continue to play whack-a-mole in trying to squash every instance of this bug.

Instead, I opted to (in 570e12852780ca80837d41ed384b5a47340f124c) update every usage of `helper/schema.Schema.finalizeDiff`, which can only be used in the `helper/schema` package, to check the return value before using it. If `helper/schema.Schema.finalizeDiff` returns `nil`, it just won't be included in the `helper/schema.ResourceDiff`, which fixes the bug for _every_ usage of `helper/schema.ResourceDiff`. Or, at least, it should, unless something else is putting a `nil` value in there, in which case, I still think we may want to fix that at the source.

This took so long to come up in such an old and well-used resource because it only triggers when there's _technically_ a diff and `helper/schema.Schema.finalizeDiff` decides that the diff is not a "real" diff and should be ignored. To make this true, the following situation needed to happen:

1. The provider needed to use CustomizeDiff.
2. The CustomizeDiff needed to update a field that was computed.
3. The field needed to be a string
4. The field needed to _not_ be the child of _another_ field that CustomizeDiff updated
5. The field needed a value set in state.
6. The field needed `""` to be set _explicitly_ in the config.

When those conditions are met, `helper/schema.Schema.finalizeDiff` interprets the `""` as "unset" and cancels out the diff, returning a `nil` `*terraform.ResourceAttrDiff`. According to a code comment, this is to preserve the legacy behavior of interpreting `""` as "unset", which for computed fields means "use what I have in state", meaning no diff.

I added a log line to explicitly call out when these conditions are met and we're squashing a diff for legacy reasons.

Mind taking another look?

kmoe (approved):
Approved pending test

PR #692, opened 2021-01-30
Cache ObjectSpec on Block

therve:
This seems to be a big performance hit on complex schema. We observed that on the Datadog provider, where our dashboard resource has a really big schema. This shows up in usage and easily in CI.

Before: 183 tests in 733.311s
After: 183 tests in 94.616s

This patch is against v1 as that's what we use currently, but I believe the same approach can be used in v2, and with the same impact.

This basically a backport to v1 of https://github.com/hashicorp/terraform/pull/26577. The interesting part is that it's in the SDK, not in terraform. I'm looking out to see if the fix is still applicable to v2.

### DIFF
--- a/internal/configs/configschema/decoder_spec.go
+++ b/internal/configs/configschema/decoder_spec.go
@@ -18,6 +18,10 @@ func (b *Block) DecoderSpec() hcldec.Spec {
 		return ret
 	}
 
+	if b.Spec != nil {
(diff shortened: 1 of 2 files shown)


paultyng:
@therve per our discussion going to close this on v1. Its possible its not applicable in v2 due to the binary testing model but I'm not sure.

PR #694, opened 2021-02-05
Make binary testing compatible with Terraform 0.15

### LINKED ISSUE #697: Acceptance Testing Not Compatible After Terraform 0.15.0-alpha20210127
_This is a known issue and is only submitted for discoverability._

### SDK version

```
v2.4.2
```

Any existing testing code. 😄 

### Expected Behavior

Acceptance testing to continue working as with prior versions of Terraform CLI.

### Actual Behavior

Test failures without workarounds, such as:

```
=== CONT  TestAccAWSAcmCertificateDataSource_multipleIssued
          data_source_aws_acm_certificate_test.go:110: Error running init: Usage: terraform init [options] [DIR]
              
                Initialize a new or existing Terraform working directory by creating
                initial files, loading any remote state, downloading modules, etc.
              
                This is the first command that should be run for any new or existing
                Terraform configuration per machine. This sets up all the local data
                necessary to run Terraform that is typically not committed to version
                control.
              
                This command is always safe to run multiple times. Though subsequent runs
                may give errors, this command will never delete your configuration or
                state. Even so, if you have important information, please back it up prior
                to running this command, just in case.
              
                If no arguments are given, the configuration in this working directory
                is initialized.
              
              Options:
              
                -backend=true        Configure the backend for this configuration.
              
                -backend-config=path This can be either a path to an HCL file with key/value
                                     assignments (same format as terraform.tfvars) or a
                                     'key=value' format. This is merged with what is in the
                                     configuration file. This can be specified multiple
                                     times. The backend type must be in the configuration
                                     itself.
              
                -force-copy          Suppress prompts about copying state data. This is
                                     equivalent to providing a "yes" to all confirmation
                                     prompts.
              
                -from-module=SOURCE  Copy the contents of the given module into the target
                                     directory before initialization.
              
                -get=true            Download any modules for this configuration.
              
                -input=true          Ask for input if necessary. If false, will error if
                                     input was required.
              
                -no-color            If specified, output won't contain any color.
              
                -plugin-dir          Directory containing plugin binaries. This overrides all
                                     default search paths for plugins, and prevents the
                                     automatic installation of plugins. This flag can be used
                                     multiple times.
              
                -reconfigure         Reconfigure the backend, ignoring any saved
                                     configuration.
              
                -upgrade=false       If installing modules (-get) or plugins, ignore
                                     previously-downloaded objects and install the
                                     latest version allowed within configured constraints.
      --- FAIL: TestAccAWSAcmCertificateDataSource_multipleIssued (2.43s)
```

### Steps to Reproduce

1. `TF_ACC=1 TF_ACC_TERRAFORM_PATH=/path/to/terraform0.15.0-alpha20210127 go test -v ./aws`

### References

- Initial Terraform CLI breaking change release: https://github.com/hashicorp/terraform/releases/tag/v0.15.0-alpha20210127
- Additional Terraform CLI breaking changes: https://github.com/hashicorp/terraform/pull/27664
- terraform-exec fixes (v.0.13.0): https://github.com/hashicorp/terraform-exec/pull/120
- Potential terraform-plugin-sdk fixes: https://github.com/hashicorp/terraform-plugin-sdk/pull/694

kmoe:
Terraform 0.15 deprecates a number of CLI arguments, including the `[DIR]` positional argument used in `terraform init` and `terraform apply`. The binary test framework made use of these arguments as a way of working around plugin source limitations in v1. This workaround is no longer needed.

The main difference here is that previously we would instantiate one `tfexec.Terraform` per test config file, but `wd.workDir` set as the working directory. Since we no longer have to symlink the plugin binary itself in `wd.workDir`, we can now just use `wd.configDir` as the Terraform working dir instead.

Provider source files (testdata in directories alongside test source files) must still be symlinked into the Terraform working directory. Since this is now `wd.configDir`, not `wd.workDir`, and each `wd.workDir` has several `wd.configDirs`, this symlinking must unfortunately be carried out more often. Performance should not be impacted as we should still be symlinking only a small number of directories.

Example test temp directory structure before this change:

```
/tmp/plugintest594266265
└── work819712804
    ├── config403653211
    │   └── terraform_plugin_test.tf
    ├── config502955059
    │   └── terraform_plugin_test.tf
    ├── terraform.tfstate
    ├── testdata -> /home/katy/dev/go/src/github.com/hashicorp/terraform-provider-random/internal/provider/testdata
    └── tfplan

```

Example test temp directory structure after this change:
```
/tmp/plugintest099393048
└── work460918423
    ├── config243863818
    │   ├── terraform_plugin_test.tf
    │   └── testdata -> /home/katy/dev/go/src/github.com/hashicorp/terraform-provider-random/internal/provider/testdata
    └── config848766255
        ├── terraform_plugin_test.tf
        ├── terraform.tfstate
        ├── testdata -> /home/katy/dev/go/src/github.com/hashicorp/terraform-provider-random/internal/provider/testdata
        └── tfplan
```

## Update following 515c54261

In order to reinstate the automatic persistence of `terraform.tfstate` between test steps, 515c54261 simplifies the test temp directory structure by removing config dirs altogether. The structure is now:

```
/tmp/plugintest594266265
└── work819712804
    ├── terraform_plugin_test.tf
    ├── terraform.tfstate
    ├── testdata -> /home/katy/dev/go/src/github.com/hashicorp/terraform-provider-random/internal/provider/testdata
    └── tfplan
```
where `terraform_plugin_test.tf` gets overwritten each time a new test config is set.

## Testing

Tested this change locally with the following providers:
 - `terraform-provider-random`
 - `terraform-provider-archive`
 - `terraform-provider-dns`

and the following Terraform versions:
 - 0.14.5
 - `master` as of 2nd February 2021 (c8f83e184b67a5f79255b2fffb86524c6c8ef811)

fixes #697

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/hashicorp/go-version v1.2.1
 	github.com/hashicorp/hcl/v2 v2.3.0
 	github.com/hashicorp/logutils v1.0.0
-	github.com/hashicorp/terraform-exec v0.12.0
+	github.com/hashicorp/terraform-exec v0.13.0
 	github.com/hashicorp/terraform-json v0.8.0
 	github.com/hashicorp/terraform-plugin-go v0.2.1
 	github.com/hashicorp/yamux v0.0.0-20181012175058-2f1d1f20f75d // indirect
--- a/go.sum
+++ b/go.sum
@@ -198,8 +198,8 @@ github.com/hashicorp/hcl/v2 v2.3.0 h1:iRly8YaMwTBAKhn1Ybk7VSdzbnopghktCD031P8ggU
 github.com/hashicorp/hcl/v2 v2.3.0/go.mod h1:d+FwDBbOLvpAM3Z6J7gPj/VoAGkNe/gm352ZhjJ/Zv8=
 github.com/hashicorp/logutils v1.0.0 h1:dLEQVugN8vlakKOUE3ihGLTZJRB4j+M2cdTm/ORI65Y=
 github.com/hashicorp/logutils v1.0.0/go.mod h1:QIAnNjmIWmVIIkWDTG1z5v++HQmx9WQRO+LraFDTW64=
-github.com/hashicorp/terraform-exec v0.12.0 h1:Tb1VC2gqArl9EJziJjoazep2MyxMk00tnNKV/rgMba0=
-github.com/hashicorp/terraform-exec v0.12.0/go.mod h1:SGhto91bVRlgXQWcJ5znSz+29UZIa8kpBbkGwQ+g9E8=
+github.com/hashicorp/terraform-exec v0.13.0 h1:1Pth+pdWJAufJuWWjaVOVNEkoRTOjGn3hQpAqj4aPdg=
+github.com/hashicorp/terraform-exec v0.13.0/go.mod h1:SGhto91bVRlgXQWcJ5znSz+29UZIa8kpBbkGwQ+g9E8=
 github.com/hashicorp/terraform-json v0.8.0 h1:XObQ3PgqU52YLQKEaJ08QtUshAfN3yu4u8ebSW0vztc=
 github.com/hashicorp/terraform-json v0.8.0/go.mod h1:3defM4kkMfttwiE7VakJDwCd4R+umhSQnvJwORXbprE=
 github.com/hashicorp/terraform-plugin-go v0.2.1 h1:EW/R8bB2Zbkjmugzsy1d27yS8/0454b3MtYHkzOknqA=
--- a/internal/plugintest/helper.go
+++ b/internal/plugintest/helper.go
@@ -107,12 +107,6 @@ func (h *Helper) NewWorkingDir() (*WorkingDir, error) {
 		return nil, err
 	}
 
-	// symlink the provider source files into the base directory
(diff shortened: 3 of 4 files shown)


paultyng (approved):
This LGTM, not sure if you want @paddycarver to take a look too.

kmoe:
Actually @paddycarver hold on, there seems to be a failure in the DNS provider testing. That provider may have been a bad choice. Sorting it out now.

kmoe:
In fact I believe this might break certain test flows with multiple steps. Back to WIP.

kmoe:
I believe we can simplify this even further by removing config dirs altogether. This allows for Terraform state to persist automatically between test steps, as it does currently on `master`. Please see "Update following 515c542" in the PR description and compare the three directory structures.

Either way, this PR is a substantial enough change in the binary test framework that I am running tests on at least one major cloud provider to catch regressions.

kmoe:
AWS TC run ongoing.

bflad (approved):
515c54261a3427c3957572917876c7b9002cba3d looks good to me with AWS Provider acceptance testing with Terraform CLI 0.12.30, 0.13.5, 0.14.6, 0.15.0alpha20210127, and HEAD. 🚀

paddycarver (approved):
This makes sense to me and the test runs make me more confident.

PR #702, opened 2021-02-11
v1: fixing import testing

tombuildsstuff:
This PR fixes an issue when running import tests using Binary Testing in v1 - when running an import test at this point we get the error:

```
=== CONT  TestAccResourceGroup_basic
    testing_new_import_state.go:140: Failed to find schema for resource with ID ID = /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/acctestRG-210210191428921623
        % = 5
        location = westeurope
        name = acctestRG-210210191428921623
        tags.% = 0
        timeouts.% = 4
        Tainted = false
```

Tracking this down, this is due to trying to parse `provider.registry.terraform.io/hashicorp/blah.blah_resource` - which results in a failure, which isn't propagated here:

Screenshot 2021-02-11 at 10 52 02

As such this PR fixes this string to be `provider.blah.blah_resource` rather than `provider.registry.terraform.io/hashicorp/blah.blah_resource` - meaning this parses correctly - and the import tests succeed.


tombuildsstuff:
With this change, this parses correctly:

Screenshot 2021-02-11 at 10 57 20

And the test passes:

```
2021/02/11 10:54:57 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccResourceGroup_basic
=== PAUSE TestAccResourceGroup_basic
=== CONT  TestAccResourceGroup_basic
--- PASS: TestAccResourceGroup_basic (226.29s)
PASS
```

You could argue that we should raise this error, which I'd agree with - but there's a number of these so perhaps that wants tackling in a separate sweep?

helper/resource/testing_new_import_state.go
```diff
@@ -135,7 +135,14 @@ func testStepNewImportState(t *testing.T, c TestCase, wd *tftest.WorkingDir, ste
 			// this shouldn't happen in any reasonable case.
 			// KEM CHANGE FROM OLD FRAMEWORK: Fail test if this happens.
 			var rsrcSchema *schema.Resource
-			providerAddr, diags := addrs.ParseAbsProviderConfigStr("provider." + r.Provider + "." + r.Type)
+
+			// r.Provider at this point is `registry.terraform.io/hashicorp/blah` but we need `blah`
+			segments := strings.Split(r.Provider, "/")
```
paddycarver, line 140: I'd prefer if we could use the core code for parsing this string instead of relying on `strings.Split`; I think it makes it a little more robust and offers a little more flexibility about input, accepting the old version _and_ the new version at the same time?

paddycarver (changes requested):

helper/resource/testing_new_import_state.go
tombuildsstuff: sure, can you point me towards this?

paddycarver: https://pkg.go.dev/github.com/hashicorp/terraform/addrs#ParseProviderSourceString is what I had in mind

tombuildsstuff: done

kmoe: I'd rather not import the whole of Core into the SDK to solve this small problem, after we spent so much effort decoupling it (for reasons I won't repeat here). The `strings.Split` solution is fine for now, or you can use https://github.com/hashicorp/terraform-registry-address.

paddycarver: Can't use that package, it's not public yet.

Once https://github.com/hashicorp/terraform-plugin-test/pull/44 is merged I've got a local fix for this.

kmoe: What kind of local fix - do you mean a way to avoid importing Core?

paddycarver: Yeah, I copied over the functions instead of importing them.

paddycarver (approved):

Issue #710, opened 2021-02-20
Attribute with nested Set is marked as changed even if values was not touched


oliver-schoenherr:
### SDK version

latest Plugin SDK version in master

### Relevant SDK source code

`resource_data.go` takes not in consideration that the special case of comparing `Set` can happen not only on the first level but even deeper in the resource attribute tree.

```go
func (d *ResourceData) HasChange(key string) bool {
	o, n := d.GetChange(key)
        
        // If the type implements the Equal interface, then call that
	// instead of just doing a reflect.DeepEqual. An example where this is
	// needed is *Set
	if eq, ok := o.(Equal); ok {
		return !eq.Equal(n)
	}

	return !reflect.DeepEqual(o, n)
}
```

### Terraform Configuration Files

AWS ECS Service resource use such kind of schema:

```hcl
resource "aws_ecs_service" "default" {
  ....
  network_configuration {
    security_groups = [
    aws_security_group.default.id]
    subnets          = var.subnet_ids
    assign_public_ip = false
  }
}
```

### Expected Behavior

If `network_configuration` has no change, Terraform should not try to update the AWS ECS service.

### Actual Behavior

Terraform tries to update the AWS ECS service always, even without any change in the `network_configuration`.

### Steps to Reproduce

PR which enhance the `resource_data_test.go` test cases to reproduce the behaviour: https://github.com/oliver-schoenherr/terraform-plugin-sdk/pull/1/files

### References

https://github.com/hashicorp/terraform-provider-aws/issues/14711
https://github.com/hashicorp/terraform-provider-aws/issues/13658

PR #711, opened 2021-02-20
Fix issue #710

oliver-schoenherr:
Currently the `hasChanged` function does not take in consideration that the special case comparing a `Set` can happen on any level of the tree, not only on the top.

Instead of copy the `reflect.DeepEqual` function I just use the `go-cmp` library, which is already used in the code base, but for testing only.

### DIFF
--- a/helper/schema/resource_data.go
+++ b/helper/schema/resource_data.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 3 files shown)


bagedevimo:
Is there anything I can do to help get this merged/released? It'd be great as the solve for https://github.com/hashicorp/terraform-provider-aws/issues/13658.

richstokes (approved):

inhumantsar (approved):

PR #713, opened 2021-02-24
Test all support Go versions in CI.

paddycarver:
Update our test matrix to test both our supported Go versions in CI.
Also, fix the name of existing test steps to match the Go version
they're actually running with.


paddycarver:
#714 fixes the vet errors in Go 1.16 about our use of `t.Fatal` outside of test goroutines.

kmoe (approved):
Prefer rebase after #714 so tests are green before we merge this

Issue #721, opened 2021-03-09
Support configurations in JSON syntax in the provider test framework


rudo-thomas:
### SDK version
```
v2.4.4
```

### Use-cases

The end goal is being able to use the JSON syntax when authoring provider tests.

Currently, you're forced to write the test configurations in the native syntax (see [resource.TestStep.Config](
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.4.4/helper/resource#TestStep.Config
)).

We're generating our Terraform configuration in JSON syntax. Having the feature would simplify testing the generated JSON with Terraform CLI directly (which is what [resource.Test](
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.4.4/helper/resource#Test
) does), rather than passing the configuration through a conversion before feeding it to the test framework.

### Attempted Solutions

As stated above, converting from the native syntax to JSON would work, but that adds a conversion layer that is not present outside of the tests. Outside of the tests, the JSON configuration is fed to Terraform CLI directly via `*.tf.json` files.

### Proposal

(This is a revised proposal and now incorporates suggestions from [this comment](https://github.com/hashicorp/terraform-plugin-sdk/pull/722#issuecomment-941379029).)

Auto-detect whether [resource.TestStep.Config](
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.4.4/helper/resource#TestStep.Config
) contains valid JSON. When it does, the test framework will write out a `.tf.json` file, rather than a `.tf` one, before running the Terraform CLI.

I have a PR that achieves this: #722 

### References

- #235 would very likely solve the issue but seems much more involved

PR #722, opened 2021-03-09
Support testing of configurations in JSON syntax.

rudo-thomas:
This implements the proposal described in https://github.com/hashicorp/terraform-plugin-sdk/issues/721

The exported [plugintest.WorkingDir.SetConfig](
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.4.4/internal/plugintest#WorkingDir.SetConfig
) API was deliberately left unmodified to avoid a breaking change.

The only possible breaking change is the [mid-struct addition](
https://github.com/hashicorp/terraform-plugin-sdk/compare/master...rudo-thomas:support-json-syntax-in-tests?expand=1#diff-ef6290017990b7efb9ea81ebcad8974eb0a441f9f2adca610f403dbd32ea5619R405
) of the `ConfigIsJSON` field to `TestStep` which could break users that initialise the structure without using the the _Field_: syntax. Such uses are very unlikely, as the structure had 19 fields already, 20 with this change.

I wasn't able to find a good place to add tests for the new functionality. I'll be happy to take any advice on this. Thanks.

### DIFF
--- a/internal/plugintest/working_dir.go
+++ b/internal/plugintest/working_dir.go
@@ -3,6 +3,7 @@ package plugintest
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -15,8 +16,9 @@ import (
 )
 
 const (
-	ConfigFileName = "terraform_plugin_test.tf"
-	PlanFileName   = "tfplan"
+	ConfigFileName     = "terraform_plugin_test.tf"
+	ConfigFileNameJSON = ConfigFileName + ".json"
+	PlanFileName       = "tfplan"
 )
 
 // WorkingDir represents a distinct working directory that can be used for
@@ -28,6 +30,10 @@ type WorkingDir struct {
 
 	// baseDir is the root of the working directory tree
 	baseDir string
+
+	// configFilename is the full filename where the latest configuration
+	// was stored; empty until SetConfig is called.
+	configFilename string
 
 	// baseArgs is arguments that should be appended to all commands
 	baseArgs []string
@@ -84,11 +90,20 @@ func (wd *WorkingDir) GetHelper() *Helper {
 // Destroy to establish the configuration. Any previously-set configuration is
 // discarded and any saved plan is cleared.
 func (wd *WorkingDir) SetConfig(ctx context.Context, cfg string) error {
-	configFilename := filepath.Join(wd.baseDir, ConfigFileName)
-	err := ioutil.WriteFile(configFilename, []byte(cfg), 0700)
+	outFilename := filepath.Join(wd.baseDir, ConfigFileName)
+	rmFilename := filepath.Join(wd.baseDir, ConfigFileNameJSON)
+	bCfg := []byte(cfg)
+	if json.Valid(bCfg) {
+		outFilename, rmFilename = rmFilename, outFilename
+	}
+	if err := os.Remove(rmFilename); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("unable to remove %q: %w", rmFilename, err)
+	}
+	err := ioutil.WriteFile(outFilename, bCfg, 0700)
 	if err != nil {
 		return err
 	}
+	wd.configFilename = outFilename
 
 	var mismatch *tfexec.ErrVersionMismatch
 	err = wd.tf.SetDisablePluginTLS(true)
@@ -157,11 +172,16 @@ func (wd *WorkingDir) ClearPlan(ctx context.Context) error {
 	return nil
 }
 
+var errWorkingDirSetConfigNotCalled = fmt.Errorf("must call SetConfig before Init")
+
 // Init runs "terraform init" for the given working directory, forcing Terraform
 // to use the current version of the plugin under test.
 func (wd *WorkingDir) Init(ctx context.Context) error {
-	if _, err := os.Stat(wd.configFilename()); err != nil {
-		return fmt.Errorf("must call SetConfig before Init")
+	if wd.configFilename == "" {
+		return errWorkingDirSetConfigNotCalled
+	}
+	if _, err := os.Stat(wd.configFilename); err != nil {
+		return errWorkingDirSetConfigNotCalled
 	}
 
 	logging.HelperResourceTrace(ctx, "Calling Terraform CLI init command")
@@ -171,10 +191,6 @@ func (wd *WorkingDir) Init(ctx context.Context) error {
 	logging.HelperResourceTrace(ctx, "Called Terraform CLI init command")
 
 	return err
-}
-
-func (wd *WorkingDir) configFilename() string {
-	return filepath.Join(wd.baseDir, ConfigFileName)
 }
 
 func (wd *WorkingDir) planFilename() string {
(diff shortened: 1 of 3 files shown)


bflad:
Hi @rudo-thomas 👋  Thank you for raising this, it is a nice enhancement.

I'm curious if instead of requiring an additional and explicit configuration option, if this could instead be automatically determined by the content being passed through `Config` already -- e.g. just checking if the string parses as JSON. This would save threading the new option all the way through and could save some developer trouble. What do you think?

(Aside: Please note that the maintainers handle updating the `CHANGELOG.md` in this project on merge.)

rudo-thomas:
Thanks @bflad , your suggestion makes perfect sense: Automatic detection is much easier to use, there is no ambiguity (a HCL config cannot be valid JSON and vice-versa), and the implementation is cleaner too :)

I've re-worked and re-based this PR.

PTAL

internal/plugintest/working_dir.go
```diff
@@ -135,17 +153,13 @@ func (wd *WorkingDir) ClearPlan() error {
...
+	if wd.configFilename == "" {
 		return fmt.Errorf("must call SetConfig before Init")
 	}
 
 	return wd.tf.Init(context.Background(), tfexec.Reattach(wd.reattachInfo))
 }
 
-func (wd *WorkingDir) configFilename() string {
-	return filepath.Join(wd.baseDir, ConfigFileName)
-}
```
bflad, line 147: For consistency, I think we should still keep this method around to automatically do the file joining:

```go
func (wd *WorkingDir) configFilename() string {
	return filepath.Join(wd.baseDir, wd.configFilename)
}
```

And set without baseDir via:

```go
if json.Valid(bCfg) {
	wd.configFilename = ConfigFileNameJSON
} else {
	wd.configFilename = ConfigFileName
}
```

internal/plugintest/working_dir.go
```diff
@@ -135,17 +153,13 @@ func (wd *WorkingDir) ClearPlan() error {
 // Init runs "terraform init" for the given working directory, forcing Terraform
 // to use the current version of the plugin under test.
 func (wd *WorkingDir) Init() error {
-	if _, err := os.Stat(wd.configFilename()); err != nil {
+	if wd.configFilename == "" {
```
bflad, line 156: The original os.Stat() is likely safer here, in case configFilename isn't reset to "". 👍

internal/plugintest/working_dir.go
```diff
@@ -83,8 +89,20 @@ func (wd *WorkingDir) GetHelper() *Helper {
...
 // discarded and any saved plan is cleared.
 func (wd *WorkingDir) SetConfig(cfg string) error {
-	configFilename := filepath.Join(wd.baseDir, ConfigFileName)
-	err := ioutil.WriteFile(configFilename, []byte(cfg), 0700)
+	if wd.configFilename != "" {
+		err := os.Remove(wd.configFilename)
+		if err != nil && !os.IsNotExist(err) {
+			return err
+		}
+	}
```
bflad, line 97: Rather than checking for the filename, we should unilaterally remove/overwrite both the HCL and JSON configuration files here, just to prevent any potential oddities. If we want to be slightly more efficient, it can be done based on the `json.Valid()` conditional below 👍 

```go
if json.Valid(bCfg) {
	wd.configFilename = ConfigFileNameJSON

	if err := os.Remove(filepath.Join(wd.baseDir, ConfigFileName)); err != nil && !errors.Is(err, fs.ErrNotExist) {
		return fmt.Errorf("unable to remove %q: %w", ConfigFileName, err)
	}
} else {
	wd.configFilename = ConfigFileName

	if err := os.Remove(filepath.Join(wd.baseDir, ConfigFileNameJSON)); err != nil && !errors.Is(err, fs.ErrNotExist) {
		return fmt.Errorf("unable to remove %q: %w", ConfigFileNameJSON, err)
	}
}
```

bflad (changes requested):
Overall this is looking pretty good! Just some minor refactoring and I think this will be good to go.

internal/plugintest/working_dir.go
rudo-thomas: Unlike WorkingDir.planFilename() -- which is "static" in the sense that it only depends on the Workdir.baseDir, configFilename is now "dynamic". It can be empty (SetConfig() not called yet) or one of the two cases (HCL/JSON).

When I tried your suggestion, configFilename() would:
1. either have to return (string, error) (error in case no filename has been set up yet), or the caller would be responsible,
2. or the caller would be responsible to check the filename field for emptiness.

The first option is of course the clean and correct one, but it felt cumbersome at the only place where this would be called (Init(), see below).

If you feel strongly, I can refactor it to the method that returns (string, error).

internal/plugintest/working_dir.go
rudo-thomas: Based on the error message, I feel like the original os.Stat() is here as a test whether SetConfig has ever been called.

I re-added the call to os.Stat().

internal/plugintest/working_dir.go
rudo-thomas: Done.

I have also added a test that all of the deleting works.

Thanks for review, @bflad
PTAL

bflad (approved):
Looks good to me, thank you for the updates and tests, @rudo-thomas 🚀

internal/plugintest/working_dir_json_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// TestJSONConfig verifies that TestStep.Config can contain JSON.
+// This test also proves that when changing the HCL and JSON formats back and
+// forth, the framework deletes the previous configuration file.
+func TestJSONConfig(t *testing.T) {
+	providerFactories := map[string]func() (*schema.Provider, error){
+		"tst": func() (*schema.Provider, error) { return tstProvider(), nil },
```
bflad, line 17: Hopefully should appease the linter.

```suggestion
		"tst": func() (*schema.Provider, error) { return tstProvider(), nil }, //nolint:unparam // required signature
```

internal/plugintest/working_dir_json_test.go
```diff
@@ -0,0 +1,69 @@
...
+			Config: `{"resource":{"tst_t":{"r1":{"s":"x3"}}}}`,
+			Check:  resource.TestCheckResourceAttr("tst_t.r1", "s", "x3"),
+		}},
+	})
+}
+
+func tstProvider() *schema.Provider {
+	return &schema.Provider{
+		ResourcesMap: map[string]*schema.Resource{
+			"tst_t": &schema.Resource{
```
bflad, line 38: Hopefully should appease the linter.

```suggestion
			"tst_t": {
```

internal/plugintest/working_dir_json_test.go
```diff
@@ -0,0 +1,69 @@
...
+func tstProvider() *schema.Provider {
+	return &schema.Provider{
+		ResourcesMap: map[string]*schema.Resource{
+			"tst_t": {
+				CreateContext: resourceTstTCreate,
+				ReadContext:   resourceTstTRead,
+				UpdateContext: resourceTstTCreate, // Update is the same as Create
+				DeleteContext: resourceTstTDelete,
+				Schema: map[string]*schema.Schema{
+					"s": &schema.Schema{
```
bflad, line 44:
```suggestion
					"s": {
```

PR #723, opened 2021-03-09
helper/schema: Introduce context-aware resource CRUD functions without default timeout

### LINKED ISSUE #675: Schema Resource Timeout Affects Synchronized Resources Globally With Context Aware Functions
### SDK version

```
v2.4.0
```

### Use-cases

The Terraform Plugin SDK has always supported customizable timeouts, declared at the `helper/schema.Resource.Timeouts` level with the operation aware [`helper/schema.ResourceTimeout` type](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceTimeout). These were designed to allow practitioners to implement a timeout per resource, e.g.

```terraform
resource "example_thing" "example" {
  timeouts {
    create = "60m"
  }
}
```

Where the resource logic could fetch the timeout value via [`(helper/schema.ResourceData).Timeout()` function](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceData.Timeout).

Recently, context-aware resource operation functions were introduced and the existing non-context-aware functions were deprecated. These new context-aware resource operations forcibly introduce an operation timeout (defaults to 20 minutes) for the called logic, e.g.

https://github.com/hashicorp/terraform-plugin-sdk/blob/e21a5bbb73209dcb31f4c9dff3a8c0cec1ca65ac/helper/schema/resource.go#L283-L285

Previously, the non-context-aware resources could be implemented with `helper/schema.Resource.Timeouts.Create` defined and operation logic such as:

```go
func resourceExampleThingCreate(d *schema.ResourceData, meta interface{}) error {
  mutexkv.Lock("somekey")
  defer mutexkv.Unlock("somekey")
  // ... reference to d.Timeout(schema.TimeoutCreate) ...
}
```

Where in this case, the create timeout could be referenced in the code via `d.Timeout(schema.TimeoutCreate)` after the synchronization coordination. The creation timeout only affected this single operation, even if multiple operations were going to occur.

In the case of the context-aware functionality however, the timeout is configured before the mutex. So when multiple operations are waiting on the mutex, they are now sharing the same resource operation timeout. Since the timeout cannot be extended within the resource logic, developers are stuck either setting an arbitrarily high timeout (losing its usefulness) or practitioners are arbitrarily hitting the timeout as they scale out their configuration.

To characterize how some remote systems can require that operations must be serialized or otherwise synchronized:

* Only 1 resource type can be created at a time
* Only 1 child resource type can be created at a time per parent resource
* Only 10 concurrent updates of a resource type can be performed at a time
* Remote system may support configurable synchronization limits (either by the customer or remote system operators)

This essentially means the semantics of the synchronization can be based off resource configuration, operation type, and potentially accordant to the limits set by a remote system.

### Attempted Solutions

```go
&schema.Resource{
		Create: schema.DefaultTimeout(3 * time.Hour),
	},
}
```

#### Proposal 1: New Resource Methods for Locking

Introduce a resource operation hook before the context timeout is set for the purposes of handling the synchronization. To keep this generic, we can lean on the [`sync.Locker` interface](https://pkg.go.dev/sync#Locker) while allowing developers access to the resource data (e.g. attribute values) and meta (e.g. API client).

In the Terraform Plugin SDK, this would be something like:

```go
type ResourceLockFunc func (context.Context, *ResourceData, interface{}) (sync.Locker, diag.Diagnostics)

type Resource struct{
	CreateLock ResourceLockFunc
	DeleteLock ResourceLockFunc
	ReadLock   ResourceLockFunc
	UpdateLock ResourceLockFunc
}

func (r *Resource) create(ctx context.Context, d *ResourceData, meta interface{}) diag.Diagnostics {
	if r.CreateLock != nil {
		locker, diags := r.CreateLock(ctx, d, meta)
		
		if diags.HasError() {
			return diags
		}

		if locker != nil {
			locker.Lock()
			defer locker.Unlock()
		}
	}

	// ... existing logic ...
}
```

Where a simple provider implementation may look like:

```go
var resourceExampleThingMutex *sync.Mutex

&schema.Resource{
	CreateContext: resourceExampleThingCreate,
	CreateLock:    resourceExampleThingLock,
	Timeouts:      &schema.ResourceTimeout{
		Create: schema.DefaultTimeout(30 * time.Minute),
	},
}

func resourceExampleThingCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
	// ...
}

func resourceExampleThingLock(ctx context.Context, d *ResourceData, meta interface{}) (sync.Locker, diag.Diagnostics) {
	return resourceExampleThingMutex, nil
}
```

#### Proposal 2: New Context-Aware Methods Without Timeout

Another option is to provide a separate receiver method that is called before the context is set with the timeout so as to not "force" the resource timeout (instead it would be up to developers to manually implement it).

```go
type Resource struct{
	CreateWithoutTimeout CreateContextFunc
	DeleteWithoutTimeout DeleteContextFunc
	ReadWithoutTimeout   ReadContextFunc
	UpdateWithoutTimeout UpdateContextFunc
}

func (r *Resource) create(ctx context.Context, d *ResourceData, meta interface{}) diag.Diagnostics {
	if r.Create != nil {
		if err := r.Create(d, meta); err != nil {
			return diag.FromErr(err)
		}
		return nil
	}

	if r.CreateWithoutTimeout != nil {
		return r.CreateWithoutTimeout(ctx, d, meta)
	}

	ctx, cancel := context.WithTimeout(ctx, d.Timeout(TimeoutCreate))
	defer cancel()
	return r.CreateContext(ctx, d, meta)
}
```

Where a provider implementation may look like:

```go
&schema.Resource{
	CreateContext: resourceExampleThingCreate,
	Timeouts:      &schema.ResourceTimeout{
		Create: schema.DefaultTimeout(30 * time.Minute),
	},
}

func resourceExampleThingCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
	mutexkv.Lock("somekey")
	defer mutexkv.Unlock("somekey")
	// ... reference to d.Timeout(schema.TimeoutCreate) ...
}
```

bflad:
Closes #675

(Test changes may be easiest to view with the "hide whitespace changes" option in GitHub as they are just introducing looping to cover relevant Create/CreateContext/CreateWithoutTimeout method implementations.)

The Terraform Plugin SDK has always supported customizable timeouts, declared at the `helper/schema.Resource.Timeouts` level with the operation aware [`helper/schema.ResourceTimeout` type](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceTimeout). These were designed to allow practitioners to implement a timeout per resource, e.g.

```terraform
resource "example_thing" "example" {
  timeouts {
    create = "60m"
  }
}
```

Where the resource logic could fetch the timeout value via [`(helper/schema.ResourceData).Timeout()` function](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceData.Timeout).

Recently, context-aware resource operation functions were introduced and the existing non-context-aware functions were deprecated. These new context-aware resource operations forcibly introduce an operation timeout (defaults to 20 minutes) for the called logic, e.g.

https://github.com/hashicorp/terraform-plugin-sdk/blob/e21a5bbb73209dcb31f4c9dff3a8c0cec1ca65ac/helper/schema/resource.go#L283-L285

Previously, the non-context-aware resources could be implemented with `helper/schema.Resource.Timeouts.Create` defined and operation logic such as:

```go
func resourceExampleThingCreate(d *schema.ResourceData, meta interface{}) error {
  mutexkv.Lock("somekey")
  defer mutexkv.Unlock("somekey")
  // ... reference to d.Timeout(schema.TimeoutCreate) ...
}
```

Where in this case, the create timeout could be referenced in the code via `d.Timeout(schema.TimeoutCreate)` after the synchronization coordination. The creation timeout only affected this single operation, even if multiple operations were going to occur.

In the case of the context-aware functionality however, the timeout is configured before the mutex. So when multiple operations are waiting on the mutex, they are now sharing the same resource operation timeout. Since the timeout cannot be extended within the resource logic, developers are stuck either setting an arbitrarily high timeout (losing its usefulness) or practitioners are arbitrarily hitting the timeout as they scale out their configuration.

To characterize how some remote systems can require that operations must be serialized or otherwise synchronized:

* Only 1 resource type can be created at a time
* Only 1 child resource type can be created at a time per parent resource
* Only 10 concurrent updates of a resource type can be performed at a time
* Remote system may support configurable synchronization limits (either by the customer or remote system operators)

This essentially means the semantics of the synchronization can be based off resource configuration, operation type, and potentially accordant to the limits set by a remote system.

Here we introduce separate CRUD methods that are invoked before the context is set with the timeout, so as to not "force" the resource timeout. If timeout behavior is still desired, developers must manually implement timeout handling in the function logic when using these.

Most developers should prefer the existing context-aware functions and this is documented as such. However, another minor side benefit of these functions is that it also allows developers to upgrade CRUD function signatures for context awareness without logically changing resource behavior, then slowly switch to the timeout functions over time.

### DIFF
--- a/helper/schema/grpc_provider_test.go
+++ b/helper/schema/grpc_provider_test.go
@@ -697,138 +697,228 @@ func TestPlanResourceChange_bigint(t *testing.T) {
 }
 
 func TestApplyResourceChange(t *testing.T) {
-	r := &Resource{
-		SchemaVersion: 4,
-		Schema: map[string]*Schema{
-			"foo": {
-				Type:     TypeInt,
-				Optional: true,
+	testCases := []struct {
+		Description  string
+		TestResource *Resource
+	}{
+		{
+			Description: "Create",
+			TestResource: &Resource{
+				SchemaVersion: 4,
+				Schema: map[string]*Schema{
+					"foo": {
+						Type:     TypeInt,
+						Optional: true,
+					},
+				},
+				Create: func(rd *ResourceData, _ interface{}) error {
+					rd.SetId("bar")
+					return nil
+				},
 			},
 		},
-		CreateContext: func(_ context.Context, rd *ResourceData, _ interface{}) diag.Diagnostics {
-			rd.SetId("bar")
-			return nil
+		{
+			Description: "CreateContext",
+			TestResource: &Resource{
+				SchemaVersion: 4,
+				Schema: map[string]*Schema{
+					"foo": {
+						Type:     TypeInt,
+						Optional: true,
+					},
+				},
+				CreateContext: func(_ context.Context, rd *ResourceData, _ interface{}) diag.Diagnostics {
+					rd.SetId("bar")
+					return nil
+				},
+			},
+		},
+		{
+			Description: "CreateWithoutTimeout",
+			TestResource: &Resource{
+				SchemaVersion: 4,
+				Schema: map[string]*Schema{
+					"foo": {
+						Type:     TypeInt,
+						Optional: true,
+					},
+				},
+				CreateWithoutTimeout: func(_ context.Context, rd *ResourceData, _ interface{}) diag.Diagnostics {
+					rd.SetId("bar")
+					return nil
+				},
+			},
 		},
 	}
 
-	server := NewGRPCProviderServer(&Provider{
-		ResourcesMap: map[string]*Resource{
-			"test": r,
-		},
-	})
+	for _, testCase := range testCases {
+		testCase := testCase
+		t.Run(testCase.Description, func(t *testing.T) {
+			server := NewGRPCProviderServer(&Provider{
+				ResourcesMap: map[string]*Resource{
+					"test": testCase.TestResource,
+				},
+			})
 
-	schema := r.CoreConfigSchema()
-	priorState, err := msgpack.Marshal(cty.NullVal(schema.ImpliedType()), schema.ImpliedType())
-	if err != nil {
-		t.Fatal(err)
-	}
+			schema := testCase.TestResource.CoreConfigSchema()
+			priorState, err := msgpack.Marshal(cty.NullVal(schema.ImpliedType()), schema.ImpliedType())
+			if err != nil {
+				t.Fatal(err)
+			}
 
-	// A proposed state with only the ID unknown will produce a nil diff, and
-	// should return the proposed state value.
-	plannedVal, err := schema.CoerceValue(cty.ObjectVal(map[string]cty.Value{
-		"id": cty.UnknownVal(cty.String),
-	}))
-	if err != nil {
-		t.Fatal(err)
-	}
-	plannedState, err := msgpack.Marshal(plannedVal, schema.ImpliedType())
-	if err != nil {
-		t.Fatal(err)
-	}
+			// A proposed state with only the ID unknown will produce a nil diff, and
+			// should return the proposed state value.
+			plannedVal, err := schema.CoerceValue(cty.ObjectVal(map[string]cty.Value{
+				"id": cty.UnknownVal(cty.String),
+			}))
+			if err != nil {
+				t.Fatal(err)
+			}
+			plannedState, err := msgpack.Marshal(plannedVal, schema.ImpliedType())
+			if err != nil {
+				t.Fatal(err)
+			}
 
-	testReq := &tfprotov5.ApplyResourceChangeRequest{
-		TypeName: "test",
-		PriorState: &tfprotov5.DynamicValue{
-			MsgPack: priorState,
-		},
-		PlannedState: &tfprotov5.DynamicValue{
-			MsgPack: plannedState,
-		},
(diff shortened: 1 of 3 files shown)


paddycarver (approved):

paddycarver:
This looks reasonable to me. As we discussed, I find the proliferation of CRUD functions alarming--we're up to 12--but I think this is a pragmatic tradeoff given the compatibility constraints (the ability to restore prior behavior using sed) and the fact that we accidentally opted people into behavior without realizing it matters when discussing what solution is appropriate. I think in the future a boolean on the schema.Resource struct to toggle this behavior would be the right answer, but this is a bit of a special case.

PR #739, opened 2021-04-14
Upgrade terraform-plugin-go to 0.3.0 (dev)

paddycarver:
Upgrade terraform-plugin-go to the unreleased 0.3.0 candidate (still
awaiting merge, so commit hashes won't actually match) and account for
the breaking changes.

This is based off of hashicorp/terraform-plugin-go#74, and we should wait for that to land and be released before merging this.


paddycarver:
With the release of v0.3.0, this is ready.

kmoe (approved):
LGTM

PR #761, opened 2021-05-07
add ProtoV6ProviderFactories to TestCase

### LINKED ISSUE #737: Terraform protocol v6 support
With the merging of hashicorp/terraform-plugin-go#71, we now have a way to support version 6 of the Terraform protocol in the SDK. This issue is to track us doing that.

Open questions:

* Should we migrate everything onto proto6?
* Should we make Serve serve both 5 and 6?
* Should we surface proto6 features in the SDK?
* Should we add support for proto6 providers in helper/resource, which is how plugin-go providers are tested at the moment?

If we just upgrade from tfprotov5 to tfprotov6 and don't continue to support v5, we lose support for versions of Terraform below 0.15. This is probably a non-starter.

If we make plugin.Serve offer both 5 and 6, potentially by just upgrading requests over 5 to requests over 6 transparently, we can serve 6 for Terraform 0.15 and higher, and 5 for everything else.

If we surface proto6 features in the SDK, that mainly means supporting NestedType for attributes, letting them be used in lieu of blocks. Seeing as attributes and blocks are not cleanly surfaced, I'm not keen to open that can of worms right now.

I think we could add a `ProtoV6ProviderFactories` property to the `resource.TestCase` struct, like we added `ProtoV5ProviderFactories`, and that would give us test support for proto6 providers.

kmoe:
In order to support protocol v6 providers in the test suite, we add another
field to `TestCase` along the same lines as `ProtoV5ProviderFactories`. This is
plumbed through as yet another argument in `runProviderCommand`. We hope that by
the time another protocol major version is released, a new, more flexible test
framework will be available.

`plugin.ServeOpts` and `plugin.DebugServe` also expected all provider servers to be
v5 servers, so we also add the `GRPCProviderV6Func` field to `ServeOpts` to keep
it backwards compatible. If both `GRPCProviderFunc` and `GRPCProviderV6Func` are
set, the v6 server will be ignored.

Alternatively, we could panic if both are set.

fixes #737 

needs https://github.com/hashicorp/terraform-exec/pull/182

Tested with terraform-provider-corner at https://github.com/hashicorp/terraform-provider-corner/pull/18.

### DIFF
--- a/helper/resource/plugin.go
+++ b/helper/resource/plugin.go
@@ -19,7 +19,13 @@ import (
 	testing "github.com/mitchellh/go-testing-interface"
 )
 
-func runProviderCommand(t testing.T, f func() error, wd *plugintest.WorkingDir, factories map[string]func() (*schema.Provider, error), v5factories map[string]func() (tfprotov5.ProviderServer, error), v6factories map[string]func() (tfprotov6.ProviderServer, error)) error {
+type providerFactories struct {
+	legacy  map[string]func() (*schema.Provider, error)
+	protov5 map[string]func() (tfprotov5.ProviderServer, error)
+	protov6 map[string]func() (tfprotov6.ProviderServer, error)
+}
+
+func runProviderCommand(t testing.T, f func() error, wd *plugintest.WorkingDir, factories providerFactories) error {
 	// don't point to this as a test failure location
 	// point to whatever called it
 	t.Helper()
@@ -59,7 +65,7 @@ func runProviderCommand(t testing.T, f func() error, wd *plugintest.WorkingDir, 
 	// WaitGroup to listen for all of the close channels.
 	var wg sync.WaitGroup
 	reattachInfo := map[string]tfexec.ReattachConfig{}
-	for providerName, factory := range factories {
+	for providerName, factory := range factories.legacy {
 		// providerName may be returned as terraform-provider-foo, and
 		// we need just foo. So let's fix that.
 		providerName = strings.TrimPrefix(providerName, "terraform-provider-")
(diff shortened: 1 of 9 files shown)


helper/resource/plugin.go
```diff
@@ -12,13 +12,14 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/internal/plugintest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/plugin"
 	testing "github.com/mitchellh/go-testing-interface"
 )
 
-func runProviderCommand(t testing.T, f func() error, wd *plugintest.WorkingDir, factories map[string]func() (*schema.Provider, error), v5factories map[string]func() (tfprotov5.ProviderServer, error)) error {
+func runProviderCommand(t testing.T, f func() error, wd *plugintest.WorkingDir, factories map[string]func() (*schema.Provider, error), v5factories map[string]func() (tfprotov5.ProviderServer, error), v6factories map[string]func() (tfprotov6.ProviderServer, error)) error {
```
paddycarver, line 22: At some point (not right now) we probably want to refactor this to take a "factories" struct as an argument and just populate it with all these variations.

paddycarver (commented):
This all seems straightforward and like what I expected so far!

I think, ideally, we'd be able to test the version of Terraform we were using and throw a more helpful error if the factories in use weren't supported by that version of Terraform, and it would probably be helpful to have a built-in SkipFunc that skips tests based on Terraform versions, but neither of those are directly related to this PR, so we can tackle that work separately.

kmoe:
Added https://github.com/hashicorp/terraform-plugin-sdk/issues/776 for the above suggestion, which I agree with. For now I have added a note to the docs for `ProtoV6ProviderFactories`. Also refactored along the lines discussed.

The tests in https://github.com/hashicorp/terraform-provider-corner/pull/18 are passing and that PR will be updated once this one is merged and released.

paddycarver (approved):
Solid, yeah, let's do it!

Issue #628, opened 2020-10-29
Allow multiple versions of provider in acceptance tests


dak1n1:
### SDK version

```
{
  "Path": "github.com/hashicorp/terraform-plugin-sdk/v2",
  "Version": "v2.0.4-0.20200930154456-951f045a9f14"
}
```

### Use-cases

Simulate provider upgrade on existing infrastructure. Specifically, I would like the acceptance tests to tell me when one of my local changes might accidentally delete a user's existing infrastructure. (This can happen when `ForceNew` and `Default` are both used in the same schema attribute).

Here is a specific example:

A user has a storage volume created through Terraform. This is the existing infrastructure. They go to upgrade to the latest version of a Terraform Provider which manages this volume. One of the changes pulled in through that upgrade is a new attribute of that volume resource, and someone has accidentally added a `ForceNew` and `Default` option to that attribute.

When the user goes to `terraform apply`, Terraform adds the new attribute to the volume, but in doing so, it has deleted the user's volume and created a new one. This bug should be caught in acceptance tests.

#### Working solution ####
I have this work-around that's working for the Kubernetes provider, but it would be a lot neater if the SDK had a built-in way to do this instead. Basically, what I've done is compiled the provider from the local branch and named it "kubernetes-local". And I've grabbed the latest released version and named it "kubernetes-released". Then I load both of these into the acceptance test as ExternalProviders. 

In the acceptance test, I first create a resource using the released version, and then try a second apply using the pre-released version. Then, I compare the UUID of the resource to make sure it hasn't been deleted and recreated. Here is the code. It's spread across a few files:

https://github.com/hashicorp/terraform-provider-kubernetes/blob/0045c05d3cc24ed13e64f2974d63e6989ed0aa52/.terraformrc
https://github.com/hashicorp/terraform-provider-kubernetes/blob/0045c05d3cc24ed13e64f2974d63e6989ed0aa52/GNUmakefile#L46-L52

https://github.com/hashicorp/terraform-provider-kubernetes/blob/0045c05d3cc24ed13e64f2974d63e6989ed0aa52/kubernetes/provider_test.go#L38-L57

https://github.com/hashicorp/terraform-provider-kubernetes/blob/0045c05d3cc24ed13e64f2974d63e6989ed0aa52/kubernetes/provider_test.go#L453-L467

[...]

PR #778, opened 2021-07-06
add support for cross version provider import

ms-henglu:
This solution is used to detect SDK behavioral change when upgrade SDK for terraform providers. It takes advantage of existing acceptance tests. By deploying resources with released provider and using import step with developing provider to check whether there's a behavioral change.

Here's an example of detecting breaking change like https://azure.microsoft.com/en-us/updates/zone-behavior-change/. If there's a breaking behavioral change after SDK upgrade, running acc tests will show ` ImportStateVerify attributes not equivalent`.
```

=== RUN   TestAccPublicIpStatic_standard
=== PAUSE TestAccPublicIpStatic_standard
=== CONT  TestAccPublicIpStatic_standard
    testcase.go:95: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) (len=4) {
         (string) (len=7) "zones.#": (string) (len=1) "3",
         (string) (len=7) "zones.0": (string) (len=1) "1",
         (string) (len=7) "zones.1": (string) (len=1) "2",
         (string) (len=7) "zones.2": (string) (len=1) "3"
        }
        
        
        (map[string]string) {
        }
```

And this feature is disabled by default. So it won't affect regular acceptance tests.

Similar issue: https://github.com/hashicorp/terraform-plugin-sdk/issues/628 

This is just an idea to prevent breaking change in SDK upgrade, I'm open to all suggestions. :smiley:

### DIFF
--- a/helper/resource/cross_version_import_helper.go
+++ b/helper/resource/cross_version_import_helper.go
@@ -1,0 +1,70 @@
+package resource
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+/*
+	This helper is used to replace developing provider which contains a sdk upgrade and released provider.
+    By deploying resources with released provider and importing resources with developing provider, to verify whether there's backend breaking change introduced.
+*/
+
+func isCrossVersionImportEnabled() bool {
+	if enabled, err := strconv.ParseBool(os.Getenv("TF_ACC_CROSS_VERSION_IMPORT")); err == nil {
+		return enabled
+	}
+	return false
+}
+
+func getStandardProviderName() string {
+	if provider := os.Getenv("TF_ACC_PROVIDER"); provider != "" {
+		return provider
+	}
+	return "azurerm"
+}
+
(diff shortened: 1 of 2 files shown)


bflad:
Hi @ms-henglu 👋 Thank you for raising this awhile ago and apologies for the delayed review. We wound up resolving this in a different way over in #972 and I apologize that I had missed this existing attempt prior to submitting that.

Provider acceptance testing using v2.17.0 or later can now specify providers at either the `TestCase` or `TestStep` level.

For example:

```go
resource.TestCase{
	Steps: []resource.TestStep{
		// Use existing provider release
		{
			Config: `resource "null_resource" "test" {}`,
			ExternalProviders: map[string]resource.ExternalProvider{
				"null": {
					Source:            "registry.terraform.io/hashicorp/null",
					VersionConstraint: "3.1.1",
				},
			},
		},
		// Use provider codebase
		{
			Config: `resource "null_resource" "test" {}`,
				"null": func() (*schema.Provider, error) { /* ... */ },
			},
		},
	},
}
```

Hope this helps and thanks again for the submission!

PR #799, opened 2021-08-16
Update for Go 1.17 release.

paddycarver:
Updated our supported versions and test matrices in accordance with our
support policy.


bflad (approved):
Looks good to me 🚀  It'd probably be good to bump the `go.mod` and ensure a 2.7.1 release is cut before merging this. 👍

PR #802, opened 2021-08-27
Prototype exposing cty.Values from ResourceData.

### LINKED ISSUE #793: Expose cty.Value in ResourceData
- ResourceData.GetRawConfig
 - ResourceData.GetRawState
 - ResourceData.GetRawPlan

@paddycarver to populate.

paddycarver:
Fixes #793.

We actually create a `terraform.InstanceState` when handling the gRPC methods, and then a ResourceData is created from _that_. So, like with ProviderMeta functionality (#405) we're following the same path and threading our cty values through the instance state.

This still needs unit tests.


paddycarver:
Supported by hashicorp/terraform-provider-corner#29.

bflad (approved):
This looks good to me 🚀  Nice work threading everything through.

PR #805, opened 2021-09-21
Upgrade to latest terraform-plugin-go, add logging sink.

paddycarver:
Upgrade to the latest version of terraform-plugin-go, bringing support
for terraform-plugin-log.

Use the terraform-plugin-log logging sink when setting up the
helper/resource test framework. The idea here is that we'll use the
logging sink to reroute all terraform-plugin-log logs, and leave the
`log.Print` logs the way they are with the existing override. Over time,
it may be nice to migrate those logs to use terraform-plugin-log, which
we've introduced `logging.GetTestLogContext` as a helper for. But that
may be a mistake, and maybe we should wait until we need it to introduce
that function.

In theory, this PR should result in no logs being written to the CLI
during testing, which was happening in hashicorp/terraform-plugin-go#93.
It should respect `TF_LOG` and `TF_LOG_PATH`, as well as any
terraform-plugin-log style environment variables in use.

I think, by nature of this problem, we're going to need to test it manually for the moment, but I'll leave a comment with information on how to do that.

We should also wait to land this until hashicorp/terraform-plugin-log#15 lands and a release is cut, and hashicorp/terraform-plugin-go#93 lands and a release is cut, which will let us use tagged releases in our go.mod.

### DIFF
--- a/helper/logging/logging.go
+++ b/helper/logging/logging.go
@@ -1,6 +1,7 @@
 package logging
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


helper/logging/logging.go
```diff
@@ -10,27 +11,42 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-log/tfsdklog"
 	testing "github.com/mitchellh/go-testing-interface"
 )
 
 // These are the environmental variables that determine if we log, and if
 // we log whether or not the log should go to a file.
 const (
-	EnvLog     = "TF_LOG"      // Set to True
-	EnvLogFile = "TF_LOG_PATH" // Set to a file
+	EnvLog        = "TF_LOG"          // Set to True
```
bflad, line 21:
```suggestion
	EnvLog        = "TF_LOG"          // See ValidLevels
```

helper/logging/logging.go
```diff
@@ -114,3 +138,9 @@ func isValidLogLevel(level string) bool {
 
 	return false
 }
+
+func GetTestLogContext(t testing.T) context.Context {
```
bflad, line 142: I wonder if we should mark this one as not following our normal compatibility promises:

```suggestion
// GetTestLogContext creates a context that is registered to the SDK log sink.
// This function is for internal usage only and is not supported by the project's
// compatibility promises.
func GetTestLogContext(t testing.T) context.Context {
```

bflad (approved):
Looks good to me! 🚀  Just two little (optional) things to consider.

PR #811, opened 2021-10-07
Fix nil pointer dereference in ResourceData.HasChangeExcept

w-leads:
`ResourceData.HasChangeExcept` and `ResourceData. HasChangesExcept` accesses `.diff.Attributes` which could be nil and cause nil pointer dereference error.
This PR fixes the issue by checking the value and if it's nil, then return false(meaning no changes)

### DIFF
--- a/helper/schema/resource_data.go
+++ b/helper/schema/resource_data.go
@@ -172,7 +172,7 @@ func (d *ResourceData) HasChange(key string) bool {
 //
 // This function only works with root attribute keys.
 func (d *ResourceData) HasChangeExcept(key string) bool {
-	if d.diff == nil {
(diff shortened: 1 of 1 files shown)


bflad:
Great to know, @w-leads, and sorry you ran into trouble. Would you happen to have time to file a [bug report](https://github.com/hashicorp/terraform-plugin-sdk/issues/new?assignees=&labels=bug&template=bug_report.md) so we can capture some of the details of how this happened? We would love to create some test cases to prevent future regressions. If not, no worries. Appreciate the time and effort you have put into this already.

w-leads:
Thank you very much for the information! Will look into the documents
Also thank you for the review and merging the PR!

PR #826, opened 2021-11-30
fix: add path to error when list has too many or few items

josh-barker:
This PR adds the resource path to error messages when a list type either has too many or too few elements.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -1946,7 +1946,7 @@ func (m schemaMap) validateList(
 		return append(diags, diag.Diagnostic{
 			Severity:      diag.Error,
 			Summary:       "Too many list items",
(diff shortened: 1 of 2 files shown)


bflad (changes requested):
Hi @josh-barker 👋 Thank you for submitting this and apologies for the delayed review.

This partially relates to some of the discussion in https://github.com/hashicorp/terraform/issues/31575, so therefore I think this adjustment is acceptable because even if Terraform added the attribute path information to the diagnostic output, it would only show up in newer versions of Terraform while this would cover all cases of Terraform 0.12 and later when providers update their SDK version to one containing this change.

Would you be able to rebase this and ensure the unit testing is passing? Once that is done, we can get this in. Thanks again.

bflad (approved):
Looks good to me 🚀  Thanks so much, @josh-barker

josh-barker:
You're welcome @bflad ! 🙂

PR #829, opened 2021-12-07
Restore support for empty prefix in `ResourceDiff.GetChangedKeysPrefix`

### LINKED ISSUE #827: `ResourceDiff.GetChangedKeysPrefix` changed behavior with prefix of `""`
### SDK version

```
v2.9.0
```

```go
customdiff.ComputedIf("latest_version", func(_ context.Context, diff *schema.ResourceDiff, meta interface{}) bool {
	for _, changedKey := range diff.GetChangedKeysPrefix("") {
		switch changedKey {
		case "name", "name_prefix", "description", "default_version", "update_default_version":
			continue
		default:
			return true
		}
	}
	return false
}),
```

Relates #716.
Relates https://github.com/hashicorp/terraform-provider-aws/issues/22015.

With **v2.9.0**, a call to `diff.GetChangedKeysPrefix("")` returned changed keys.
The change introduced in #716 means that now no changed keys are returned - `childAddrOf(k, "")` always returns `false`.

ewbankkit:
Closes #827.

```console
% go test ./helper/schema
ok  	github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema	0.629s
```

### DIFF
--- a/helper/schema/resource_diff_test.go
+++ b/helper/schema/resource_diff_test.go
@@ -1312,7 +1312,7 @@ func TestGetChangedKeysPrefix(t *testing.T) {
 				"foo": "abc",
 				"bar": []interface{}{
 					map[string]interface{}{
-						"baz": "abc",
+						"baz": "def",
 						"qux": "uvw",
 					},
 				},
--- a/helper/schema/resource_diff_test.go
+++ b/helper/schema/resource_diff_test.go
@@ -1275,6 +1275,62 @@ func TestGetChangedKeysPrefix(t *testing.T) {
 			},
 		},
 		{
+			Name: "nested diff with empty prefix",
+			Schema: map[string]*Schema{
+				"foo": {
+					Type:     TypeString,
+					Optional: true,
+					Computed: true,
+				},
+				"bar": {
+					Type:     TypeList,
+					Required: true,
+					MaxItems: 1,
+					Elem: &Resource{
+						Schema: map[string]*Schema{
+							"baz": {
+								Type:     TypeString,
+								Optional: true,
+							},
+							"qux": {
+								Type:     TypeString,
+								Optional: true,
+							},
+						},
+					},
+				},
+			},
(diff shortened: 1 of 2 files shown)


dacut:
Looking at `childAddrOf`, should the comparison be ported into there? It looks like this function itself is erroneous if `prefix == ""`, and it's used elsewhere.

helper/schema/resource_diff_test.go
```diff
@@ -1237,6 +1237,36 @@ func TestGetChangedKeysPrefix(t *testing.T) {
...
+			Diff: &terraform.InstanceDiff{
+				Attributes: map[string]*terraform.ResourceAttrDiff{
+					"foo": {
+						Old: "bar",
+						New: "baz",
+					},
+				},
+			},
+			Key: "",
+			ExpectedKeys: []string{
```
bflad, line 1266: Can you please also add an unchanged root attribute and changed nested attribute to ensure that they are not returned and returned respectively in this test case? Thanks!

ewbankkit: Done.

helper/schema/resource_diff_test.go
```diff
@@ -1237,6 +1237,99 @@ func TestGetChangedKeysPrefix(t *testing.T) {
...
+					"bar.#":     "1",
+					"bar.0.baz": "def",
+					"bar.0.qux": "xyz",
+				},
+			},
+			Config: testConfig(t, map[string]interface{}{
+				"foo": "abc",
+				"bar": []interface{}{
+					map[string]interface{}{
+						"baz": "abc",
```
bflad, line 1315: The method relies on the diff, but this will make state == config, just in case that might confuse future travelers 👍 

```suggestion
						"baz": "def",
```

bflad (approved):
Looks good to me, just going to pull in the one minor testing change before merging. 🚀

PR #837, opened 2021-12-23
Add logging using terraform-plugin-log.

### LINKED ISSUE #804: Integration with terraform-plugin-log
In the wake of #695, we should integrate with [`terraform-plugin-log`](https://github.com/hashicorp/terraform-plugin-log) to support more advanced logging from Terraform providers.

The minimal version of this is to update our [`terraform-plugin-go`](https://github.com/hashicorp/terraform-plugin-go) to a release that includes hashicorp/terraform-plugin-go#93, which is a blocker for this, and to feed in the provider name appropriately.

The complete version of this is to insert SDK-level log statements of our own throughout the package to help with debugging, which will likely involve refactoring several internal-only functions to include context arguments.

I anticipate this being a delicate dance of backwards compatibility that will require quite a bit of pragmatism.

paddycarver:
Start using terraform-plugin-log to log SDKv2 level logs, using an
SDKv2-specific subsystem. The environment variable `TF_LOG_SDK_V2`
controls the subsystem's log output.

Fixes #804.

### DIFF
--- a/helper/schema/grpc_provider.go
+++ b/helper/schema/grpc_provider.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"log"
 	"strconv"
 	"sync"
 
 	"github.com/hashicorp/go-cty/cty"
 	ctyconvert "github.com/hashicorp/go-cty/cty/convert"
 	"github.com/hashicorp/go-cty/cty/msgpack"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-log/tfsdklog"
 
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
@@ -21,7 +22,15 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-const newExtraKey = "_new_extra_shim"
+const (
+	newExtraKey        = "_new_extra_shim"
+	tflogSubsystemName = "sdkv2"
+)
+
+func withLogger(ctx context.Context) context.Context {
+	return tfsdklog.NewSubsystem(ctx, tflogSubsystemName,
+		tfsdklog.WithLevelFromEnv("TF_LOG_SDK_V2"))
+}
 
 func NewGRPCProviderServer(p *Provider) *GRPCProviderServer {
 	return &GRPCProviderServer{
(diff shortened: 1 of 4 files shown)


helper/schema/grpc_provider.go
```diff
@@ -21,7 +22,15 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-const newExtraKey = "_new_extra_shim"
+const (
+	newExtraKey        = "_new_extra_shim"
+	tflogSubsystemName = "sdkv2"
```
bflad, line 27: I'm not the biggest fan of including the version number in the logging key as it could be missed during upgrades, but understandably the ability to differentiate between subsystems is important.

helper/schema/grpc_provider.go
```diff
@@ -21,7 +22,15 @@ import (
...
+)
+
+func withLogger(ctx context.Context) context.Context {
+	return tfsdklog.NewSubsystem(ctx, tflogSubsystemName,
+		tfsdklog.WithLevelFromEnv("TF_LOG_SDK_V2"))
```
bflad, line 32: Similarly here, although this has the worse consequence of requiring developer and practitioner changes when the next version is released so I'm a little more hesitant with this one. I'm not sure if I have a better naming choice though, since "SDK" is the name of this project. Maybe terraform-plugin-go should've used the sdk.go subsystem name and TF_LOG_SDK_GO while here this could use sdk.sdk subsystem name and TF_LOG_SDK_SDK (likely to be just generally documented as TF_LOG_SDK). For what its worth, terraform-plugin-framework doesn't use a subsystem, so that could also be an option for now.

bflad (approved):
A little hesitant on the subsystem name and environment variable including versioning information (maybe that versioning could be added as a separate key-value to the logging context), but otherwise looks good to me. 🚀

helper/schema/grpc_provider.go
paddycarver: What do we think of `helper_schema`?

helper/schema/grpc_provider.go
paddycarver: We definitely could just not use a subsystem, if that's the approach you prefer. Otherwise, I'll pitch `TF_LOG_SDK_HELPER_SCHEMA` as an environment variable? 🤔 

There's an argument to be made, I think, that if we use a subsystem here we should in framework, too. I don't know that I have a compelling argument either way, tbh, except that if we _don't_ use a subsystem, if we want to turn it off but keep the other SDK loggers on, we need to set each of their environment variables. If we put this in a subsystem, we can turn that subsystem off, and leave `TF_LOG_SDK` at whatever we want the other loggers to use.

bflad (approved):
Love using `helper_schema` (and `helper_resource` for the testing framework in the future) 👍 👍 👍

helper/schema/grpc_provider.go
bflad: We're going with `TF_LOG_SDK_HELPER_SCHEMA` -- thank you for following through with a good suggestion!

helper/schema/grpc_provider.go
bflad: Love it, great suggestion.

PR #840, opened 2021-12-30
Fix a tiny typo in the time helper docs

will-in-wi:
I noticed this while working on a project. Thanks for Terraform!


bflad:
Thanks for finding and fixing this, @will-in-wi! 🎉  It looks like `IsDayOfTheWeek` also has the same issue, would you be willing to adjust that Go documentation as well?

will-in-wi:
Sure, just pushed that up too. Thanks!

bflad (approved):
Awesome! Thanks so much, @will-in-wi 🚀

Issue #773, opened 2021-06-23
schema.EnvDefaultFunc doesn't work with int resource attributes


rudo-thomas:
`schema.EnvDefaultFunc` doesn't work when used as `DefaultFunc` on a `TypeInt` resource attribute; works fine with `TypeString`, `TypeBool`, `TypeFloat` attributes.

### SDK version

```
github.com/hashicorp/terraform-plugin-sdk/v2 v2.6.1
```

### Steps to Reproduce

Run `go test` in this repository: https://github.com/rudo-thomas/terraform-sdk-default-int-from-env-bug (a snapshot at b3f1b82 is 
[attached](https://github.com/hashicorp/terraform-plugin-sdk/files/6701177/terraform-sdk-default-int-from-env-bug.tar.gz)).

### Relevant provider source code

```go
"i": &schema.Schema{
	Type:        schema.TypeInt,
	Optional:    true,
	DefaultFunc: schema.EnvDefaultFunc("SOME_VAR", nil),
}
```

### Expected Behavior
The test should pass.

### Actual Behavior
```
Error: Attribute must be a whole number, got 1
```

Note: The `1` in the error message is of `string` type, which I found out by printing the type at https://github.com/hashicorp/terraform-plugin-sdk/blob/5747fc30ee97967b4bec39de8dbf3d879616b4cb/helper/schema/schema.go#L2053

### Workaround

To work this around, the `DefaultFunc` set in the `Schema` can wrap `schema.EnvDefaultFunc(...)()` and convert the `string` to an `int`.

rudo-thomas:
PR #841 fixes this on `main` (past `v2.10.1`).

PR #841, opened 2022-01-07
helper/schema: Make int fields accept string values from DefaultFunc().

rudo-thomas:
This brings the int field up to par with bool and float ones.

Fixes https://github.com/hashicorp/terraform-plugin-sdk/issues/773


bflad (approved):
Thanks for this contribution, @rudo-thomas. I'm pretty sure this should be okay to allow. 🚀  

I'll add a changelog entry post merge.

PR #862, opened 2022-01-26
deps: Remove github.com/golang/mock and github.com/go-test/deep

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-sdk/pull/851
Reference: https://github.com/hashicorp/terraform-plugin-sdk/pull/849

The `mockgen` command is not used in this repository and likely was a vestiage of splitting this repository from github.com/hashicorp/terraform.

This repository (and many of the other Terraform Plugin repositories) overwhelmingly use `go-cmp` instead of `deep`. The three usages are migrated for consistency and to remove this dependency.


ewbankkit (approved):
LGTM :rocket:.

PR #867, opened 2022-01-31
add docs on why old might be empty in diffsuppressfunc

megan07:
add docs on why old might be empty in diffsuppressfunc


bflad (approved):
Thanks, @megan07 🚀

Issue #865, opened 2022-01-31
Resolve golangci-lint Issues


bflad:
### Description

Enabling various `golangci-lint` linters reveals a broad set of stylistic and potentially buggy code implementations:

```console
❯ golangci-lint run ./...
terraform/state_filter.go:39:8: append to slice `as` with non-zero initialized length (makezero)
                as = append(as, &resourceAddress{Index: -1})
                     ^
terraform/state.go:387:30: `(*State).removeModule` - `path` is unused (unparam)
func (s *State) removeModule(path []string, v *ModuleState) {
                             ^
terraform/state.go:415:32: `(*State).removeInstance` - `path` is unused (unparam)
func (s *State) removeInstance(path []string, r *ResourceState, v *InstanceState) {
                               ^
terraform/state.go:1240:2: S1008: should use 'return s.Primary.Equal(other.Primary)' instead of 'if !s.Primary.Equal(other.Primary) { return false }; return true' (gosimple)
        if !s.Primary.Equal(other.Primary) {
        ^
terraform/state.go:1026:20: S1039: unnecessary use of fmt.Sprintf (gosimple)
                        buf.WriteString(fmt.Sprintf("\n  Dependencies:\n"))
                                        ^
terraform/resource.go:157:2: variable copy has same name as predeclared identifier (predeclared)
        copy, err := copystructure.Config{Lock: true}.Copy(c)
        ^
terraform/state.go:565:2: variable copy has same name as predeclared identifier (predeclared)
        copy, err := copystructure.Config{Lock: true}.Copy(s)
        ^
terraform/state.go:1433:2: variable copy has same name as predeclared identifier (predeclared)
        copy, err := copystructure.Config{Lock: true}.Copy(s)
        ^
terraform/resource_test.go:200:4: variable copy has same name as predeclared identifier (predeclared)
                        copy := rc.DeepCopy()
                        ^
terraform/state.go:431:5: SA4006: this value of `is` is never used (staticcheck)
                                is, is[len(is)-1] = append(is[:i], is[i+1:]...), nil
                                ^
internal/addrs/module_instance.go:95:4: S1023: redundant break statement (gosimple)
                        break
                        ^
internal/addrs/module_instance.go:125:4: S1023: redundant break statement (gosimple)
                        break
                        ^
helper/logging/transport.go:55:15: Error return value of `json.Indent` is not checked (errcheck)
                        json.Indent(&out, b, "", " ")
                                   ^
[...]

PR #870, opened 2022-02-01
Enable ineffassign linter and fix issues

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-go/tree/main/internal/logging

The majority of these issues were in the gRPC server handling for some recent inclusion of logging that did not include other context handling. This adds some logic handoff trace logs to the helper/schema logger and starts an internal logging package, similar to the terraform-plugin-go implementation.

Previously:

```text
internal/plugintest/util.go:13:4: ineffectual assignment to err (ineffassign)
                        err = os.Chmod(dest, srcInfo.Mode())
                        ^
helper/schema/grpc_provider.go:75:2: ineffectual assignment to ctx (ineffassign)
        ctx = withLogger(ctx)
        ^
helper/schema/grpc_provider.go:126:2: ineffectual assignment to ctx (ineffassign)
        ctx = withLogger(ctx)
        ^
helper/schema/grpc_provider.go:230:2: ineffectual assignment to ctx (ineffassign)
        ctx = withLogger(ctx)
        ^
helper/schema/grpc_provider.go:249:2: ineffectual assignment to ctx (ineffassign)
        ctx = withLogger(ctx)
        ^
helper/schema/grpc_provider.go:501:2: ineffectual assignment to ctx (ineffassign)
        ctx = withLogger(ctx)
        ^
helper/schema/grpc_provider.go:754:16: ineffectual assignment to err (ineffassign)
        plannedAttrs, err := diff.Apply(priorState.Attributes, schemaBlock)
                      ^
helper/schema/resource_data.go:531:2: ineffectual assignment to level (ineffassign)
        level := "set"
        ^
```

### DIFF
--- a/internal/logging/context.go
+++ b/internal/logging/context.go
@@ -1,0 +1,14 @@
+package logging
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-log/tfsdklog"
+)
+
+// InitContext creates SDK logger contexts.
+func InitContext(ctx context.Context) context.Context {
+	ctx = tfsdklog.NewSubsystem(ctx, SubsystemHelperSchema, tfsdklog.WithLevelFromEnv(EnvTfLogSdkHelperSchema))
+
+	return ctx
+}
--- a/internal/plugintest/util.go
+++ b/internal/plugintest/util.go
@@ -1,20 +1,31 @@
 package plugintest
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 )
 
-func symlinkFile(src string, dest string) (err error) {
-	err = os.Symlink(src, dest)
-	if err == nil {
-		srcInfo, err := os.Stat(src)
-		if err != nil {
-			err = os.Chmod(dest, srcInfo.Mode())
-		}
+func symlinkFile(src string, dest string) error {
+	err := os.Symlink(src, dest)
+
+	if err != nil {
+		return fmt.Errorf("unable to symlink %q to %q: %w", src, dest, err)
 	}
 
-	return
+	srcInfo, err := os.Stat(src)
+
+	if err != nil {
+		return fmt.Errorf("unable to stat %q: %w", src, err)
+	}
+
+	err = os.Chmod(dest, srcInfo.Mode())
+
+	if err != nil {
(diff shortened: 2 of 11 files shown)


internal/plugintest/util.go
```diff
@@ -1,20 +1,31 @@
...
+		return fmt.Errorf("unable to stat %q: %w", src, err)
+	}
+
+	err = os.Chmod(dest, srcInfo.Mode())
+
+	if err != nil {
+		return fmt.Errorf("unable to set %q permissions: %w", dest, err)
+	}
+
+	return nil
```
detro, line 28: I was trying to read the previous version and getting very confused, until I realised it was starting with a `if err == nil`, the opposite of what I'd have expected.

Now it's WAY more readable and I see even more why you liked that article about "keep the happy path on level-1" :)

detro (approved):
I see that it looks consistent with the other `terraform-plugin-go` library, but do you have something I can read to better grasp the work around logging that was done?

There is a lot of context I still need to absorb :(

internal/logging/context.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/hashicorp/terraform-plugin-log/tfsdklog"
+)
+
+// InitContext creates SDK logger contexts.
+func InitContext(ctx context.Context) context.Context {
+	ctx = tfsdklog.NewSubsystem(ctx, SubsystemHelperSchema, tfsdklog.WithLevelFromEnv(EnvTfLogSdkHelperSchema))
+
+	return ctx
+}
```
detro, line 14:
```suggestion
func WithLogging(ctx context.Context) context.Context {
	return tfsdklog.NewSubsystem(ctx, SubsystemHelperSchema, tfsdklog.WithLevelFromEnv(EnvTfLogSdkHelperSchema))
}
```

detro: or `WithLogger`. Just to kinda mimic stuff like https://pkg.go.dev/context#WithValue

detro (approved):
Thank you for taking the time (in slack and here) to clarify things for me. I have left a suggestion on naming (just because I have always see those kind of "wrapping" using the `WithX` nomenclature).

Other than that, and it's a totally optional suggestion, LGTM

PR #871, opened 2022-02-01
Enable nilerr linter and fix issues

bflad:

The field reader/writing handling of this codebase is generally considered frozen, but we want to catch and fix other potential `nilerr` issues.

Previously:

```text
helper/schema/field_reader_config.go:206:3: error is not nil (line 204) but it returns nil (nilerr)
                return FieldReadResult{}, nil
                ^
helper/schema/field_reader_diff.go:131:3: error is not nil (line 129) but it returns nil (nilerr)
                return FieldReadResult{}, nil
                ^
helper/schema/field_reader_map.go:66:3: error is not nil (line 64) but it returns nil (nilerr)
                return FieldReadResult{}, nil
                ^
helper/schema/grpc_provider.go:199:3: error is not nil (line 139) but it returns nil (nilerr)
                return resp, nil
                ^
```


detro (approved):
LGTM

PR #878, opened 2022-02-03
helper/schema: Update misleading error messaging during import

### LINKED ISSUE #877: Misleading Legacy Error Messaging During Import
### SDK version

```
v2.10.1
```

Any resource code that incorrectly implements `helper/schema.Resource.Importer` and returns a `ResourceData` missing an ID.

N/A

### Expected Behavior

The error messaging should refer practitioners running into this issue to report the bug to the provider, not Terraform.

### Actual Behavior

The SDK returns a legacy error from when Terraform, the SDK, and providers used to live in the same repository.

```console
example_thing.example: Importing from ID "0123-456789-7ya8xyyz"...
╷
│ Error: nil entry in ImportState results. This is always a bug with
│ the resource that is being imported. Please report this as
│ a bug to Terraform.
```

### Steps to Reproduce

1. tf import example_thing.example 0123-456789-7ya8xyyz

### References

- https://github.com/hashicorp/terraform/issues/30463

bflad:
Closes #877

Ideally, this method would've been updated to use and return diagnostics in addition to the new error handling, however its signature is protected by major version compatibility guarantees.

### DIFF
--- a/helper/schema/provider.go
+++ b/helper/schema/provider.go
@@ -391,6 +391,21 @@ func (p *Provider) ImportState(
 	// Convert the results to InstanceState values and return it
 	states := make([]*terraform.InstanceState, len(results))
 	for i, r := range results {
+		if r == nil {
+			return nil, fmt.Errorf("The provider returned a missing resource during ImportResourceState. " +
+				"This is generally a bug in the resource implementation for import. " +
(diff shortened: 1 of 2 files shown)


detro (approved):
LGTM

PR #881, opened 2022-02-04
Enable makezero linter and fix issue

bflad:

Previously:

```text
terraform/state_filter.go:39:8: append to slice `as` with non-zero initialized length (makezero)
                as = append(as, &resourceAddress{Index: -1})
                     ^
```


detro (approved):
LGTM

PR #887, opened 2022-02-09
Comments to clarify TestCheckResourceAttrSet()

### LINKED ISSUE #885: testCheckResourceAttrSet() is broken for all but TypeString attributes
I recently spent about 12 hours trying to find a bug in my code which would result in the testing function `testCheckResourceAttrSet()` reporting a certain resource attribute was not set.

In fact, the attribute WAS set, and the function contains a bug; when you try to check an attribute which is not a simple `map[string]string` and instead, eg. `[]map[string]interface{}`, the function will fail.

The proposal here is to add a comment to the function so anyone using autocomplete in the future is aware this function should be used for string attributes only. 

To verify this is the case, I wrote a simple `panic()` test inside my resource code:
```go
if dashs, ok := d.GetOk("dashboards"); ok {
  panic(fmt.Errorf("dashboards attr = %#v", dashs))
}
```

this **guaranteed** the test would exist in a stack trace if the resource attribute `dashboards` was set; it did. Removing this small bit of code resulted in the test framework error:

```console
?       github.com/grafana/terraform-provider-grafana   [no test files]
=== RUN   TestAccDataSourceDashboardsAllAndByFolderID
    data_source_dashboards_test.go:38: Step 1/1 error: Check failed: Check 5/9 error: data.grafana_dashboards.all: Attribute 'dashboards' expected to be set
--- FAIL: TestAccDataSourceDashboardsAllAndByFolderID (3.46s)
```

justinTM:
closes #885

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -677,10 +677,10 @@ func ComposeAggregateTestCheckFunc(fs ...TestCheckFunc) TestCheckFunc {
 	}
 }
 
-// TestCheckResourceAttrSet is a TestCheckFunc which ensures a value
+// TestCheckResourceAttrSet is a TestCheckFunc which ensures a string value
 // exists in state for the given name/key combination. It is useful when
 // testing that computed values were set, when it is not possible to
-// know ahead of time what the values will be.
+// know ahead of time what the values will be. List or map values will fail.
 func TestCheckResourceAttrSet(name, key string) TestCheckFunc {
 	return checkIfIndexesIntoTypeSet(key, func(s *terraform.State) error {
 		is, err := primaryInstanceState(s, name)


bflad:
Thinking about this more, one complication here is that indeed checking the list/set/map attribute itself, not one of the special sub-keys, will likely not be represented in the flatmap state. I'm wondering if it might be best to mention both things in that case by having a statement such as "Collection attributes can only be inspected by the special key syntax, not the attribute key itself."

bflad:
Hi again, @justinTM 👋  Since I haven't heard back from you for awhile, I'm going to close this. I'll submit the above suggestions separately so the Go documentation is updated appropriately. Thanks again for raising this.

PR #889, opened 2022-02-17
helper/resource: Add (testing.T).Logf() call and augment documentation for TestStep.SkipFunc

### LINKED ISSUE #824: TestStep SkipFunc should not error but print to stdout
It seems that [SkipFunc](https://github.com/hashicorp/terraform-plugin-sdk/blob/008381a3961560cdea904b9663725fa4243fabb4/helper/resource/testing.go#L457) would be more helpful in test steps if it printed a message to stdout instead of erroring the test. that is how `t.Skip()` works

```
type TestStep struct {
    ...

	// SkipFunc is called before applying config, but after PreConfig
	// This is useful for defining test steps with platform-dependent checks
	SkipFunc func() (bool, error)
```

work arounds are easy enough for now:
```
SkipFunc: func() (bool, error) {
	if myVar != "" {
		return false, nil
	}
	fmt.Println("Environment variable MYVAR must be set.")
	return true, nil
},
```

bflad:
Closes #824
Reference: https://pkg.go.dev/testing#T.Logf

The additional logging will be more visible, as it will appear in stdout instead of disabled by default.

The documentation updates should make the purpose and correct usage of this field more clear.


detro (approved):

drewmullen:
Awesome! Thank you!

PR #892, opened 2022-02-22
helper/resource: Only call final Terraform CLI show command each TestStep when TestCase.IDRefreshName is set

### LINKED ISSUE #730: TestCase ID-Only Refresh Testing Documentation / Generally Extraneous State Fetch
### SDK version

```
v2.4.4
```

### General Description

This issue is going to summarize a few quick findings with the acceptance testing framework's identifier only refresh testing configuration. Happy to split this up into multiple issues.

Given the current [`helper/resource#TestCase.IDRefreshName` Go documentation](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource#TestCase.IDRefreshName):

```
 	// The settings below control the "ID-only refresh test." This is
	// an enabled-by-default test that tests that a refresh can be
	// refreshed with only an ID to result in the same attributes.
	// This validates completeness of Refresh.
```

ID-only refresh testing is always performed for all testing. This however is not the case according to the implementation:

https://github.com/hashicorp/terraform-plugin-sdk/blob/1e9f42ca7c3b7f326d09c04bdfbdea56bd96c2de/helper/resource/testing_new_config.go#L18

https://github.com/hashicorp/terraform-plugin-sdk/blob/1e9f42ca7c3b7f326d09c04bdfbdea56bd96c2de/helper/resource/testing_new_config.go#L217

The preference here should likely be to keep the current behavior and not automatically run this type of testing -- instead updating the Go documentation to reflect the current state of affairs. Switching this may cause a large number of tests to fail with no method to disable it.

I found this through a failing test that was combining `IDRefreshName` (for legacy reasons) and `ProviderFactories` with an alternate provider configuration in the `TestStep.Config`, which causes Terraform CLI to emit this error:

```
=== CONT  TestAccAWSVPCPeeringConnection_region
    testing_new.go:214: Error running terraform refresh: exit status 1

        Error: Provider configuration not present

        To work with aws_vpc.peer (orphan) its original provider configuration at
        provider["registry.terraform.io/hashicorp/awsalternate"] is required, but it
        has been removed. This occurs when a provider configuration is removed while
        objects created by that provider still exist in the state. Re-add the provider
        configuration to destroy aws_vpc.peer (orphan), after which you can remove the
        provider configuration again.

--- FAIL: TestAccAWSVPCPeeringConnection_region (24.38s)
```

The ID-only refresh testing would need the provider configuration to actually be successful in this case, which as far as I understand is not the purpose of this functionality. I'm guessing this limitation should also be added to the Go documentation for the `IDRefresh*` fields.

Also peeking at the testing code right before the conditional into the ID-only refresh testing, the Terraform CLI is invoked to fetch the current state:

https://github.com/hashicorp/terraform-plugin-sdk/blob/1e9f42ca7c3b7f326d09c04bdfbdea56bd96c2de/helper/resource/testing_new_config.go#L206-L217

This seems like it should only occur if ID-only refresh testing will be called. It might save a few CPU/disk cycles.

All this said, providers should _generally_ prefer `ImportState` testing over ID-only refresh testing in the current testing framework as it is more useful and accomplishes the same overall goal. Essentially if the resource supports ID-only refresh well, there isn't much more effort to implementing resource import (e.g. import state passthrough). If there are actual valid use cases it might be worth considering documenting them, otherwise it might be good to consider this for deprecation with upcoming testing framework proposals.

In the AWS Provider, going to submit issues to discourage/remove this testing functionality since it can cause confusing errors.

```go
const (
	// Provider name for single configuration testing
	ProviderNameAws = "aws"

	// Provider name for alternate configuration testing
	ProviderNameAwsAlternate = "awsalternate"
)

func TestAccAWSVPCPeeringConnection_region(t *testing.T) {
	var connection ec2.VpcPeeringConnection
	var providers []*schema.Provider
	rName := fmt.Sprintf("tf-testacc-pcx-%s", acctest.RandString(17))
	resourceName := "aws_vpc_peering_connection.test"

	resource.ParallelTest(t, resource.TestCase{
		PreCheck: func() {
			testAccPreCheck(t)
			testAccMultipleRegionPreCheck(t, 2)
		},
		ErrorCheck:      testAccErrorCheck(t, ec2.EndpointsID),
		IDRefreshName:   resourceName,
		IDRefreshIgnore: []string{"auto_accept"},

		ProviderFactories: testAccProviderFactoriesAlternate(&providers),
		CheckDestroy:      testAccCheckAWSVpcPeeringConnectionDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccVpcPeeringConfig_region_autoAccept(rName, false),
				Check: resource.ComposeTestCheckFunc(
					testAccCheckAWSVpcPeeringConnectionExists(
						resourceName,
						&connection,
					),
					resource.TestCheckResourceAttr(
						resourceName,
						"accept_status",
						"pending-acceptance",
					),
				),
			},
		},
	})
}

// Deprecated: Use testAccMultipleRegionProviderConfig instead
func testAccAlternateRegionProviderConfig() string {
	return testAccNamedRegionalProviderConfig(ProviderNameAwsAlternate, testAccGetAlternateRegion())
}

// testAccNamedRegionalProviderConfig creates a new provider named configuration with a region.
//
// This can be used to build multiple provider configuration testing.
func testAccNamedRegionalProviderConfig(providerName string, region string) string {
	//lintignore:AT004
	return fmt.Sprintf(`
provider %[1]q {
  region = %[2]q
}
`, providerName, region)
}

func testAccVpcPeeringConfig_region_autoAccept(rName string, autoAccept bool) string {
	return testAccAlternateRegionProviderConfig() + fmt.Sprintf(`
resource "aws_vpc" "test" {
  cidr_block = "10.0.0.0/16"

  tags = {
    Name = %[1]q
  }
}

resource "aws_vpc" "peer" {
  provider = "awsalternate"

  cidr_block = "10.1.0.0/16"

  tags = {
    Name = %[1]q
  }
}

resource "aws_vpc_peering_connection" "test" {
  vpc_id      = aws_vpc.test.id
  peer_vpc_id = aws_vpc.peer.id
  peer_region = %[3]q
  auto_accept = %[2]t

  tags = {
    Name = %[1]q
  }
}
`, rName, autoAccept, testAccGetAlternateRegion())
}
```

### Debug Output

Please let me know if you want these. 😄 

### Steps to Reproduce

1. `TF_ACC=1 go test ./aws -v -count 1 -timeout 120m -parallel 20 -run='TestAccAWSVPCPeeringConnection_region`

### References

- https://github.com/search?q=org%3Ahashicorp+IDRefreshName&type=code
- https://github.com/hashicorp/terraform-provider-aws/issues/18354
- https://github.com/hashicorp/terraform-plugin-sdk/pull/291
- https://github.com/bflad/tfproviderlint/issues/228

bflad:
Closes #730

Also clarifies the Go documentation for TestCase type IDRefreshName and IDRefreshIgnore fields.


detro (approved):
I admit that I don't understand what an "id-only refresh" is yet.

But LGTM.

PR #893, opened 2022-02-23
Extra `TestCheckFunc`s to check attribute's string length

detro:
## Background

Sometimes, when writing tests for a provider, one needs to check that a value is a certain length or length range, while cannot rely on expecting a specific value to match. This is for example true for situations when the provider will generate a value entirely randomised, but that "should be at least this long/short".

## Solution proposed

2 additional functions in `helper/resource/testing.go`:

* `func TestRangeLengthResourceAttr(name, key string, min, max int) TestCheckFunc`
* `func TestMatchLengthResourceAttr(name, key string, length int) TestCheckFunc`

This PR comes with tests to exercise this utility.

## Optional: `.gitignore` was missing

I have taken the liberty to provide this repository with a `.gitignore`. To avoid basing it on personal taste alone, I have set it up based on https://github.com/github/gitignore.

If this is a problem, I'd be happy to move this to a separate, dedicated PR.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -1031,6 +1031,36 @@ func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst strin
 	return nil
 }
 
+// TestRangeLengthResourceAttr is a TestCheckFunc which checks that the length of the value
+// in state for the given name/key is within an expected (closed) range
+func TestRangeLengthResourceAttr(name, key string, min, max int) TestCheckFunc {
+	return checkIfIndexesIntoTypeSet(key, func(s *terraform.State) error {
+		is, err := primaryInstanceState(s, name)
+		if err != nil {
+			return err
+		}
+
+		valLen := len(is.Attributes[key])
+		if valLen < min || valLen > max {
+			return fmt.Errorf(
+				"%s: Attribute '%s' length is %d, not within the expected range [%d, %d]",
+				name,
+				key,
+				valLen,
+				min,
+				max)
+		}
+
+		return nil
+	})
+}
+
+// TestMatchLengthResourceAttr is a TestCheckFunc which checks that the length of the value
+// in state for the given name/key matches a specific length
+func TestMatchLengthResourceAttr(name, key string, length int) TestCheckFunc {
+	return TestRangeLengthResourceAttr(name, key, length, length)
+}
+
 // TestCheckOutput checks an output in the Terraform configuration
 func TestCheckOutput(name, value string) TestCheckFunc {
 	return func(s *terraform.State) error {
--- a/helper/resource/testing_test.go
+++ b/helper/resource/testing_test.go
@@ -874,6 +874,87 @@ func TestCheckResourceAttr_empty(t *testing.T) {
 	}
 }
 
+func TestTestRangeLengthResourceAttr(t *testing.T) {
+	resName := "test_resource"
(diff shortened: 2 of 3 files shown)


bookshelfdave:
I'll suggest moving the .gitignore changes to a separate PR.

helper/resource/testing.go
```diff
@@ -1031,6 +1031,36 @@ func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst strin
...
+	return checkIfIndexesIntoTypeSet(key, func(s *terraform.State) error {
+		is, err := primaryInstanceState(s, name)
+		if err != nil {
+			return err
+		}
+
+		valLen := len(is.Attributes[key])
+		if valLen < min || valLen > max {
+			return fmt.Errorf(
+				"%s: Attribute '%s' length is %d, not within the expected range [%d, %d]",
```
bookshelfdave, line 1046:
```suggestion
				"%s: Attribute '%s' length is %d, which is not within the expected closed range [%d, %d]",
```

helper/resource/testing.go
```diff
@@ -1031,6 +1031,36 @@ func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst strin
 	return nil
 }
 
+// TestRangeLengthResourceAttr is a TestCheckFunc which checks that the length of the value
+// in state for the given name/key is within an expected (closed) range
```
bookshelfdave, line 1035: I'm not sure if it matters in this repo, but the last linter PR I looked at wanted `.`'s at the end of comments.

helper/resource/testing_test.go
```diff
@@ -874,6 +874,87 @@ func TestCheckResourceAttr_empty(t *testing.T) {
...
+					Attributes: map[string]string{
+						resKey: resVal,
+					},
+				},
+			},
+		},
+	})
+
+	t.Run("matches_length", func(t *testing.T) {
+		check := TestMatchLengthResourceAttr(resName, resKey, 10)
```
bookshelfdave, line 933: Would it make more sense to refer to the [length](https://pkg.go.dev/unicode/utf8#RuneCountInString) of the `resVal` string rather than hardcoding 10?

helper/resource/testing.go
detro, line 1035:
```suggestion
// in state for the given name/key is within an expected (closed) range.
```

helper/resource/testing_test.go
detro: Well, it's on purpose: given that this test is especially focused on "string lengths", keeping it hardcoded makes it break if anything changes, like the string length.

As you can see below, the length becomes part of the errors we expect to see (i.e. `11 == string is too short` and `2 == string is too long`).

bookshelfdave (approved):

helper/resource/testing.go
```diff
@@ -1031,6 +1031,36 @@ func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst strin
...
+				min,
+				max)
+		}
+
+		return nil
+	})
+}
+
+// TestMatchLengthResourceAttr is a TestCheckFunc which checks that the length of the value
+// in state for the given name/key matches a specific length
```
detro, line 1059:
```suggestion
// in state for the given name/key matches a specific length.
```
```suggestion
// in state for the given name/key matches a specific length
```

PR #905, opened 2022-03-11
plugin: Always forward tf6server close channel

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-sdk/pull/857

Replicating the `tf5server` close channel handling right above it -- it was a copy-paste miss on my part. Without this, `tf6server` (such as terraform-plugin-framework providers) will hang until test timeouts. This hasn't been released yet, so does not require a CHANGELOG entry, but it is a blocker for releasing 2.11.0.

As part of troubleshooting this, was able to add some helpful trace and debug logging for future travelers.

### DIFF
--- a/.github/workflows/ci-go.yml
+++ b/.github/workflows/ci-go.yml
@@ -4,7 +4,7 @@ name: ci-go
 on:
   pull_request:
     paths:
-      - .github/workflows/test.yml
+      - .github/workflows/ci-go.yml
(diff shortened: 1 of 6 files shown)


bflad:
I'm going to switch the terraform-provider-corner integration testing to run all types of providers because they all inherently depend on this project for the testing, which would've caught this sooner.

detro (approved):

PR #911, opened 2022-03-22
helper/resource: Additional Go documentation for TestCheck and TestMatch functions

### LINKED ISSUE #885: testCheckResourceAttrSet() is broken for all but TypeString attributes

bflad:
Closes #885

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -99,9 +99,9 @@ func AddTestSweepers(name string, s *Sweeper) {
 //
 // Sweeper flags added to the "go test" command:
 //
-// -sweep: Comma-separated list of locations/regions to run available sweepers.
-// -sweep-allow-failues: Enable to allow other sweepers to run after failures.
-// -sweep-run: Comma-separated list of resource type sweepers to run. Defaults
+//     -sweep: Comma-separated list of locations/regions to run available sweepers.
+//     -sweep-allow-failues: Enable to allow other sweepers to run after failures.
+//     -sweep-run: Comma-separated list of resource type sweepers to run. Defaults
 //             to all sweepers.
 //
 // Refer to the Env prefixed constants for environment variables that further
@@ -573,16 +573,16 @@ func ParallelTest(t testing.T, c TestCase) {
 // This function will automatically find or install Terraform CLI into a
 // temporary directory, based on the following behavior:
 //
-// - If the TF_ACC_TERRAFORM_PATH environment variable is set, that Terraform
-//   CLI binary is used if found and executable. If not found or executable,
-//   an error will be returned unless the TF_ACC_TERRAFORM_VERSION environment
-//   variable is also set.
-// - If the TF_ACC_TERRAFORM_VERSION environment variable is set, install and
-//   use that Terraform CLI version.
-// - If both the TF_ACC_TERRAFORM_PATH and TF_ACC_TERRAFORM_VERSION environment
-//   variables are unset, perform a lookup for the Terraform CLI binary based
-//   on the operating system PATH. If not found, the latest available Terraform
-//   CLI binary is installed.
+//     - If the TF_ACC_TERRAFORM_PATH environment variable is set, that
+//        Terraform CLI binary is used if found and executable. If not found or
+//        executable, an error will be returned unless the
+//        TF_ACC_TERRAFORM_VERSION environment variable is also set.
+//     - If the TF_ACC_TERRAFORM_VERSION environment variable is set, install
+//        and use that Terraform CLI version.
+//     - If both the TF_ACC_TERRAFORM_PATH and TF_ACC_TERRAFORM_VERSION
+//        environment variables are unset, perform a lookup for the Terraform
+//        CLI binary based on the operating system PATH. If not found, the
+//        latest available Terraform CLI binary is installed.
 //
 // Refer to the Env prefixed constants for additional details about these
 // environment variables, and others, that control testing functionality.
@@ -763,13 +763,13 @@ func ComposeAggregateTestCheckFunc(fs ...TestCheckFunc) TestCheckFunc {
 // Use this as a last resort when a more specific TestCheckFunc cannot be
 // implemented, such as:
 //
-// - TestCheckResourceAttr: Equality checking of non-TypeSet state value.
-// - TestCheckResourceAttrPair: Equality checking of non-TypeSet state value,
-//                              based on another state value.
-// - TestCheckTypeSet*: Equality checking of TypeSet state values.
-// - TestMatchResourceAttr: Regular expression checking of non-TypeSet state
-//                          value.
-// - TestMatchTypeSet*: Regular expression checking on TypeSet state values.
+//     - TestCheckResourceAttr: Equality checking of non-TypeSet state value.
+//     - TestCheckResourceAttrPair: Equality checking of non-TypeSet state
+//        value, based on another state value.
+//     - TestCheckTypeSet*: Equality checking of TypeSet state values.
+//     - TestMatchResourceAttr: Regular expression checking of non-TypeSet
+//        state value.
+//     - TestMatchTypeSet*: Regular expression checking on TypeSet state values.
 //
 // For managed resources, the name parameter is combination of the resource
 // type, a period (.), and the name label. The name for the below example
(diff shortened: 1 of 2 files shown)


helper/resource/testing.go
```diff
@@ -752,10 +755,47 @@ func ComposeAggregateTestCheckFunc(fs ...TestCheckFunc) TestCheckFunc {
...
+//
+// For data sources, the name parameter is a combination of the keyword "data",
+// a period (.), the data source type, a period (.), and the name label. The
+// name for the below example configuration would be
+// "data.myprovider_thing.example".
+//
+//     data "myprovider_thing" "example" { ... }
+//
+// The key parameter is an attribute path in Terraform CLI 0.11 and earlier
+// "flatmap" syntax. Keys start with the attribute name of a top-level
```
detro, line 788: Does it make sense to make references to what is now an "ancient" version of Terraform? Do we have a public doc explaining this `"flatmap" syntax`?

helper/resource/testing.go
```diff
@@ -789,8 +829,41 @@ func testCheckResourceAttrSet(is *terraform.InstanceState, name string, key stri
...
+// attribute. Use the following special key syntax to inspect list, map, and
+// set attributes:
+//
+// - .{NUMBER}: List value at index, e.g. .0 to inspect the first element. Use
+//              the TestCheckTypeSet* and TestMatchTypeSet* functions instead
+//              for sets.
+// - .{KEY}: Map value at key, e.g. .example to inspect the example key value
+// - .#: Number of elements in list or set
+// - .%: Number of elements in map
```
detro, line 859: As above, an "end to end" example would put a nice bow here.

helper/resource/testing.go
```diff
@@ -844,8 +917,36 @@ func testCheckResourceAttr(is *terraform.InstanceState, name string, key string,
...
+//
+//     data "myprovider_thing" "example" { ... }
+//
+// The key parameter is an attribute path in Terraform CLI 0.11 and earlier
+// "flatmap" syntax. Keys start with the attribute name of a top-level
+// attribute. Use the following special key syntax to inspect underlying
+// values of a list or map attribute:
+//
+// - .{NUMBER}: List value at index, e.g. .0 to inspect the first element
+// - .{KEY}: Map value at key, e.g. .example to inspect the example key value
```
detro, line 944: About the example, as above.

helper/resource/testing.go
```diff
@@ -892,8 +993,39 @@ func testCheckNoResourceAttr(is *terraform.InstanceState, name string, key strin
...
+// "flatmap" syntax. Keys start with the attribute name of a top-level
+// attribute. Use the following special key syntax to inspect list, map, and
+// set attributes:
+//
+// - .{NUMBER}: List value at index, e.g. .0 to inspect the first element. Use
+//              the TestCheckTypeSet* and TestMatchTypeSet* functions instead
+//              for sets.
+// - .{KEY}: Map value at key, e.g. .example to inspect the example key value
+// - .#: Number of elements in list or set
+// - .%: Number of elements in map
```
detro, line 1023: ditto

helper/resource/testing.go
```diff
@@ -949,8 +1084,38 @@ func TestCheckModuleResourceAttrPtr(mp []string, name string, key string, value
...
```
detro, line 1118: ditto

detro (approved):
This is AWESOMELY USEFUL. I myself have learned more about those functions by reading this doc.

I left a couple of comments, but are nitpicks and it's ok if you don't feel like it.

helper/resource/testing.go
bflad: I would love to reference something, but I'm not sure if there's any terraform.io documentation left on it. 🙁  I'll do some digging, but I'm doubtful. In place of finding that, I'm wondering if [Go testable examples](https://go.dev/blog/examples) might help at least slightly here.

helper/resource/testing.go
```diff
@@ -752,10 +755,47 @@ func ComposeAggregateTestCheckFunc(fs ...TestCheckFunc) TestCheckFunc {
...
+//
+//     data "myprovider_thing" "example" { ... }
+//
+// The key parameter is an attribute path in Terraform CLI 0.11 and earlier
+// "flatmap" syntax. Keys start with the attribute name of a top-level
+// attribute. Use the following special key syntax to inspect underlying
+// values of a list or map attribute:
+//
+// - .{NUMBER}: List value at index, e.g. .0 to inspect the first element
+// - .{KEY}: Map value at key, e.g. .example to inspect the example key value
```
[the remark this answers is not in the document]
bflad: Yeah totally -- I meant to implement [Go testable examples](https://go.dev/blog/examples) for these as part of this update.

It's going to be a little awkward though, because its only recommended to implement these for Computed values, but I'll essentially have to show the configuration equivalent so it makes more sense looking at it.

helper/resource/testing.go
bflad: The closest thing I can find is: https://www.terraform.io/language/configuration-0-11/interpolation#attributes-of-other-resources, but I think linking to that would honestly be more confusing than anything. The flatmap bits were a detail of the previous state storage mechanism. I've uploaded some TestCheckFunc examples instead.

helper/resource/testing.go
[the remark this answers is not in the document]
bflad: I've uploaded some TestCheckFunc examples.

helper/resource/testing.go
bflad: I've uploaded some TestCheckFunc examples.

helper/resource/testing.go
bflad: I've uploaded some TestCheckFunc examples.

helper/resource/testing.go
bflad: I've uploaded some TestCheckFunc examples.

helper/resource/testing.go
bflad: I've uploaded some TestCheckFunc examples.

helper/resource/testing.go
```diff
@@ -752,10 +755,47 @@ func ComposeAggregateTestCheckFunc(fs ...TestCheckFunc) TestCheckFunc {
 	}
 }
 
-// TestCheckResourceAttrSet is a TestCheckFunc which ensures a value
-// exists in state for the given name/key combination. It is useful when
-// testing that computed values were set, when it is not possible to
-// know ahead of time what the values will be.
+// TestCheckResourceAttrSet ensures any value exists in the state for the
```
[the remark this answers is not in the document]
detro: Of course. Naming, especially that lasts the test of time, is hard.

detro (approved):
The use of Go testable examples is a great solution, much better than keeping it in code.

In Rust there is something similar: the compiler can be told that part of the doc is an executable example and so it handles that in a similar fashion to here.

detro:
Screenshot 2022-03-24 at 11 42 51
I was curious to see the test examples in action, so I ran `godoc` locally.

I think you might need to tweak the formatting a bit, as the godoc renderer is getting a bit confused in some points.

detro:
But godoc formatting aside, this is excellent.

PR #913, opened 2022-03-23
internal/plugintest: Set CHECKPOINT_DISABLE=1 in terraform-exec directly

### LINKED ISSUE #684: Remove wd.Setenv and wd.Unsetenv
These are not used in the SDK. Are they used anywhere else? If not, remove them.

This is a breaking change.

bflad:
Closes #684

### DIFF
--- a/.changelog/913.txt
+++ b/.changelog/913.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+helper/resource: Execute Terraform CLI commands during acceptance testing with `CHECKPOINT_DISABLE=1` set, removing extraneous calls to checkpoint.hashicorp.com to check for latest Terraform CLI version
+```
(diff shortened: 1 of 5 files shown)


detro (approved):
LG but I do have a question: why?

I looked at the linked issue, and it doesn't explain why we are removing this, how it came to be in the SDK, and why there shouldn't be calls in the `plugin.go` to disable it, while it's ok in the `plugintest` module.

bflad:
Terraform CLI performs HTTP call(s) to checkpoint.hashicorp.com to verify that you're running the latest version of Terraform CLI when you execute some (all?) commands. Additional documentation about this feature can be found on the [Terraform website](https://www.terraform.io/cli/commands#upgrade-and-security-bulletin-checks). Here's the [original issue for its removal](https://github.com/hashicorp/terraform-plugin-sdk/issues/640).

This removes the extraneous HTTP call since:

* The output is not visible at all to developers when running testing
* When explicitly running testing against versions of Terraform CLI via `TF_ACC_TERRAFORM_PATH` or `TF_ACC_TERRAFORM_VERSION`, this check may be purposefully irrelevant as the intention was to run a non-latest version
* Less networking calls the better as Checkpoint call failures could unnecessarily fail testing runs
* Developers may be working in environments that require explicit egress, so no sense making this call if it doesn't provide any particular value

detro:
Thank you, that is super helpful and makes perfect sense.

PR #915, opened 2022-03-23
helper/validation: Prevented panics with `ToDiagFunc()` function when used inside `Schema` type `Elem` field

### LINKED ISSUE #734: ToDiagFunc() panics when used on a list element
### SDK version

```
v2.4.0
```

```go
	"foo": {
		Type:     schema.TypeList,
		Optional: true,
		Elem: &schema.Schema{
			Type:             schema.TypeString,
			// ValidateFunc:     validation.StringIsNotEmpty, // Ok
			ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotEmpty), // Panics
		},
	}
}
```

```hcl
resource "some_resource" "valid" {
  foo = ["bar"]
}

resource "some_resource" "invalid" {
  foo = [""]
}
```

### Expected Behavior

Each list element should be validated and the provider should not crash.

### Actual Behavior

Using the deprecated `ValidateFunc` field to validate a list element works fine.
However, using `ValidateDiagFunc: validation.ToDiagFunc(…)` on a list element causes the following panic:

```
panic: interface conversion: cty.PathStep is cty.IndexStep, not cty.GetAttrStep

goroutine 59 [running]:
github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation.ToDiagFunc.func1(…)
        /…/github.com/hashicorp/terraform-plugin-sdk/v2@v2.4.0/helper/validation/meta.go:69
```

This is probably due to https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.4.0/helper/validation/meta.go#L69 not handling the case where `attr` is of `cty.IndexStep` type.

### Steps to Reproduce

Validate a list (map and set too ?) element using `ValidateDiagFunc: validation.ToDiagFunc(…)`.

bflad:
Closes #734


helper/validation/meta.go
```diff
@@ -66,8 +66,21 @@ func ToDiagFunc(validator schema.SchemaValidateFunc) schema.SchemaValidateDiagFu
 	return func(i interface{}, p cty.Path) diag.Diagnostics {
 		var diags diag.Diagnostics
 
-		attr := p[len(p)-1].(cty.GetAttrStep)
-		ws, es := validator(i, attr.Name)
+		// A practitioner-friendly key for any SchemaValidateFunc output.
+		// Generally this should be the last attribute name on the path.
+		// If not found for some unexpected reason, an empty string is fine
```
detro, line 71: Can you provide a bit of details about why it wouldn't not be, in some cases, that the last element in the path is not the one that implements `GetAttrStep`?
And if it's not, then what are those? `IndexStep`?

detro: FYI I dug a bit into `go-cty` to try to understand this a bit: I do get that those `*Step` structs are representation of operations that can be done on a `Path`, but I don't fully follow how they get assembled and why we would hit a scenario like the one described in the comment you added.

detro (approved):
It's great to see this bug (that I hit 2 weeks ago) getting fixed.
I left a question that is more for my learning than anything.

helper/validation/meta.go
bflad: > why it wouldn't not be, in some cases, that the last element in the path is not the one that implements GetAttrStep?

An `IndexStep` in `cty` represents a path inside a list, map, or set for one of the elements. For example, the path to the first element in a list is 0.

> I don't fully follow how they get assembled

Paths are assembled as part of decoding a value. The core folks would be best to ask how this works in terms of how a configuration/schema gets decoded to certain types and paths.

> why we would hit a scenario like the one described in the comment you added

If a new path type is added to cty. Just coding for future safety because the runtime code for providers should never introduce an unnecessary panic.

bflad: I tried out the [`forcetypeassert` linter](https://github.com/gostaticanalysis/forcetypeassert) in golangci-lint and it catches some other potentially problematic code, but unfortunately doesn't catch this particular code pattern of a type assertion on a slice element. Not sure its worth the effort of enabling it in this project, but maybe some of the others.

detro: Big 👍 from me on adding this linter everywhere.

PR #925, opened 2022-04-02
Improve "unsupported test mode" message

armsnyder:
The "unsupported test mode" message does not tell you which step of the test failed, unlike the other fatal errors in the `runNewTest` function here. This change adds the step number to the error message to make it more useful and more consistent.

### DIFF
--- a/helper/resource/testing_new.go
+++ b/helper/resource/testing_new.go
@@ -197,7 +197,7 @@ func runNewTest(ctx context.Context, t testing.T, c TestCase, helper *plugintest
 			continue
 		}
 
-		t.Fatal("Unsupported test mode")
+		t.Fatalf("Step %d/%d, unsupported test mode",  i+1, len(c.Steps))
 	}
 }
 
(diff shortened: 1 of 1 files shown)


bflad (approved):
Thanks so much for this, @armsnyder 🚀

helper/resource/testing_new.go
```diff
@@ -197,7 +197,7 @@ func runNewTest(ctx context.Context, t testing.T, c TestCase, helper *plugintest
 			continue
 		}
 
-		t.Fatal("Unsupported test mode")
+		t.Fatalf("Step %d/%d, unsupported test mode",  i+1, len(c.Steps))
```
bflad, line 200: To fix the linter.

```suggestion
		t.Fatalf("Step %d/%d, unsupported test mode", i+1, len(c.Steps))
```

armsnyder: Hehe thanks

Issue #863, opened 2022-01-27
Using the default logger causes duplicate timestamp and severity values in log entries


gdavison:
### SDK version

```
github.com/hashicorp/terraform-plugin-sdk/v2 v2.10.1
```

### Terraform version

```
Terraform v1.1.4
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v3.73.0
```

### Expected Behavior

Log lines should look like the following

> 2022-01-26T16:25:33.123-0800 [DEBUG]  provider.terraform-provider-aws_v3.73.0_x5: Waiting for state to become: [success]: timestamp=2022-01-26T16:25:33.123-0800

### Actual Behavior

Log lines look like the following

> 2022-01-26T16:25:33.123-0800 [INFO]  provider.terraform-provider-aws_v3.73.0_x5: 2022/01/26 16:25:33 [DEBUG] Waiting for state to become: [success]: timestamp=2022-01-26T16:25:33.123-0800

I assume the first set, `2022-01-26T16:25:33.123-0800 [INFO]  provider.terraform-provider-aws_v3.73.0_x5:` comes from the Terraform CLI.

The second set, `2022/01/26 16:25:33 [DEBUG]`, comes from the AWS provider.

Level inference is also not working, since the provider is specifying `DEBUG`, but Terraform is logging as `INFO`.

### Steps to Reproduce

1. Call `TF_LOG=DEBUG terraform plan` with a configuration using the AWS provider.

### Additional Information

We are using the standard logger as provided by `terraform-plugin-sdk`.

Disabling the date and time on the default logger by calling `log.SetFlags(0)` before `plugin.Serve(opts)` is a workaround.

bflad:
@gdavison out of curiosity, is this any better with terraform-plugin-sdk/v2@v2.12.0?

bflad:
With the following configuration:

```terraform
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "4.8.0"
    }
  }
}

provider "aws" {
  region = "us-west-2"
}

resource "aws_vpc" "test" {
  cidr_block = "10.0.0.0/16"
}
```

terraform-plugin-sdk/v2@v2.12.0 with the `log.SetFlags()` handling in terraform-provider-aws@4.8.0 returns these log messages:

```console
$ TF_LOG=TRACE terraform apply -auto-approve |& grep 'Waiting for state to become'
[...]

Issue #834, opened 2021-12-15
Consider Updating Code and Documentation for Go 1.18 any -> interface{} Alias


bflad:
### SDK version

```
v2.10.0
```

### Use-cases

This framework generally requires heavy usage of empty interface types (`interface{}`, `[]interface{}`, `map[string]interface{}`) when working with `helper/schema.ResourceData`. This syntax is a little verbose and is not intuitive for newer provider developers or Gophers coming from projects which typically use concrete types.

When Go 1.18 releases early next year, it will include the equivalent of:

```go
type any = interface{}
```

It seems like it would be great to update our code and examples to replace `interface{}` with `any`.

### Proposal

After Go 1.18 is released:

- Update code examples on https://terraform.io/docs/
- Update code examples on https://learn.hashicorp.com
- Update code in https://github.com/hashicorp/terraform-provider-scaffolding
- Create issues on HashiCorp maintained utility providers (as they may , if they are still using this framework

Since this framework must remain compatible with Go 1.17 until Go 1.19 is released later next year due to our compatibility promises, this codebase should not be updated yet.

### References

- https://github.com/golang/go/issues/33232

PR #940, opened 2022-04-13
all: Replace various log.Printf calls with tfsdklog calls

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-sdk/issues/863

By default, the standard library `log` package will include the short timestamp and date flags, which when output will prefix any messaging, e.g.

```text
# From log.Printf("[DEBUG] Test message")
2022/01/26 16:25:33 [DEBUG] Test message
```

When a provider is running in production and the log output is being sent across the go-plugin boundary, Terraform CLI generally expects a hc-log formatted entry (square bracket log level prefixed message) or will output the raw message. So in many cases, this class of logging shows up in production logs as the raw message and the log level is not recognized, which is confusing for practitioners and provider developers:

```text
2022-01-26T16:25:33.123-0800 [INFO] provider.terraform-provider-example_v9.9.9_x5: 2022/01/26 16:25:33 [DEBUG] Test Message: timestamp=2022-01-26T16:25:33.123-0800
```

Setting the `log` package default logger to not include those flags via `log.SetFlags(0)` early in the SDK lifecycle may represent a breaking change for any providers depending on the current behavior, so this easier class of fix is not acceptable.

Now that this SDK implements the `tfsdklog` package, logging can be converted to that style, which fixes the above issue and these messages will now also include all the gathered logging context up to the call such as the RPC name and a request UUID.

There are quite a few other `log.Printf()` calls across the project, however they are non-trivial to update as:

* The logging context is not present or not guaranteed to be present (as is the case with `(helper/resource.StateChangeConf).WaitContext()`)
* The functions/methods are exported, so changing the signature is an unacceptable change
* The functions/methods while unexported are eventually surfaced via an exported type/function, so it could indirectly require breaking changes

Given that `helper/resource.StateChangeConf`, which also underlies `helper/resource.Retry*` functionality, is quite common across provider usage, a separate custom logging solution may be introduced there as a separate change.

### DIFF
--- a/helper/resource/plugin.go
+++ b/helper/resource/plugin.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
 	"strings"
 	"sync"
@@ -330,7 +329,7 @@ func runProviderCommand(ctx context.Context, t testing.T, f func() error, wd *pl
 	// started.
 	err := f()
 	if err != nil {
-		log.Printf("[WARN] Got error running Terraform: %s", err)
+		logging.HelperResourceWarn(ctx, "Error running Terraform CLI command", map[string]interface{}{logging.KeyError: err})
 	}
 
 	logging.HelperResourceTrace(ctx, "Called wrapped Terraform CLI command")
--- a/helper/schema/grpc_provider.go
+++ b/helper/schema/grpc_provider.go
@@ -75,11 +75,11 @@ func (s *GRPCProviderServer) GetProviderSchema(ctx context.Context, req *tfproto
 	}
 
 	resp.Provider = &tfprotov5.Schema{
-		Block: convert.ConfigSchemaToProto(s.getProviderSchemaBlock()),
+		Block: convert.ConfigSchemaToProto(ctx, s.getProviderSchemaBlock()),
 	}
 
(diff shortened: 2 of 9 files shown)


helper/resource/plugin.go
```diff
@@ -330,7 +329,7 @@ func runProviderCommand(ctx context.Context, t testing.T, f func() error, wd *pl
 	// started.
 	err := f()
 	if err != nil {
-		log.Printf("[WARN] Got error running Terraform: %s", err)
+		logging.HelperResourceWarn(ctx, "Error running Terraform CLI command", map[string]interface{}{logging.KeyError: err})
```
detro, line 332: Nothing to change here, but for when we go 1.18 here, we could consider using `map[string]any{...}` instead. It's still not ideal (I wish we could have variadic argument list that could enforce even-ness), but a bit shorter.

What do you think?

Again, there is nothing to change here.

detro (approved):

helper/resource/plugin.go
bflad: I do agree that once we're minimum Go 1.18 we should change all the `interface{}` usage to `any` 😉  https://github.com/hashicorp/terraform-plugin-sdk/issues/834

detro: Of course there is already a story for this :D

PR #950, opened 2022-04-28
Adding `resource.TestCheckResourceAttrWith` test utility

### LINKED ISSUE #948: Provide `TestCheckFunc`s that allow developers to implement personalized checking of resource attribute values
### SDK version
```
2.14.0
```

### Use-cases

The implementations of `TestCheckFunc` provided in `helper/resource/testing.go` are, most of the time, enough to run acceptance tests against providers: a declarative approach, where comparisons and regexp matches are usually enough.

But every now and then there are cases where an `equal` or `regexp.Match` are not enough: this applies to values that are generated at each acceptance test execution, or that are so long (example: PEM certificates), that writing a regexp for it kinda makes it harder than in needs to be.

Other cases, a bit more complex, is when one of the attributes of a resource is an encoding of something, and the test needs to decode and check that the content matches expectations.

For this, the current helpers run short.

### Attempted Solutions

It's always possible to implement a `TestCheckFunc` (`func(s *terraform.State) error {}`) and dig into the state with approaches like:

* Adding `output`s to the TF config, and then dig into ` s.RootModule().Outputs[]` to get to the values
* Dig directly into the state via `s.RootModule().Resources[].Primary`

But this leads to have to write lots of boilerplate code that ensures the given `(address, key)` pair matches something reasonable. It can get very verbose, repetitive and error prone.

Plus, for cases where the test requires checking for things like string length, `regexp.Match` has limits around the amount of repetitions that can be used on a group: this leads to having to write very long and convoluted regexps.

### Proposal

Provide the developer with an additional `TestCheckFunc`, designed to do all the boilerplate validation internally, and then pass back the attribute value to a custom function (`customCheckFunc`): this way the developer can write the specific validation logic for that value, and then return an `error` in case it fails.

It would looks something like this:

```golang
// for root resources
func TestCheckResourceAttrWith(name, key string, customCheckFunc func(value string) error) TestCheckFunc {...}

// for module resources
func TestCheckModuleResourceAttrWith(mp []string, name, key string, customCheckFunc func(value string) error) TestCheckFunc {...}
```

detro:
Closes #948

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -13,10 +13,11 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-multierror"
-	testing "github.com/mitchellh/go-testing-interface"
+	"github.com/mitchellh/go-testing-interface"
 
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
 	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/internal/addrs"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/internal/logging"
@@ -49,7 +50,7 @@ var flagSweepAllowFailures = flag.Bool("sweep-allow-failures", false, "Enable to
 var flagSweepRun = flag.String("sweep-run", "", "Comma seperated list of Sweeper Tests to run")
 var sweeperFuncs map[string]*Sweeper
 
-// type SweeperFunc is a signature for a function that acts as a sweeper. It
+// SweeperFunc is a signature for a function that acts as a sweeper. It
 // accepts a string for the region that the sweeper is to be ran in. This
 // function must be able to construct a valid client for that region.
 type SweeperFunc func(r string) error
@@ -957,6 +958,34 @@ func testCheckResourceAttr(is *terraform.InstanceState, name string, key string,
 	}
 
 	return nil
+}
+
+// CheckResourceAttrWithFunc is the callback type used to apply a custom checking logic
+// when using TestCheckResourceAttrWith and a value is found for the given name and key.
+//
+// When this function returns an error, TestCheckResourceAttrWith will fail the check.
+type CheckResourceAttrWithFunc func(value string) error
+
+// TestCheckResourceAttrWith - as per TestCheckResourceAttr but the attribute value
+// checking logic can be customised.
+//
+// If a value is found for the given name and key, it is passed to the CheckResourceAttrWithFunc function.
+// The CheckResourceAttrWithFunc function can then apply any checking logic:
+// should return an error for the check to fail, or `nil` to succeed.
+func TestCheckResourceAttrWith(name, key string, f CheckResourceAttrWithFunc) TestCheckFunc {
+	return checkIfIndexesIntoTypeSet(key, func(s *terraform.State) error {
+		is, err := primaryInstanceState(s, name)
+		if err != nil {
+			return err
+		}
+
+		err = testCheckResourceAttrSet(is, name, key)
+		if err != nil {
+			return err
+		}
+
+		return f(is.Attributes[key])
+	})
 }
 
 // TestCheckNoResourceAttr ensures no value exists in the state for the
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -984,7 +984,13 @@ func TestCheckResourceAttrWith(name, key string, f CheckResourceAttrWithFunc) Te
 			return err
 		}
 
-		return f(is.Attributes[key])
+		err = f(is.Attributes[key])
+
+		if err != nil {
+			return fmt.Errorf("%s: Attribute %q value: %w", name, key, err)
+		}
+
+		return nil
 	})
 }
 
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -966,13 +966,42 @@ func testCheckResourceAttr(is *terraform.InstanceState, name string, key string,
 // When this function returns an error, TestCheckResourceAttrWith will fail the check.
 type CheckResourceAttrWithFunc func(value string) error
 
-// TestCheckResourceAttrWith - as per TestCheckResourceAttr but the attribute value
-// checking logic can be customised.
+// TestCheckResourceAttrWith ensures a value stored in state for the
+// given name and key combination, is checked against a custom logic.
+// State value checking is only recommended for testing Computed attributes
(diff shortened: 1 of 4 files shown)


helper/resource/testing_test.go
```diff
@@ -1402,6 +1402,444 @@ func TestTestCheckResourceAttr(t *testing.T) {
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			key:           "test_map_attribute",
+			value:         "test-value",
+			expectedError: fmt.Errorf("map attribute 'test_map_attribute' must be checked by element count key (test_map_attribute.%%) or element value keys (e.g. test_map_attribute.examplekey)"),
```
bendbennett, line 1632: Should `test_map_attribute.%%` be `test_map_attribute.%`?

bendbennett (approved):

helper/resource/testing_test.go
detro: In golang fmt strings `%%` it's the escape sequence for `%` I believe.

bendbennett: 🙃 more coffee required!

helper/resource/testing.go
```diff
@@ -959,6 +960,34 @@ func testCheckResourceAttr(is *terraform.InstanceState, name string, key string,
...
+//
+// When this function returns an error, TestCheckResourceAttrWith will fail the check.
+type CheckResourceAttrWithFunc func(value string) error
+
+// TestCheckResourceAttrWith - as per TestCheckResourceAttr but the attribute value
+// checking logic can be customised.
+//
+// If a value is found for the given name and key, it is passed to the CheckResourceAttrWithFunc function.
+// The CheckResourceAttrWithFunc function can then apply any checking logic:
+// should return an error for the check to fail, or `nil` to succeed.
```
bflad, line 974: I like this description here, I'm just wondering if we should also give provider developers all the usage information when they reference it in Go documentation or via the Go language server (e.g. hovering in an editor), similar to the other `TestCheckFunc`

```go
// TestCheckResourceAttrWith ensures a specific value is stored in state for
// the given name and key combination, by checking it against a custom
// validation function. State value checking is only recommended for testing
// Computed attributes and attribute defaults.
//
// For managed resources, the name parameter is combination of the resource
// type, a period (.), and the name label. The name for the below example
// configuration would be "myprovider_thing.example".
//
//     resource "myprovider_thing" "example" { ... }
//
// For data sources, the name parameter is a combination of the keyword "data",
// a period (.), the data source type, a period (.), and the name label. The
// name for the below example configuration would be
// "data.myprovider_thing.example".
//
//     data "myprovider_thing" "example" { ... }
//
// The key parameter is an attribute path in Terraform CLI 0.11 and earlier
// "flatmap" syntax. Keys start with the attribute name of a top-level
// attribute. Use the following special key syntax to inspect list, map, and
// set attributes:
//
//     - .{NUMBER}: List value at index, e.g. .0 to inspect the first element.
//        Use the TestCheckTypeSet* and TestMatchTypeSet* functions instead
//        for sets.
//     - .{KEY}: Map value at key, e.g. .example to inspect the example key
//        value.
//
// The f parameter is a function containing the custom value validation logic.
// Any returned errors from that function will cause the check and test to fail.
```

helper/resource/testing.go
```diff
@@ -959,6 +960,34 @@ func testCheckResourceAttr(is *terraform.InstanceState, name string, key string,
...
+		if err != nil {
+			return err
+		}
+
+		err = testCheckResourceAttrSet(is, name, key)
+		if err != nil {
+			return err
+		}
+
+		return f(is.Attributes[key])
```
bflad, line 987: It'd be awesome to split this so the testing framework can add error context with the name and key automatically:

```suggestion
		err = f(is.Attributes[key])

		if err != nil {
			return fmt.Errorf("%s: Attribute %q value: %w", name, key, err)
		}

		return nil
```

helper/resource/testing_example_test.go
```diff
@@ -209,6 +211,67 @@ func ExampleTestCheckResourceAttr_typeString() {
...
+	//
+	// NOTE: State value checking is only necessary for Computed attributes,
+	//       as the testing framework will automatically return test failures
+	//       for configured attributes that mismatch the saved state, however
+	//       this configuration and test is shown for illustrative purposes.
+
+	// Verify the attribute value string length is above 1000
+	resource.TestCheckResourceAttrWith("example_thing.test", "example_string_attribute", func(value string) error {
+		if len(value) <= 1000 {
+			return fmt.Errorf("example_string_attribute should be longer than 1000 characters")
```
bflad, line 236: Paired with the TestCheckResourceAttrWith error context change above

```suggestion
			return fmt.Errorf("should be longer than 1000 characters")
```

helper/resource/testing_example_test.go
```diff
@@ -209,6 +211,67 @@ func ExampleTestCheckResourceAttr_typeString() {
...
+
+	// Verify the attribute value is an integer, and it's between 5 (included) and 20 (excluded)
+	resource.TestCheckResourceAttrWith("example_thing.test", "example_string_attribute", func(value string) error {
+		valueInt, err := strconv.Atoi(value)
+		if err != nil {
+			return err
+		}
+
+		if valueInt < 5 && valueInt >= 20 {
+			return fmt.Errorf("example error message")
```
bflad, line 269:
```suggestion
			return fmt.Errorf("should be between 5 and 20")
```

bflad (commented):
Overall I think this looks pretty good, just a few minor things to consider.

helper/resource/testing.go
detro: Makes sense. I'll update.

detro: I have provided a much more comprehensive godoc for this function.

Let me know your thoughts

bflad (approved):
Looks good to me 🚀

helper/resource/testing_example_test.go
```diff
@@ -209,6 +211,67 @@ func ExampleTestCheckResourceAttr_typeString() {
...
+	// The following TestCheckResourceAttrWith can be written to assert against
+	// the expected state values.
+	//
+	// NOTE: State value checking is only necessary for Computed attributes,
+	//       as the testing framework will automatically return test failures
+	//       for configured attributes that mismatch the saved state, however
+	//       this configuration and test is shown for illustrative purposes.
+
+	// Verify the attribute value is an integer, and it's between 5 (included) and 20 (excluded)
+	resource.TestCheckResourceAttrWith("example_thing.test", "example_string_attribute", func(value string) error {
```
inkel, line 262: Shouldn't the second parameter be `example_int_attribute` in this case?

PR #961, opened 2022-05-10
CHANGELOG for 2.16.0

detro:


bendbennett (approved):
LGTM

.changelog/960.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Updated version of [hc-install](github.com/hashicorp/hc-install) in response to change in HashiCorp Release API [sending back a different `Content-Type` header](https://github.com/hashicorp/hc-install/issues/56). This was causing failures when the SDK attempted to install Terraform.
```
bflad, line 2: Nit: affects helper/resource

bflad (approved):
Looks good to me

Issue #536, opened 2020-08-14
Acceptance Testing Framework for_each Resource Support (unexpected index type (string))


bflad:
### SDK version

```
v2.0.1
```

### Use-cases

For parity with real world Terraform configurations, we would like to utilize resource `for_each` support in acceptance testing configurations for the Terraform AWS Provider, to automatically validate documented configuration practices.

Essentially, we would like to enable the commented configuration below, which matches our resource and examples documentation.

https://github.com/terraform-providers/terraform-provider-aws/blob/d104477b8209c4829d1d9601ca3cac7e5cf1cdd5/aws/resource_aws_acm_certificate_validation_test.go#L202-L245

### Attempted Solutions

The acceptance testing framework currently throws an error when applying a test configuration with `for_each` usage:

```
    TestAccAWSAcmCertificateValidation_basic: testing_new_config.go:52: unexpected index type (string) for "aws_route53_record.test", for_each is not supported
    TestAccAWSAcmCertificateValidation_basic: testing_new.go:57: unexpected index type (string) for "aws_route53_record.test", for_each is not supported
--- FAIL: TestAccAWSAcmCertificateValidation_basic (70.91s)
```

Since test configuration workarounds are currently applied, any resources changes will require manual re-verification of certain functionality.

### Proposal

Support `for_each` indexed resources in the acceptance test configurations, similar to `count` resources. In our case, we don't necessarily need all other testing functionality to support `for_each` indexed resources (e.g. `TestCheckFunc`), just the successful apply of the configuration would be great for our situation.

### References

- https://github.com/terraform-providers/terraform-provider-aws/pull/14199
- https://registry.terraform.io/providers/hashicorp/aws/3.1.0/docs/resources/acm_certificate_validation#dns-validation-with-route-53

paddycarver:
I'm not going to lie, I'm not _entirely_ sure why this doesn't work right now. In theory, it should--the test framework shouldn't care about for_each at all, it should never even see it. Terraform should just handle it.

I have a sneaking suspicion there's a deeper problem here or the SDK has trouble handling the output of for_each or something. But we're going to need to do a bit more digging to figure out what's going on here, because you definitely shouldn't be getting apply errors from configs that standalone Terraform can apply; after all, the whole point of the new test driver is that standalone Terraform _is_ applying them.

akselleirv:
Any update on this issue? I would also like to add tests for resources using `for_each`.

[...]

PR #965, opened 2022-05-11
docs: Add mention about for_each limitation in TestSteps

remyleone:
https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/resource/state_shim.go#L168


remyleone:
https://github.com/hashicorp/terraform-plugin-sdk/issues/536

remyleone:
@austinvalle 👋 what do you think about it?

austinvalle (approved):
Thanks for this @remyleone ✅ 

This is unforuntately still a problem despite the failing code blocking `for_each` support not being needed for the [newer state check functionality](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/state-checks/resource). It will likely be an issue until we get around to releasing a v2 of the test framework 🙁

PR #972, opened 2022-05-26
helper/resource: Support TestStep provider handling

### LINKED ISSUE #253: Testing tools for state migrations
### Current Terraform Version
```
Terraform v0.12.13
```

### Use-cases
Developing state migrations!

There are two closely related requests here: 

1. The `StateUpgrader` seems to be the only place in the plugin SDK where a developer is asked to interact directly with a state object in the form of a `map[string]interface{}`. This type of structure (as far as I can tell) is not mentioned anywhere else in the Terraform Custom Provider docs, and no formal explanation is given of how its contents relate to the `schema.Schema` and how to manipulate them. The [documentation on state migrations](https://www.terraform.io/docs/extend/resources/state-migration.html) is extremely limited - it only contains one example, which is a trivial one, and doesn't providing any insight into manipulating the state object. I suspect that an understanding of the state object could be gleaned by understanding the `ctype` package and how it's used by the GRPC server, but these are internal details that are not part of the SDK API, and I think that knowledge needs to be illuminated explicitly in the documentation.

2. When a Terraform provider responds to an RPC request with data that doesn't match the expected schema, the error messages are completely opaque. This doesn't affect normal operation, since of course there should never be responses that don't match the schema, but it makes developing and testing state migrations very difficult.

### Why is this coming up

I'm currently working on fixing a bug in [`terraform-provider-docker`](github.com/terraform-providers/terraform-provider-docker) that was introduced by the HCL2 migration and requires a breaking change to the schema structure. When attempting to test my migrators, they all run successfully, but Terraform produces the following errors:

```
Error: missing expected [

Error: missing expected [
```

By hacking debug statements and stacktraces into my `vendor`, I've managed to establish that this happens in the gRPC server while attempting to marshal the newly migrated state object, indicating that my migrated state doesn't agree with my new schema. However, the error message tells me nothing about _what part_ of the state object was invalid, not even which resource failed to migrate. Stack traces also don't help, because this marshaling takes place in a goroutine.

### Proposal
I think the opaque errors is probably out of reach to fix, but I'd like to hear other's opinions on that. It seems like it would help to (1) validate the schema and (2) marshal the JSON in two separate steps, so that an schema mismatch doesn't result in an error that walks and quacks like a JSON syntax error. However, that's just not the way `ctype` does things, and a complete re-write of the `ctype` project is pretty out of scope for a Terraform feature request.

The docs thing certainly doesn't sound to me like too huge of an ask if there's somebody out there familiar with the state object willing to write it up. However, perhaps a more powerful alternative would be to leave the docs as they are and expand the suite of testing tools for Terraform providers to include a testbed for state migrations. This would provide an environment to validate new migrations without involving any GRPC interactions, and so it would be an opportunity to provide better error messages without re-writing `ctpye`.

### LINKED ISSUE #628: Allow multiple versions of provider in acceptance tests

### LINKED ISSUE #779: Support changing provider configurations during test steps
### SDK version

```
v2.7.0
```

### Use-cases

To be able to test upgrading providers between different versions, it would be helpful if we could specify a different set of providers per test step, not just in the test case.

### Proposal

Adding the `ProviderFactories`, `ProtoV5ProviderFactories`, `ProtoV6ProviderFactories`, and `ExternalProviders` properties from `resource.TestCase` to `resource.TestStep`.

bflad:
Closes #253
Closes #628
Closes #779

Provider developers can now select whether to configure providers for acceptance testing at the `TestCase` or `TestStep` level. Only one level may be used in this current implementation, however it may be possible to allow merged `TestCase` and `TestStep` configuration with additional validation logic to ensure a single provider is not specified multiple times across the merge result of all fields.

This change also introduces some upfront `TestCase` and `TestStep` configuration validation when calling any of the `Test` functions, failing the test early if a problem is detected. There are other validations that are possible, however these are considered out of scope.


bendbennett (approved):
Amazing!

Issue #609, opened 2020-10-09
ExpectError is ignored for test case where destroy action expected to have error


IronSavior:
Maybe I'm doing this wrong?

### SDK version
```
{
  "Path": "github.com/hashicorp/terraform-plugin-sdk/v2",
  "Version": "v2.0.2"
}
```

### Relevant provider source code

```go
// In provider resource DELETE function:
_, err := client.Remove(ctx, req)
if err != nil {
	logger.WithError(err).Error("could not remove")
	return diag.FromErr(err) // <-- Code path under test
}
```

```go
// In test case:
client.On("Remove", mock.Anything, req).Return(nil, &err) // Set up mock client to return error

resource.UnitTest(t, resource.TestCase{
	Providers: testAccProviders,
	Steps: []resource.TestStep{{
		Config:      loadFixtureString("testdata/%s.tf", t.Name()),
		ExpectError: regexp.MustCompile(`.*`), // <-- should match any error at all
		Check: resource.ComposeAggregateTestCheckFunc(
			// irrelevant
		),
	}},
})

```
### Debug Output
Test output:

```
    testing_new.go:22: WARNING: destroy failed, so remote objects may still exist and be subject to billing
    testing_new.go:22: failed to destroy: 
        Error: Cannot remove
```

### Expected Behavior

Test should pass because the delete operation is expected to fail

### Actual Behavior

Test fails because the resource is not destroyed.

### Steps to Reproduce

Create an apply/destroy test case where the destroy is guaranteed to fail.

### References

https://github.com/hashicorp/terraform-plugin-sdk/issues/347 - Except their problem was during resource creation.

sloloris:
I am hitting a similar issue. I want to test a case where the attempted deletion of a resource with dependencies should fail. The apply does fail with the expected error, but then the test itself errors out due to dangling resources:
```
    resource_launchdarkly_feature_flag_test.go:1027: Step 3/3 error: Error running apply: exit status 1
        
        Error: flag "prereq" in project "tx10w6a1r8" cannot be deleted: 400 Bad Request: {"code":"invalid_request","message":"A flag that is a prerequisite of other flags cannot be archived"}
        
        
    testing_new.go:70: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: failed to delete flag "prereq" from project "tx10w6a1r8": 404 Not Found: {"message":"Unknown project key tx10w6a1r8"}
```
[...]

PR #976, opened 2022-06-03
Add handling for destroy errors

matthewcummings:


matthewcummings:
I'm not sure exactly how to test this, it would be easy to try it out in my local env if I could branch off the main repo instead of my fork as the module namespacing breaks everything.

There's likely a workaround for doing this, please lmk.

bflad:
Hi @matthewcummings 👋 Thank you for submitting this.

Before we discuss any potential implementation details, let's continue the discussion in #609 to determine if and what should be done. The testing framework has long tried to prevent provider developers from purposefully creating situations which can leave dangling infrastructure that would require manual cleanup. If there's a bug when trying to use `Destroy: true` and `ExpectError` in the same `TestStep`, then that's likely a bug that should be resolved differently than introducing another field to the mix.

> it would be easy to try it out in my local env if I could branch off the main repo instead of my fork as the module namespacing breaks everything

You may find using the [`go mod edit -replace`](https://go.dev/ref/mod#go-mod-edit) command handy for this situation in a provider to override the location of the `github.com/hashicorp/terraform-plugin-sdk/v2` temporarily, e.g. for a local path:

```shell
go mod edit -replace github.com/hashicorp/terraform-plugin-sdk/v2=/path/to/local/terraform-plugin-sdk
```

Or remote reference:

```shell
go mod edit -replace github.com/hashicorp/terraform-plugin-sdk/v2=github.com/matthewcummings/terraform-plugin-sdk@check-destroy-error-handler
```

To undo it, the `replace` directive can be removed from the `go.mod` file or the `go mod edit -dropreplace` command can be used.

matthewcummings:
Thanks for the suggestion @bflad. I will hold off for now until a decision is made to consider this enhancement or not per the discussion in #609

matthewcummings:
FWIW, it was not a simple path/import replacement across all the .go files, I started down that path and hit nested import issues with the code in `/internal`.

PR #999, opened 2022-07-08
.github: Remove @hashicorp/team-tw-packer-and-terraform from CODEOWNERS

bflad:
Per team request, removing from automatic review requests. Will instead request reviews manually when desired.


detro (approved):
999... so close...

Issue #546, opened 2020-08-24
helper/logging/transport.go http.RoundTripper implementation allow correlating request and responses


edulop91:
### SDK version

Master branch

### Use-cases

I was spending some time debugging a performance issue with the TF Okta provider. As part of this exercise, I set the log TF_LOG level to trace. Through [this](https://github.com/hashicorp/terraform-plugin-sdk/blob/46a198e66e23b565ce2a74d7bdddcff64fb55971/helper/logging/transport.go#L17) code, we get nice output for request/response. Unfortunately, as far as I can tell there is no way to correlate a request with a given response in this http.RoundTripper implementation (especially in the `IsDebugOrHigher` branch) .

### Proposal

It would be nice to be able to add enough information to correlate a request/response pair. This could be as simple as adding a UUID that would be printed when `IsDebugOrHigher` for both the request and response output. I'm happy to submit the change if appropriate!

Alternatively, it would be really nice if we can start instrumenting terraform and providers with opentelemetry.

paddycarver:
We've opened #695 to pitch a proposal that I believe would enable this. We're very interested in feedback on that proposal, if people have any.

bflad:
terraform-plugin-sdk version 2.20.0 now includes a [`terraform-plugin-log`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-log) based `HTTP.Transport` that includes a unique identifier for each HTTP transaction in the `tf_http_trans_id` structured logging field.

Additional details about this functionality can be found on the website documentation: https://www.terraform.io/plugin/sdkv2/logging/http-transport

PR #1006, opened 2022-07-18
Introduce `NewLoggingHTTPTransport` and deprecate `NewTransport`

detro:
Related: https://github.com/hashicorp/terraform-provider-tfe/pull/479
Related: #2 
Closes: #546

## Background

Current SDK offers an helper `http.RoundTripper` in `helpers/logging`, that is created via `logging.NewTransport()`. This `http.RountTripper` can then be used as `.Transport` in an `http.Client`.

When logging level is _debug or higher_, it does a simple dump of the whole content of the request and of the following response.

This exposes any sensitive information that might be going over HTTP, and there is no way for developers using this code, to setup any filtering.

## Proposal

This PR introduced a 2 new implementations of `http.RoundTripper`, one designed to log via the provider root logger, the other via a user-defined subsystem.

```golang
func NewLoggingHTTPTransport(t http.RoundTripper) *loggingHttpTransport {
	return &loggingHttpTransport{"", false, t}
}

func NewSubsystemLoggingHTTPTransport(subsystem string, t http.RoundTripper) *loggingHttpTransport {
	return &loggingHttpTransport{subsystem, true, t}
}
```

This transport can be used exactly like the previous one, but it's built around the `tflog` package of [terraform-plugin-log](https://github.com/hashicorp/terraform-plugin-log), and offers an opt-in hook to configure the HTTP Request `context.Context`.

For example, if the user wants to setup log filtering against the provider root logger, and ensure that it's applied to each request handled by this transport:

```golang
// ctx == context.Context, provided by the SDK...

ctx = tflog.MaskFieldValuesWithFieldKeys(ctx, "MY_SENSITIVE_FIELD")
ctx = tflog.MaskMessageStrings(ctx, "MY_SECRET_STRING")

transport := logging.NewLoggingHTTPTransport(http.DefaultTransport)
client := http.Client{
  Transport: transport,
}

req, _ := http.NewRequest("GET", "https://www.terraform.io", nil)

res, err := client.Do(req.WithContext(ctx))
```

As a consequence, the existing `NewTransport` would be documented as deprecated.

## Related

* How to setup [log filtering](https://www.terraform.io/plugin/log/filtering) with `terraform-plugin-log`
* Fast follow-up: #1007

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,17 +22,17 @@ WEBSITE_DOCKER_RUN_FLAGS=--interactive \
 
 default: test
 
-test: fmtcheck generate
+test: generate
 	go test ./...
+
+lint:
+	golangci-lint run
 
 generate:
 	go generate ./...
 
 fmt:
-	gofmt -w $(GOFMT_FILES)
-
-fmtcheck:
-	@sh -c "'$(CURDIR)/scripts/gofmtcheck.sh'"
+	gofmt -s -w -e $(GOFMT_FILES)
 
 # Run the terraform.io website to preview local content changes 
 website:
@@ -55,4 +55,4 @@ website/build-local:
 	@docker build https://github.com/hashicorp/terraform-website.git\#$(WEBSITE_BRANCH) \
 		-t $(WEBSITE_DOCKER_IMAGE_LOCAL)
 
-.PHONY: default fmt fmtcheck generate test website website/local website/build-local
+.PHONY: default fmt lint generate test website website/local website/build-local
--- a/helper/logging/logging_http_transport.go
+++ b/helper/logging/logging_http_transport.go
@@ -34,7 +34,7 @@ import (
 // logging behaviour via the above-mentioned context: please see
 // https://www.terraform.io/plugin/log/writing.
 func NewLoggingHTTPTransport(t http.RoundTripper) *loggingHttpTransport {
-	return &loggingHttpTransport{"", false, t}
+	return &loggingHttpTransport{"", t}
 }
 
 // NewSubsystemLoggingHTTPTransport creates a wrapper around an *http.RoundTripper,
@@ -56,8 +56,11 @@ func NewLoggingHTTPTransport(t http.RoundTripper) *loggingHttpTransport {
 // This also gives the developer the flexibility to further configure the
 // logging behaviour via the above-mentioned context: please see
 // https://www.terraform.io/plugin/log/writing.
+//
+// Please note: setting `subsystem` to an empty string it's equivalent to
+// using NewLoggingHTTPTransport.
 func NewSubsystemLoggingHTTPTransport(subsystem string, t http.RoundTripper) *loggingHttpTransport {
-	return &loggingHttpTransport{subsystem, true, t}
+	return &loggingHttpTransport{subsystem, t}
 }
 
 const (
@@ -90,15 +93,14 @@ const (
 )
 
 type loggingHttpTransport struct {
-	subsystem      string
-	logToSubsystem bool
-	transport      http.RoundTripper
+	subsystem string
+	transport http.RoundTripper
 }
 
 func (t *loggingHttpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
 
-	//// Decompose the request bytes in a message (HTTP body) and fields (HTTP headers), then log it
+	// Decompose the request bytes in a message (HTTP body) and fields (HTTP headers), then log it
 	msg, fields, err := decomposeRequestForLogging(req)
 	if err != nil {
 		t.Error(ctx, "Failed to parse request bytes for logging", map[string]interface{}{
@@ -128,7 +130,7 @@ func (t *loggingHttpTransport) RoundTrip(req *http.Request) (*http.Response, err
 }
 
 func (t *loggingHttpTransport) Debug(ctx context.Context, msg string, fields ...map[string]interface{}) {
-	if t.logToSubsystem {
+	if t.subsystem != "" {
 		tflog.SubsystemDebug(ctx, t.subsystem, msg, fields...)
 	} else {
 		tflog.Debug(ctx, msg, fields...)
@@ -136,7 +138,7 @@ func (t *loggingHttpTransport) Debug(ctx context.Context, msg string, fields ...
 }
 
 func (t *loggingHttpTransport) Error(ctx context.Context, msg string, fields ...map[string]interface{}) {
-	if t.logToSubsystem {
+	if t.subsystem != "" {
 		tflog.SubsystemError(ctx, t.subsystem, msg, fields...)
 	} else {
 		tflog.Error(ctx, msg, fields...)
@@ -181,28 +183,31 @@ func decomposeResponseForLogging(res *http.Response) (string, map[string]interfa
 	fields[FieldHttpResponseStatusCode] = res.StatusCode
 	fields[FieldHttpResponseStatusReason] = res.Status
 
-	// Get the full body of the response:
-	// this is necessary because the http.Response `Body` is a ReadCloser,
-	// so the only way to read it's content without affecting the final consumer
-	// of this response, is to rely on `httputil.DumpRequestOut`,
-	// that handles duplicating the underlying buffers and make the extraction
-	// of these bytes transparent.
-	resBytes, err := httputil.DumpResponse(res, true)
+	// Set the headers as fields to log
+	for k, v := range res.Header {
+		if len(v) == 1 {
+			fields[k] = v[0]
+		} else {
+			fields[k] = v
+		}
+	}
+
+	// Read the whole response body
+	resBody, err := io.ReadAll(res.Body)
 	if err != nil {
 		return "", nil, err
 	}
 
-	// Create a reader around the response full body
-	resReader := textproto.NewReader(bufio.NewReader(bytes.NewReader(resBytes)))
+	// Make a copy of the response body bytes, to be returned as log message
+	msgBytes := make([]byte, len(resBody))
+	copy(msgBytes, resBody)
 
-	err = readHeadersIntoFields(resReader, fields)
-	if err != nil {
-		return "", nil, err
-	}
+	// Wrap the bytes from the response body, back into an io.ReadCloser,
+	// to respect the interface of http.Response, as expected by users of the
+	// http.Client
+	res.Body = io.NopCloser(bytes.NewBuffer(resBody))
 
-	// Read the rest of the body content,
-	// into what will be the log message
-	return readRestToString(resReader), fields, nil
+	return string(resBody), fields, nil
 }
 
 func readHeadersIntoFields(reader *textproto.Reader, fields map[string]interface{}) error {
--- a/helper/logging/logging_http_transport.go
+++ b/helper/logging/logging_http_transport.go
@@ -203,54 +203,17 @@ func decomposeRequestForLogging(req *http.Request) (map[string]interface{}, erro
 	// Create a reader around the request full body
 	reqReader := textproto.NewReader(bufio.NewReader(bytes.NewReader(reqBytes)))
 
-	err = readHeadersIntoFields(reqReader, fields)
+	err = fieldHeadersFromRequestReader(reqReader, fields)
 	if err != nil {
 		return nil, err
 	}
 
 	// Read the rest of the body content
-	fields[FieldHttpRequestBody] = readRestToString(reqReader)
+	fields[FieldHttpRequestBody] = bodyFromRestOfRequestReader(reqReader)
 	return fields, nil
 }
 
-func decomposeResponseForLogging(res *http.Response) (map[string]interface{}, error) {
-	fields := make(map[string]interface{}, len(res.Header)+4)
-	fields[FieldHttpOperationType] = OperationHttpResponse
-
-	fields[FieldHttpResponseProtoVersion] = res.Proto
-	fields[FieldHttpResponseStatusCode] = res.StatusCode
-	fields[FieldHttpResponseStatusReason] = res.Status
-
-	// Set the headers as fields to log
-	for k, v := range res.Header {
-		if len(v) == 1 {
-			fields[k] = v[0]
-		} else {
-			fields[k] = v
-		}
-	}
-
-	// Read the whole response body
-	resBody, err := io.ReadAll(res.Body)
-	if err != nil {
-		return nil, err
-	}
-
-	// Make a copy of the response body bytes, to be returned as log message
-	msgBytes := make([]byte, len(resBody))
-	copy(msgBytes, resBody)
-
-	// Wrap the bytes from the response body, back into an io.ReadCloser,
-	// to respect the interface of http.Response, as expected by users of the
-	// http.Client
-	res.Body = io.NopCloser(bytes.NewBuffer(resBody))
-
-	fields[FieldHttpResponseBody] = string(resBody)
-
-	return fields, nil
-}
-
-func readHeadersIntoFields(reader *textproto.Reader, fields map[string]interface{}) error {
+func fieldHeadersFromRequestReader(reader *textproto.Reader, fields map[string]interface{}) error {
 	// Ignore the first line: it contains non-header content
 	// that we have already captured.
 	// Skipping this step, would cause the following call to `ReadMIMEHeader()`
@@ -278,7 +241,7 @@ func readHeadersIntoFields(reader *textproto.Reader, fields map[string]interface
 	return nil
 }
 
-func readRestToString(reader *textproto.Reader) string {
+func bodyFromRestOfRequestReader(reader *textproto.Reader) string {
 	var builder strings.Builder
 	for {
 		line, err := reader.ReadContinuedLine()
@@ -289,4 +252,37 @@ func readRestToString(reader *textproto.Reader) string {
 	}
 
 	return builder.String()
+}
+
+func decomposeResponseForLogging(res *http.Response) (map[string]interface{}, error) {
+	fields := make(map[string]interface{}, len(res.Header)+4)
+	fields[FieldHttpOperationType] = OperationHttpResponse
+
+	fields[FieldHttpResponseProtoVersion] = res.Proto
+	fields[FieldHttpResponseStatusCode] = res.StatusCode
+	fields[FieldHttpResponseStatusReason] = res.Status
+
+	// Set the headers as fields to log
+	for k, v := range res.Header {
+		if len(v) == 1 {
+			fields[k] = v[0]
+		} else {
+			fields[k] = v
+		}
+	}
+
+	// Read the whole response body
+	resBody, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	// Wrap the bytes from the response body, back into an io.ReadCloser,
+	// to respect the interface of http.Response, as expected by users of the
+	// http.Client
+	res.Body = io.NopCloser(bytes.NewBuffer(resBody))
+
+	fields[FieldHttpResponseBody] = string(resBody)
+
+	return fields, nil
 }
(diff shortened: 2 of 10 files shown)


helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,210 @@
...
+		headerParts := strings.Split(parts[i], ": ")
+		fields[headerParts[0]] = headerParts[1]
+	}
+
+	// HTTP Response Body: the last part is always the body (can be empty)
+	return parts[len(parts)-1], fields
+}
+
+func decomposeResponseBytes(b []byte) (string, map[string]interface{}) {
+	parts := strings.Split(string(b), "\r\n")
```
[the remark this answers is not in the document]
detro: Will look into it

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,210 @@
...
+	"net/http/httputil"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+const (
+	// FieldHttpOperationType is the field key used by NewLoggingHTTPTransport when logging the type of operation via tflog.
+	FieldHttpOperationType = "__HTTP_OP_TYPE__"
```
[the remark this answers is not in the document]
detro: Totally random: I can adopt any format that is good for this context.

I was just looking for a conversation starter.

Makefile
```diff
@@ -22,17 +22,17 @@ WEBSITE_DOCKER_RUN_FLAGS=--interactive \
...
+	golangci-lint run
+
 generate:
 	go generate ./...
 
 fmt:
-	gofmt -w $(GOFMT_FILES)
-
-fmtcheck:
-	@sh -c "'$(CURDIR)/scripts/gofmtcheck.sh'"
```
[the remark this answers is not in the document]
detro: Felt like a perfect place for "leaving the campfire better than I found it".

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (t *loggingHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Create a new subsystem logging context from the request context
+	ctx := tflog.NewSubsystem(req.Context(), t.subsystem)
+
+	// If set, allow for further configuration of the new subsystem logging context
+	if t.configureReqCtx != nil {
+		ctx = t.configureReqCtx(ctx, t.subsystem)
+	}
```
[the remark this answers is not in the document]
detro: In the light of the above, this will be entirely removed.

helper/logging/logging_http_transport.go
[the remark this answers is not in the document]
detro: Actually, `ReadMIMEHeader()` doesn't like the full body of a req/res (what I have here), but only the sub-section that contains actual headers.

So I can't use it to read the `HTTP Request Line`, nor the `HTTP Message Status Line`.
I'll see if it makes sense to ue it for the headers though.

helper/logging/logging_http_transport.go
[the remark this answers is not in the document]
bflad: This is an excellent conversation starter, then! Maybe we should provide some guidance and document recommendation for field keys (e.g. in the /plugin/log/writing documentation?).

I actually do not mind to much whichever guidance we give, although I will throw it out there that while uppercase field keys can certainly make them easier to spot in amongst the other text, they _may_ come across as overly "aggressive" as in the plaintext version of logs, e.g.

```text
1970-01-01T00:00:00 [DEBUG] Example message: HTTP_OP_TYPE=HTTP_REQ HTTP_REQ_METHOD=GET HTTP_REQ_URI=https://terraform.io HTTP_REQ_VERSION=1.1
```

versus

```text
1970-01-01T00:00:00 [DEBUG] Example message: http_op_type=HTTP_REQ http_req_method=GET http_req_uri=https://terraform.io http_req_version=1.1
```

Just my personal opinion though. 😄

helper/logging/logging_http_transport.go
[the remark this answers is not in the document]
bflad: I was peeking at the `ReadResponse` function in https://go.dev/src/net/http/response.go, which looks like it does `ReadLine()` then `ReadMIMEHeaders()`, but then I realized you already have a [`http.Response` object](https://pkg.go.dev/net/http#Response). Can we just use that information? How about with the request?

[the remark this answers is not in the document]
bflad: The usage of `httputil.Dump*` functionality in the original implementation may have been to just simplify the plaintext-only representation there. Since we want to provide richer information, I think we can use the higher-level abstractions available in `http.Request` and `http.Response`. 😄

detro:
I have implemented all the feedback of the PR review so far.

I still need to:

- re-do a test that uses log filtering
- update the PR proposal above with the actual implementation
- website doc

Additionally, I have set explicitly the version of golangci-lint (`1.47.1`), given that trying to get the GH Action to pick automatically the latest version didn't work (caching?).

I can revert this once we are reaching a mergeable state for this PR.

helper/logging/logging_http_transport.go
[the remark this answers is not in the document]
detro: I haven't finished yet, but I think you are right: for `http.Response` shouldn't be too tricky to just use the struct as is.

For `http.Request` I don't know yet, but will give it another go.

The key issue though seems to be the bit that those `Dump*` function did with the body: being the body a `ReadCloser`, it's tricky to grab that without interfering with the request/response. Those `Dump*` utilities do lots of "tricks" around the body to ensure they return a copy, without affecting the underlying buffer and the expectation of http client that is sending/receiving.

I'll have to do the same to avoid disrupting things, and this is the main reason I was sticking with the `Dump*` methods: splitting strings around the most well-formed standard heading in the industry, felt a safe long term bet.

But 🙆‍♂️ .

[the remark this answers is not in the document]
detro: Update (focusing on the response right now, but request isn't much different).

1. A detail that we missed so far is that `Response.Body` contains the WHOLE body of the response. Meaning, while I can use `Response.Header`, `Response.Status` etc. to populate most fields, `Response.Body` contains the whole body of the request (html/json/whatever), plus all the same headers above it.

2. In addition to that, the process to read the bytes out of a `Response.Body`, while preserving the response to that the client can consume it, it's requiring me to re-implement `httputil.DumpResponse` entirely. So I'm planning to keep using it instead, to grab the `Body`, while using the `Response.Header` (etc) fields to populate the other fields.

What do you advice to do about 1. and what do you think of 2.?

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,247 @@
...
+	fields := make(map[string]interface{}, len(req.Header)+4)
+	fields[FieldHttpOperationType] = OperationHttpRequest
+
+	fields[FieldHttpRequestMethod] = req.Method
+	fields[FieldHttpRequestUri] = req.URL.RequestURI()
+	fields[FieldHttpRequestProtoVersion] = req.Proto
+
+	// Get the full body of the request, including headers appended by http.Transport:
+	// this is necessary because the http.Request at this stage doesn't contain
+	// all the headers that will be eventually sent.
```
bflad, line 156: It's a bummer that these headers go beyond just `User-Agent` (which maybe we should set one to mention `terraform-plugin-sdk`, if its not already set?), since that would certainly simplify this logic.

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,247 @@
...
+//
+// It's responsibility of the developer using this transport, to ensure that each
+// http.Request it handles is configured with the SDK-initialized Provider Root Logger
+// context.Context, that it's passed to all resources/data-sources/provider entry-points
+// (i.e. schema.Resource fields like `CreateContext`, `ReadContext`, etc.).
+//
+// This also gives the developer the flexibility to further configure the
+// logging behaviour via the above-mentioned context: please see
+// https://www.terraform.io/plugin/log/writing.
+func NewLoggingHTTPTransport(t http.RoundTripper) *loggingHttpTransport {
```
bflad, line 36: Can you ensure there is a changelog file added for these two new function enhancements and a note about the deprecation of the existing transport? Thanks!

helper/logging/logging_http_transport_test.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	if len(entries) != 2 {
+		t.Fatalf("unexpected amount of logs produced; expected 2, got %d", len(entries))
+	}
+
+	reqEntry := entries[0]
+	if diff := cmp.Diff(reqEntry, map[string]interface{}{
+		"@level":              "debug",
+		"@message":            "An example multiline request body",
```
bflad, line 48: I wonder if instead of putting the body into the message field, we should instead put it into its own `tf_http_req_body` (or whatever) field and set the message to something like "Sending HTTP request"?

helper/logging/logging_http_transport_test.go
```diff
@@ -0,0 +1,233 @@
...
+		"tf_http_res_status_reason": "200 OK",
+	}
+	for ek, ev := range expectedResEntryFields {
+		if resEntry[ek] != ev {
+			t.Fatalf("Unexpected value for field %q; expected %q, got %q", ek, ev, resEntry[ek])
+		}
+	}
+
+	expectedNonEmptyEntryFields := []string{
+		"@message", "Etag", "Date", "X-Frame-Options", "Server",
```
bflad, line 79: Similar note about potentially using a new `tf_http_res_body` field and setting message to "Received HTTP response" so its clear what folks are seeing in the logs.

bflad (commented):
Overall I think this is almost there -- just a few little things then should be good to go. 👍

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,247 @@
...
+	// FieldHttpResponseStatusCode is the field key used by NewSubsystemLoggingHTTPTransport when logging a response status code via tflog.
+	FieldHttpResponseStatusCode = "tf_http_res_status_code"
+
+	// FieldHttpResponseStatusReason is the field key used by NewSubsystemLoggingHTTPTransport when logging a response status reason phrase via tflog.
+	FieldHttpResponseStatusReason = "tf_http_res_status_reason"
+)
+
+type loggingHttpTransport struct {
+	subsystem      string
+	logToSubsystem bool
```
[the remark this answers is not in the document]
detro: I was conflicted if to do that or not. 
I didn't want to have a case where user sets the subsystem name to `""` and get an unexpected fall back (though, it's a heck of a corner case).

I'll refactor and add an error or something.

helper/logging/logging_http_transport.go
detro: yep, now that signatures are agreed, I can do that.

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,247 @@
...
+	fields[FieldHttpResponseProtoVersion] = res.Proto
+	fields[FieldHttpResponseStatusCode] = res.StatusCode
+	fields[FieldHttpResponseStatusReason] = res.Status
+
+	// Get the full body of the response:
+	// this is necessary because the http.Response `Body` is a ReadCloser,
+	// so the only way to read it's content without affecting the final consumer
+	// of this response, is to rely on `httputil.DumpRequestOut`,
+	// that handles duplicating the underlying buffers and make the extraction
+	// of these bytes transparent.
```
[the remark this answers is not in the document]
detro: Let me see if I can refactor this way. Would certainly simplify things.

helper/logging/logging_http_transport_test.go
detro: I think we talked in zoom briefly about this. The only argument I could find for this, was that the "log masking" can be applied to a portion of the `@message` this way.
If we make the body a field, it's a "all or nothing": either they mask the whole value, or not.

I can go either way. I just feel that, in the case of masking, it would be "too blunt". 

What's your preference?

helper/logging/logging_http_transport_test.go
detro: As above.
I can go this way, I don't mind. Just wanted a final check about the log masking limitations we would encounter.

helper/logging/logging_http_transport.go
detro: Yeah, that's why this `http.DumpRequestOut` utility is so tricky, because internally it essentially "pretends" to send the request, and grabs the output at the "end of the pipe" (if it makes sense).

detro: But at least you made me get rid of manual parsing and relying on `textproto`, and that's good.

helper/logging/logging_http_transport.go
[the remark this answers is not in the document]
detro: Yeah, I think this will work. Thank you.

helper/logging/logging_http_transport.go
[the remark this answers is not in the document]
detro: I went for a godoc comment that warns of the "fallback" behaviour.

helper/logging/logging_http_transport_test.go
bflad: Aye, yes given the current filtering functionality, that does make this a little less developer friendly. 🙁 Maybe it is worth getting this developer-facing problem sorted before potentially changing it later? Flipping this problem around, being able to also have the option to redact the entire body via field key would be kinda nice (and is generally a feature for HTTP loggers)!

Regardless of where we land with this particular functionality, I think adding the ability to mask _portions_ of field values makes sense. Whether that masking applies to "any key" or "this specific key" is probably an important distinction, sigh. I'll create an upstream feature request issue for that. If distinguishing between message masking and field masking is too much of an ask, then maybe having a "just mask all the things" thing is appropriate.

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,252 @@
...
+
+	// Read the whole response body
+	resBody, err := io.ReadAll(res.Body)
+	if err != nil {
+		return "", nil, err
+	}
+
+	// Make a copy of the response body bytes, to be returned as log message
+	msgBytes := make([]byte, len(resBody))
+	copy(msgBytes, resBody)
```
bflad, line 203: Drive-by note: Do we _need_ to copy the bytes here? We are not modifying them and they could potentially be "a lot" of memory. Having a reader pointing at the same bytes we originally read shouldn't have any effect in this case, I think.

detro: I don't know: we would have a shared piece of memory between the logging library, and the user of that response body.

And if we make modification to that shared portion of memory, we end up changing it under the hood for the user of the response body too.

helper/logging/logging_http_transport_test.go
detro: I think it's unlikely that something that a user explicitly wants masked in a message, would be ok to leave plan in a field (and viceversa).

I think the correct thing to do would be to apply the same "find and replace" we do to messages, to the content of each key in the fields.

detro: All right, here is a PR to address the missing feature in terraform-plugin-log: https://github.com/hashicorp/terraform-plugin-log/pull/87

detro: Code that implements this, is now pushed.

The version of `terraform-plugin-log` has to be released before we can merge though.

helper/logging/logging_http_transport_test.go
detro: Done.

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,279 @@
...
+	// and NewSubsystemLoggingHTTPTransport when logging an HTTP response body via tflog.
+	FieldHttpResponseBody = "tf_http_res_body"
+
+	// MessageHttpRequest is the message used by NewLoggingHTTPTransport
+	// and NewSubsystemLoggingHTTPTransport when logging an HTTP request via tflog.
+	MessageHttpRequest = "Sending HTTP Request"
+
+	// MessageHttpResponse is the message used by NewLoggingHTTPTransport
+	// and NewSubsystemLoggingHTTPTransport when logging an HTTP response via tflog.
+	MessageHttpResponse = "Received HTTP Response"
```
bflad, line 117: Should these messages be exported in the public interface? They're okay to be constants, but I'm not sure we want to signal any sort of compatibility or code-desirable use cases for the string values themselves.

helper/logging/logging_http_transport.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+type loggingHttpTransport struct {
+	subsystem string
+	transport http.RoundTripper
+}
+
+func (t *loggingHttpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
```
bflad, line 125: Taking another look at https://github.com/hashicorp/terraform-plugin-sdk/issues/546, I think we may want to consider one more thing there to cover the ask: generating our own "HTTP request ID". Having a separately generated ID means you can peek at a request or response, copy that ID value, then find anything related to that specific HTTP request/response.

This might be something like the `tf_req_id` log field, which is managed upstream by terraform-plugin-go to automatically generate a UUID value that gets persisted with the logger contexts for each RPC request/response. [Upstream code reference](https://github.com/hashicorp/terraform-plugin-go/blob/4a24b5d5377f223ca890cdada0b6fdbab7410d45/internal/logging/context.go#L58-L71).

I think we can separately decide whether we should be setting our own User-Agent if none is configured, so we can add that generated ID into there are well.

helper/logging/logging_http_transport.go
bflad: Downstream `io.Reader` consumers cannot modify the underlying bytes as far as I know and we control this implementation, but let's add a code comment explaining that choice then, e.g. "we're doing this in case its modified" 👍

bflad (approved):
Approving to unblock and this looks good overall, however I do think we should do something for #546 specifically (or not close it yet) as part of this. 👍

helper/logging/logging_http_transport.go
detro: I'll give it a quick test: maybe I have over-thought this and it's way easier than I thought.

helper/logging/logging_http_transport.go
detro: I was conflicted indeed. Yeah, I'll make them private, good call.

helper/logging/logging_http_transport.go
detro: I think it makes sense to add that UUID. But are you asking me to use the one set by `terraform-plugin-go` or to add a new UUID, generated for each Req/Res pair that goes through the Transport?

detro: To clarify: I'm asking because each request, comes in with a `Context` and we have documented (and will add it to the website doc), that they need to pass in the context they receive as part of the RPC callbacks into the provider code.

So, my assumption is, that UUID is... already there?

bflad: Sorry I wasn't clear 😅 I meant generate a new UUID for each HTTP request, separate from the other one. Each RPC may result in multiple HTTP requests so its good to provide separate IDs for each. 👍

detro: Ah, great.

Yeah, totally happy to do it, if that closes #546 properly: seems like a long overdue ask 👍

helper/logging/logging_http_transport.go
detro: Silly me: the string replacement we do during masking is going to generate a new string every time, so there is no chance of us messing up the same underlying byte buffer.

I can entirely remove this copying.

detro:
All right, I think this is addressing everything @bflad - Indeed I'm pretty happy we get to address #546 in its entirety. Also, got rid of the pointless copying around of the response buffer.

Even if you approved, I think I'll wait for a confirmation the UUID looks good before merging.

bflad (approved):
Looks good to me 🚀

PR #1013, opened 2022-07-27
helper/schema: Clarify Schema type Deprecated field usage

### LINKED ISSUE #1012: helper/schema: Clarify Deprecated Field Usage
### SDK version

```
2.19.0
```

### Use-cases

Currently, the [`helper/schema.Schema` type `Deprecated` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema.Deprecated) is a message that gets translated into a warning diagnostic during Terraform's validation phase:

```
Warning: Argument is deprecated

{Deprecated field string}
```

This handling however, due to limitations in the protocol, only occurs for `Required` or `Optional` attributes when they have a known configuration value. There is currently no effect when the attribute is `Computed`-only (read-only). The SDK already enables the `deprecated` boolean field available in the protocol, so if/when Terraform core does something with that protocol field in a future version, it'll have an effect on read-only attributes then.

### Proposal

Update the Go documentation and website documentation to ensure it denotes this limitation. We could go further and begin failing unit testing for attributes with `Deprecated` and only `Computed` via `schema.InternalValidate()`, however that feature can be treated separately, if desired.

### References

- https://github.com/hashicorp/terraform/issues/7569

bflad:
Closes #1012

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -311,11 +311,32 @@ type Schema struct {
 	// "parent_block_name.0.child_attribute_name".
 	RequiredWith []string
 
-	// Deprecated indicates the message to include in a warning diagnostic to
-	// practitioners when this attribute is configured. Typically this is used
-	// to signal that this attribute will be removed in the future and provide
-	// next steps to the practitioner, such as using a different attribute,
-	// different resource, or if it should just be removed.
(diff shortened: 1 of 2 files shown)


detro (approved):

bendbennett (approved):
LGTM

Issue #1014, opened 2022-07-28
ExpectError supports for Regexp is not sufficient in some cases


obourdon:
### SDK version

```
2.19.0
```

### Use-cases
See details in https://github.com/hashicorp/terraform-provider-aws/issues/25986

in case of early fatal AWS ASGs errors occurring after timeout, the error message is exactly the same
as reported with the fix provided plus a prefix
```
Waiting up to .*: Need at least 1 healthy instances in ASG(?ms).*SAME_ERROR_MESSAGE
```

but due to the fact that the basic Golang regex implementation does not
support Perl like regex and more specifically exclude pattern, I did not find any proper way using the
current SDK code to provide tests which fail under timeout and which succeed with early error reporting

### Attempted Solutions
See above

### Proposal
See https://github.com/hashicorp/terraform-plugin-sdk/pull/1015

### References
- https://github.com/hashicorp/terraform-provider-aws/issues/25986
- https://github.com/hashicorp/terraform-provider-aws/pull/25987

bflad:
Hi @obourdon 👋 Thank you for raising this.

For more advanced error checking/handling, there is the [`resource.TestCase` type `ErrorCheck` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource#TestCase.ErrorCheck), which accepts a `func(error) error` that can be used to skip tests when encountering certain errors in test steps, or in your particular use case, use custom logic to detect certain errors and `return nil` so the detected error causes the step to pass while other errors will still cause the step to fail.

Can you please try that in your situation?

obourdon:
@bflad many thanks for your comments. I indeed considered using this but it did not seem to fit what I was trying to achieve because I don't want the test to be skipped if the condition is met or not I really want to make the difference between getting:

- <SOME-COMMON-SUFFIX-REGEXP>
- <RESOURCE-TIMEOUT-CASE-ADDED-PREFIX>(?ms).*<SOME-COMMON-SUFFIX-REGEXP>

to see that either my test and fix fails/succeed

Unless I misunderstood your point, I don't think `ErrorCheck` field allows to achieve this

obourdon:
@bflad any progress on this ?

bflad:
[...]

PR #1015, opened 2022-07-28
Complex ExpectError handling functions

obourdon:
See the following links for details:
- https://github.com/hashicorp/terraform-plugin-sdk/issues/1014
- https://github.com/hashicorp/terraform-provider-aws/issues/25986
- https://github.com/hashicorp/terraform-provider-aws/pull/25987

The documentation has been also updated to include meaningful example for this new functionnality

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -494,7 +494,7 @@ type TestStep struct {
 	// ExpectError allows the construction of test cases that we expect to fail
 	// with an error. The specified regexp must match against the error for the
 	// test to pass.
-	ExpectError *regexp.Regexp
+	ExpectError interface{}
 
 	// PlanOnly can be set to only run `plan` with this configuration, and not
 	// actually apply it. This is useful for ensuring config changes result in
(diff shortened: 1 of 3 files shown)


bflad:
Hi @obourdon 👋 Thank you for submitting this.

In general, our preference for field types should always be to make them more type constrained, rather than less. Consumers encountering fields that accept `any`/`interface{}` are left without Go compiler help for their implementations and the additional burden to fully understand the implementation details of the underlying functionality by reading into the Go or website documentation further.

In this case, if we were to do anything here, the preference would be to create a new `ErrorCheck func(error) error` field at the `TestStep` level which mimics the existing behavior at the `TestCase` level. We'll also need to ensure there is covering unit testing of the expected behaviors. The testing framework validation logic can be augmented to raise an error if both `ExpectError` and `ErrorCheck` are defined in the same `TestStep`. If you're interested in _that_ type of implementation, then let's discuss that further in the open issue or in a new pull request.

Thanks again.

Issue #1030, opened 2022-08-12
Modify Underlying Logic Behind `TestStep.Taint` To Call The `terraform taint` Command


bendbennett:
### SDK version
```
v2.20.0
```

### Use-cases
When running tests that include `TestStep.Taint` resources do not appear to be tainted and do not undergo a destroy-create cycle.

### Proposal
This issue results from the fact that the underlying logic behind Taint was never migrated to call the terraform taint command via terraform-exec when the rest of the testing framework was migrated to actual Terraform commands.

A `(*WorkingDir)Taint()` func could be added to https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.20.0/internal/plugintest/working_dir.go and then the `testStepTaint()` func can be modified to accept `*plugintest.WorkingDir` so that `wd.Taint()` can be called within `testStepTaint()`.

### References
- #625

Issue #625, opened 2020-10-28
SDK 2.1.0 does not taint resource between TestSteps using TestStep.Taint


Didainius:
### SDK version

```
2.1.0
```

### Relevant provider source code

Main point here is Taint in Steps 2 and 3
```go
Steps: []resource.TestStep{
			// Step 1
			resource.TestStep{
				Config: configTextVM,
				Check: resource.ComposeAggregateTestCheckFunc(
					testAccCheckVcdVAppVmExists(netVappName, netVmName1, "vcd_vapp_vm.test-vm", &vapp, &vm),
					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "name", netVmName1),
					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "network.#", "0"),
....
				),
			},
			// Step 2 - join org domain must fail
			resource.TestStep{
				Taint:  []string{"vcd_vapp_vm.test-vm"},
				Config: configTextVMStep1,
				Check: resource.ComposeAggregateTestCheckFunc(
					testAccCheckVcdVAppVmExists(netVappName, netVmName1, "vcd_vapp_vm.test-vm", &vapp, &vm),
					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "name", netVmName1),
					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "network.#", "0"),
...
				),
			},
			// Step 3 - join org domain enabled
			resource.TestStep{
				Taint:  []string{"vcd_vapp_vm.test-vm"},
				Config: configTextVMStep2,
				// Our testing suite does not have Windows OS to actually try domain join so the point of this test is
				// to prove that values are actually set and try to be applied on vCD.
				ExpectError: regexp.MustCompile(`Join Domain is not supported for OS type .*`),
				Check: resource.ComposeAggregateTestCheckFunc(
					testAccCheckVcdVAppVmExists(netVappName, netVmName1, "vcd_vapp_vm.test-vm", &vapp, &vm),
					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "name", netVmName1),
					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "network.#", "0"),
...
				),
			},
```

### Terraform Configuration Files

NO HCL because it is an issue in testing framework and not bound to specific HCL

### Debug Output

Single line in log does log that resource will be tainted.
[...]

PR #1031, opened 2022-08-12
Modifying underlying logic behind `TestStep.Taint` to call the `terrafom taint` command

bendbennett:
Closes: #1030 
Closes: #625


helper/resource/teststep_providers_test.go
```diff
@@ -342,6 +344,66 @@ func TestTest_TestStep_ExternalProviders_To_ProviderFactories_StateUpgraders(t *
...
+			},
+		},
+	})
+
+	if idOne == idTwo {
+		t.Errorf("taint is not causing destroy-create cycle, idOne == idTwo: %s == %s", idOne, idTwo)
+	}
+}
+
+func ExtractResourceAttr(resourceName string, attributeName string, attributeValue *string) TestCheckFunc {
```
bendbennett, line 387: Perhaps we could move this to a non-test file so that it's available for provider developers?

bflad: Let's create a followup issue to gauge interest 👍 

Although we may want be careful to unexport this function here since it could easily wind up somewhere where it could be added to the public API.

```suggestion
func extractResourceAttr(resourceName string, attributeName string, attributeValue *string) TestCheckFunc {
```

.changelog/1031.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+helper/resource: `TestStep.Taint` now calls `terraform taint` command ([#1030](https://github.com/hashicorp/terraform-plugin-sdk/issues/1030))
```
bflad, line 2: I think this is more of a bug fix for existing functionality, rather than something new 😄 

``````suggestion
```release-note:bug
helper/resource: Fixed `TestStep` type `Taint` field usage to properly recreate resources
``````

File: helper/resource/teststep_providers_test.go [hunk 3]
```diff
@@ -342,6 +344,66 @@ func TestTest_TestStep_ExternalProviders_To_ProviderFactories_StateUpgraders(t *
...
+				Taint: []string{"random_string.test"},
+				Config: `resource "random_string" "test" {
+					length = 10
+				}`,
+				Check: ComposeAggregateTestCheckFunc(
+					TestCheckResourceAttr("random_string.test", "length", "10"),
+					ExtractResourceAttr("random_string.test", "id", &idTwo),
+				),
+			},
+		},
```
bflad, line 379: Rather than requiring the testing to lookup, download, and install an external provider, we can inline create one, e.g.

```suggestion
		ProviderFactories: map[string]func() (*schema.Provider, error){
			"random": func() (*schema.Provider, error) { //nolint:unparam // required signature
				return &schema.Provider{
					ResourcesMap: map[string]*schema.Resource{
						"random_id": {
							CreateContext: func(_ context.Context, d *schema.ResourceData, _ interface{}) diag.Diagnostics {
								d.SetId(time.Now().String())
								return nil
							},
							DeleteContext: func(_ context.Context, _ *schema.ResourceData, _ interface{}) diag.Diagnostics {
								return nil
							},
							ReadContext: func(_ context.Context, _ *schema.ResourceData, _ interface{}) diag.Diagnostics {
								return nil
							},
							Schema: map[string]*schema.Schema{},
						},
					},
				}, nil
			},
		},
		Steps: []TestStep{
			{
				Config: `resource "random_id" "test" {}`,
				Check: ComposeAggregateTestCheckFunc(
					extractResourceAttr("random_string.test", "id", &idOne),
				),
			},
			{
				Taint: []string{"random_id.test"},
				Config: `resource "random_id" "test" {}`,
				Check: ComposeAggregateTestCheckFunc(
					extractResourceAttr("random_string.test", "id", &idTwo),
				),
			},
		},
```

internal/plugintest/working_dir.go
```diff
@@ -303,6 +304,17 @@ func (wd *WorkingDir) Import(ctx context.Context, resource, id string) error {
 	return err
 }
 
+// Taint runs terraform taint
+func (wd *WorkingDir) Taint(ctx context.Context, address string) error {
+	logging.HelperResourceTrace(ctx, "Calling Terraform CLI taint command")
+
+	err := wd.tf.Taint(context.Background(), address)
+
+	logging.HelperResourceTrace(ctx, "Called Terraform CLI import command")
```
bflad, line 313:
```suggestion
	logging.HelperResourceTrace(ctx, "Called Terraform CLI taint command")
```

bflad (approved):
Otherwise, looks good to me 🚀

bflad:
Does this also close #625?

Issue #1038, opened 2022-08-22
Timeouts are not applied to a Data Source read during apply


crhino:
### SDK version

```
github.com/hashicorp/terraform-plugin-sdk/v2 v2.21.0
```

### Relevant provider source code

```go
func dataSourceAzurePeeringConnection() *schema.Resource {
	return &schema.Resource{
		Description: "The Azure peering connection data source provides information about a peering connection between an HVN and a peer Azure VNet.",
		ReadContext: dataSourceAzurePeeringConnectionRead,
		Timeouts: &schema.ResourceTimeout{
			Read: 35 * time.Minute,
		},
...
```
[Github link](https://github.com/hashicorp/terraform-provider-hcp/blob/25ffc77f95fcbed03ce471e75de32f5824abbc5e/internal/provider/data_source_azure_peering_connection.go#L19)

### Terraform Configuration Files

```hcl
data "hcp_azure_peering_connection" "peering" {
  hvn_link              = var.hvn.self_link
  peering_id            = hcp_azure_peering_connection.peering.peering_id
  wait_for_active_state = true
  timeouts {
    read = "1m"
  }
}
```

### Expected Behavior

I expect that the Read timeout value is correctly applied to the data source read (35 minutes in the Go code, 1 minute in the HCL example).

### Actual Behavior

For all data source reads, the default 20 minutes timeout appears to be applied.

### References

I attempted to dig a bit into the SDK code, and found the `ReadDataApply` [function](https://github.com/hashicorp/terraform-plugin-sdk/blob/e14d3b611f2e257d2a0862e5ea0f90ea96fd5bf8/helper/schema/resource.go#L930), which does not set the `data.timeouts` field before calling `read()`, unlike similar [functions such as](https://github.com/hashicorp/terraform-plugin-sdk/blob/e14d3b611f2e257d2a0862e5ea0f90ea96fd5bf8/helper/schema/resource.go#L1004-L1008) `RefreshWithoutUpgrade`.

If I am reading that correctly, then we are not correctly sourcing the timeouts set on the `Resource.Timeouts` field into the `ResourceData` struct?

bcmdarroch:
Thanks for investigating this! We'll work with the plugin team to get that upstream fix.

bflad:
Hi @crhino 👋 Thank you for raising this and sorry you ran into trouble here.

[...]

PR #1041, opened 2022-08-26
Check for practioner-set timeout during ReadDataApply

bcmdarroch:
Fixes:
- https://github.com/hashicorp/terraform-plugin-sdk/issues/1038
- https://github.com/hashicorp/terraform-provider-hcp/issues/377

To test, pull this commit into your provider's go.mod. Set any datasource timeout to 1 min. Apply.
```
var readTimeout = time.Minute * 1
...
	Read: &readTimeout,
},
```

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -939,6 +939,14 @@ func (r *Resource) ReadDataApply(
 		return nil, diag.FromErr(err)
 	}
 
+	rt := ResourceTimeout{}
+	if _, ok := d.Meta[TimeoutKey]; ok {
+		if err := rt.DiffDecode(d); err != nil {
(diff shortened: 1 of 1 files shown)


bflad:
Hi again @bcmdarroch! 👋 

It looks like over in #1038 the workaround of using `ReadWithoutTimeout` was determined. In general, not using practitioner-configurable data source timeouts would be our recommendation for this situation since it could lead to practitioner confusion or configuration ambiguity.

While this particular code change may start the resolution process should practitioner-configurable timeouts have a proper use case, there would also need to be some additional testing to ensure that this wouldn't potentially introduce unexpected panics or other unintended data source behaviors. Since those tests aren't currently included, I'm going to close this for now. If you or someone else is interested in implementing them, we'd be happy to take another look.

Thanks again!

Issue #895, opened 2022-02-25
Apply results in an incorrect Update for TypeSet with an unexpected empty element


prashantv:
### SDK version
```
github.com/hashicorp/terraform-plugin-sdk/v2 v2.10.1
```

### Relevant provider source code

Repo with a full provider + repro instructions: https://github.com/prashantv/terraform-set-null-repro

Relevant provider code:
https://github.com/prashantv/terraform-set-null-repro/blob/main/internal/provider/service_resource.go

### Terraform Configuration Files

This configuration refers to a dummy resource that was created to repro the issue without any external dependencies.
See https://github.com/prashantv/terraform-set-null-repro for details.

To reproduce the issue, namespace is changed to some other value, and then applied. The apply results in an unexpected job set that contains 2 elements, where one is empty.

```terraform
resource "tftest_service" "s" {
  name    = "svc1"
  job {
    namespace = "NS1"
    search_tags = {
      "foo" = "bar"
    }
  }
}
...
```

### Debug Output

https://gist.github.com/prashantv/b7847ee5e35fa0167b7580cd10598faf

### Expected Behavior

Apply to send the provider a single element in the "job" set.

### Actual Behavior

Apply causes an Update that has 2 elements in the "job" set, where the first is empty, formatting the panic message:
```
panic: jobSet has more elements than expected: [
  map[namespace: search_tags:map[]]    ## unexpected empty element in set
  map[namespace:NS2 search_tags:map[foo:bar]]
]
```

### Steps to Reproduce

To reproduce this issue, a repository is created with a specific dummy provider (that has no external dependencies), and with exact repro steps,
https://github.com/prashantv/terraform-set-null-repro

The custom provider is required, as this issue requires a specific setup:
 * TypeSet field which contains a nested schema (rather than just a simple primitive like string)
 * In the nested schema, one of the fields is a TypeMap, and the TypeMap has values specified in the .tf file

### References

Originally filed on the terraform repo, but directed to the SDK,
https://github.com/hashicorp/terraform/issues/30581

### Additional Information

In case it helps anyone else, we mitigated this issue by using a `TypeList` instead of `TypeSet`, so a field change doesn't result in the whole object changing, which seems to avoid the specific bug.

disaster37:
Same matter for me

ewbankkit:
Similar: https://github.com/hashicorp/terraform-plugin-sdk/issues/588.

bflad:
[...]

Issue #588, opened 2020-09-21
TypeSet with nested TypeMap shows empty map in plan output when a map is removed from config


anGie44:
### SDK version

```
"v2.0.1" (also tested against v1)
```

### Relevant provider source code 
Excerpt from aws/resource_aws_autoscaling_group.go:

```go
"tags": {
				Type:     schema.TypeSet,
				Optional: true,
				Elem: &schema.Schema{
					Type: schema.TypeMap,
					Elem: &schema.Schema{Type: schema.TypeString},
				},
```

### Terraform Configuration Files

https://gist.github.com/anGie44/417856ad8231c96a89cf9e0d0356817f

### Debug Output

```
020/09/20 02:02:43 [WARN] Provider "registry.terraform.io/hashicorp/aws" produced an invalid plan for aws_autoscaling_group.bar, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .tags: planned value cty.SetVal([]cty.Value{cty.MapVal(map[string]cty.Value{"key":cty.StringVal("Name"), "propagate_at_launch":cty.StringVal("true"), "value":cty.StringVal("example")}), cty.MapValEmpty(cty.String)}) does not match config value cty.SetVal([]cty.Value{cty.MapVal(map[string]cty.Value{"key":cty.StringVal("Name"), "propagate_at_launch":cty.StringVal("true"), "value":cty.StringVal("example")})}) nor prior value cty.SetVal([]cty.Value{cty.MapVal(map[string]cty.Value{"key":cty.StringVal("Name"), "propagate_at_launch":cty.StringVal("true"), "value":cty.StringVal("example")}), cty.MapVal(map[string]cty.Value{"key":cty.StringVal("Owner"), "propagate_at_launch":cty.StringVal("true"), "value":cty.StringVal("bezos")})})
```
### Expected Behavior
```hcl
   ~ tags                      = [
            {
                "key"                 = "Name"
                "propagate_at_launch" = "true"
                "value"               = "example"
            },
          - {
              - "key"                 = "Owner"
              - "propagate_at_launch" = "true"
              - "value"               = "bezos"
            },
        ]
```

### Actual Behavior
```hcl
    ~ tags                      = [
            {
                "key"                 = "Name"
                "propagate_at_launch" = "true"
                "value"               = "example"
            },
          - {
              - "key"                 = "Owner"
              - "propagate_at_launch" = "true"
[...]

PR #1042, opened 2022-08-30
Fix for TypeSet applying with unexpected empty element

prashantv:
Fixes hashicorp/terraform-plugin-sdk#895

When calculating a diff, elements in sets are tracked as a delete of all
old attributes with `NewRemoved = true`, and a create of all new
attributes. When `DiffSuppressFunc` is used for an attribute, the
`ResourceAttrDiff` is replaced with `New = Old` (no-op):
https://github.com/hashicorp/terraform-plugin-sdk/blob/fa8d313665945816f6eb6c79182e4abdc1540504/helper/schema/schema.go#L1144

However, this also drops all other metadata about the attr diff, such as
`NewRemoved`. This ends up affecting the `InstanceDiff.applyBlockDiff`
which expects to drop removed items:
https://github.com/hashicorp/terraform-plugin-sdk/blob/e14d3b611f2e257d2a0862e5ea0f90ea96fd5bf8/terraform/diff.go#L207

All attributes of the removed set element get removed here, except those
with `DiffSuppressFunc` since the `NewRemoved` attribute was dropped.

Instead of dropping the metadata about the attr diff, clone the original
attr diff, and just set `New = Old`.

The added test fails without this fix:
```
=== RUN   TestSchemaMap_Diff/30-Set_with_DiffSuppressFunc
    schema_test.go:3188: expected:
        *terraform.InstanceDiff{ [...]"rule.80.duration":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:false, NewRemoved:true, [...]

        got:
        *terraform.InstanceDiff{ [...]"rule.80.duration":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:false, [...] NewRemoved:false, [...]
```

Previously, `NewRemoved` was set to false for the sustain field even though it belonged to an element being removed.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -1149,10 +1149,10 @@ func (m schemaMap) diff(
 				}
 
 				logging.HelperSchemaDebug(ctx, "Ignoring change due to DiffSuppressFunc", map[string]interface{}{logging.KeyAttributePath: attrK})
-				attrV = &terraform.ResourceAttrDiff{
-					Old: attrV.Old,
-					New: attrV.Old,
-				}
+				// If the diff is suppressed, we want Old = New, but retain the other properties.
+				updatedAttr := *attrV
+				updatedAttr.New = attrV.Old
+				attrV = &updatedAttr
 			}
 		}
 		diff.Attributes[attrK] = attrV
--- a/helper/schema/schema_test.go
+++ b/helper/schema/schema_test.go
@@ -1156,6 +1156,80 @@ func TestSchemaMap_Diff(t *testing.T) {
 		},
 
 		{
+			Name: "Set with DiffSuppressFunc",
+			Schema: map[string]*Schema{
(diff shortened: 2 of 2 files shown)


prashantv:
Friendly bump, this is an issue that has affected many folks (e.g., #588). I think this issue has found the root-cause, and it would be great to get a review from the team on whether this is the right way to address the issue.

eravin-ns1:
Friendly ping, I've also run into this issue.

prashantv:
Is there a process for getting a review from the maintainers, it'd be great to get this in as it seems to be affecting multiple users.

jfarrell-NS1:
Hi prashantv, I'll give this a review get it merged once it has been validated. Thanks for the contribution!

prashantv:
Bumping one last time, are there any active maintainers that can review this small change?

Issue #160, opened 2019-09-26
StateFunc cannot be used within a TypeSet in 0.12


jbardin:
If a provider is using `Schema.StateFunc` on an attribute of a set value, the resource will see extra entries in the set obtained from `d.Get` during `Apply`.

This is because the SDK is getting different values for the set between `PlanResourceChange` and `ApplyResourceChange`, and generating different hash values for the flatmap representation. This prevents the new protocol shims from being able to correlate the planned set values with the original config values, and they appear as separate set elements.

The recommended option here is to convert the `TypeSet` to a `TypeList` in the schema, which should behave the same in most respects, other than the inherent deduplication of identical config blocks which isn't usually a concern.

Migrating the state from the hashed value to the config value is a harder prospect, and will probably require manual intervention by the user, i.e. applying once without changing the config to save the new values.

This situation is not likely to have a remedy when using the legacy SDK with 0.12 or later, but I'm leaving the issue open here for provider developers that may encounter it during the upgrade process.

magodo:
Is this gonna be handled by hashicorp/terraform-plugin-framework#11?

Hiieu:
It seems like `StateFunc` is ignored for `TypeList` as well.

PR #1043, opened 2022-08-30
Use StateFunc when generating hashes for set elements

prashantv:
Fixes hashicorp/terraform-plugin-sdk#160 and verified [repro](https://github.com/prashantv/terraform-schemaset-repro) passes with the fix.

When a set type is used with a nested schema, and one of the attributes
has a `StateFunc`, the Plan/Apply generate separate hashes for
the set element, resulting in an unexpected element being added
to the set.

The Plan stage will use the original config value currently, so if we
have a StateFunc that converts "100s" to "1m40s", then we get:
```
Plan diff:
- remove set.hash(100s).duration
- adds NewExtra for this hash with 100s

Apply diff:
- remove set.hash(1m40s).duration
- sees NewExtra for set.Hash(100s) that is missing in the operations
  and adds it as a new field
```

This apply diff results in a new element being added to the set
unexpectedly.

Using the same hash for Plan/Apply avoids this mismatch, resulting in no
unexpected added diff.

### DIFF
--- a/helper/schema/serialize.go
+++ b/helper/schema/serialize.go
@@ -109,9 +109,15 @@ func SerializeResourceForHash(buf *bytes.Buffer, val interface{}, resource *Reso
 			continue
 		}
 
+		innerVal := m[k]
+		if innerSchema.StateFunc != nil {
+			// Normalize the value used to calculate the hash so that
+			// Plan/Apply values match.
+			innerVal = innerSchema.StateFunc(innerVal)
+		}
+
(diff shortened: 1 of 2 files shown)


prashantv:
Closing as `StateFunc` can be a lossy conversion (examples mention storing the length of a string). The hash should use the original value rather than a lossy value. I'm not sure what the best fix for this issue is, but it might be better to avoid `TypeSet` with complex objects, where some fields need a `StateFunc`.

PR #1047, opened 2022-09-02
schema: fix unknown value validation bug

kmoe:
See https://github.com/hashicorp/terraform/issues/31730

The validate function should exit early in the case that a value is not wholly known. This is the intent of the comment and earlier code in this file - however, subsequent refactors introduced a bug in which one aspect of validation was run for unknown values: deprecation warnings.

Deprecation warnings for known values are generated by `validateType` lower down.

This fix ensures that deprecation warnings are not shown for unknown values, while still ensuring we exit validation early for all unknown values.


bflad (approved):
Looks good to me, @kmoe 🚀  Thanks so much for fixing this! Double checked that indeed there is deprecation warning handling later on and covering unit testing of null vs known values for it.

bflad:
For what it's worth, folks will no longer have a feedback mechanism about setting these to unknown -> null before the attribute is fully removed, resulting in a configuration breaking change for them. I assume you folks think that tradeoff is worth it, which is why I opted to pull this in. The same change also needs to be done in terraform-plugin-framework, which is only checking an attribute value is not null directly:

https://github.com/hashicorp/terraform-plugin-framework/blob/47f5b81941b34c2260859513f62410e35e46e7c8/internal/fwserver/attribute_validation.go#L119-L137

bflad:
Framework change: https://github.com/hashicorp/terraform-plugin-framework/pull/465

jbardin:
Note that this was put in specifically to help warn about users assigning values to deprecated attribute during validation. Without this validation will pass with no warnings at all. The logic of the check was that an unknown value indicates that the attribute has been assigned at some point in the configuration, therefore the user should be warned. Even if the value ends up being null, the warning serves to communicate to the user that the attribute is being removed and will result in an error in a future version. Without this warning, users can more easily end up on a situation were upgrades cause the configuration to become invalid, and rolling back may not be easy.

Issue #717, opened 2021-03-01
Allow running Check tests on imported resources


greentruff:
### SDK version
```
v1.15.0
```
I checked 2.4.4 and the behavior for imported tests is identical, state is not modified.

### Use-cases
The terraform plugin for datadog supports modifying certain properties of default rules. Default rules cannot be created by users and are instead a special resource which must be imported before being modified. See documentation the documentation [here](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/security_monitoring_default_rule).

Implementation of the resource is [here](https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_security_monitoring_default_rule.go#L59)

We would like to implement tests which check that the fields are being modified correctly.

### Attempted Solutions
The existing tests misused the test framework and use both `ImportState: True` and `Check`. The `Check`s are ignored in this case, testing with `ImportStateCheck` the state is right after import and does not take into account other changes on top of it. It is also not possible to split the check into multiple steps as `ImportState` will revert back to the previous state after an import step.

### Proposal
Add a flag to keep state after an import step such as `ImportPersistState: True` which returns the updated state so that following steps can modify it.

### References
-

bflad:
Thanks so much for filing this, @greentruff and this is a great idea! Apologies for the long delay in responding.

I believe our preference here would be something as you mention in your proposal, a new `ImportState*` flag that can be added during the "import TestStep mode". Then provider developers would be able to do something like:

```go
resource.TestCase{
  // ... other fields omitted for brevity ...
  Steps: []resource.TestStep{
    // Importing TestStep
    {
      ImportState:       true,
      ImportStateId:     /* required if first TestStep */,
      ImportStateVerify: true,
      Config: /* Required if first TestStep */,
      ImportStatePersist: true, // <-- New flag to trigger "keeping" the imported state
    },
    // Checking TestStep
    {
      Config:            fmt.Sprintf("config that manages tfe_organization.foobar"),
      Check: /* ... */,
    },
  },
}
```

Naming suggestions are welcome, but I think aligning with the other `ImportState*` ones will suffice for now. A future iteration of the testing framework can clean up the TestStep implementations so its clearer which are meant for "apply" steps versus "import" steps.

One detail we do need to capture in provider defined implementations is that if there is no prior state, then the `Config` and `ImportStateId*` are required, which should be enforced by the testing framework.

Thanks again!

sebasslash:
[...]

PR #1052, opened 2022-09-07
Use ImportStatePersist to preserve state generated by import operation

bendbennett:
Closes: #717


bflad:
@bendbennett could you also add some initial testing:

- Inline provider with ImportState and ImportStateCheck TestStep
- External provider with ImportState and ImportStateCheck TestStep
- Inline provider with first TestStep ImportState and second TestStep Check
- External provider with first TestStep ImportState and second TestStep Check

I'm having a hard sense knowing whether this looks good or not short of manually testing it. Thank you!

bflad (approved):
Looks good to me 🚀

PR #1056, opened 2022-09-12
internal/plugintest: Switch from (os.File).Readdir() to os.ReadDir()

bflad:
Reference: https://pkg.go.dev/os#File.Readdir
Reference: https://pkg.go.dev/os#ReadDir

While attempting to troubleshoot macOS kernel panic behaviors while running acceptance testing, I was able to capture a log whose last entry was `Symlinking source directories to work directory`. Since that operation tends to occur in temporary directory space, there is at least some potential there for strange macOS and Go behaviors.

The `(os.File).Readdir()` method does make this mention in particular:

> Most clients are better served by the more efficient ReadDir method.

Making this a branch so others potentially affected by macOS kernel panics can try this out. This may not be the root cause, as there is a lot more process-oriented logic that occurs to start and stop providers in the testing framework, but if nothing else it could help.

### DIFF
--- a/internal/plugintest/util.go
+++ b/internal/plugintest/util.go
@@ -28,79 +28,40 @@ func symlinkFile(src string, dest string) error {
 	return nil
 }
 
-// symlinkDir is a simplistic function for recursively symlinking all files in a directory to a new path.
-// It is intended only for limited internal use and does not cover all edge cases.
-func symlinkDir(srcDir string, destDir string) (err error) {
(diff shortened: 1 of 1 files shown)


bflad:
While this doesn't appear to be fixing the kernel panics, I think this would still be a decent technical debt update since is recommended by the Go documentation and the code work is already done.

bendbennett (approved):
Nice upgrade to have in place.

PR #1062, opened 2022-09-19
Compile `isValidFieldName` regex once

skarimo:
SSIA


bflad (approved):
Thank you for this, @skarimo. 🚀 (I'm inclined to pull this in since its not exported, is generally a recommended code practice, and is in the "hot" path for attribute validation in the SDK.)

Issue #1069, opened 2022-09-27
Proposal: Add `RefreshOnly` acceptance test step type


gdavison:
### Use-cases

Some resource types can have state changes based on interactions with other resource types.

For example, an AWS ACM Certificate issued by an ACM Private CA is ineligible for managed renewal if it is not exported or associated with another AWS service. It would be useful to be able to validate new `Computed` values without doing an `apply`. 

### Attempted Solutions

Using a `PlanOnly` step will fail because there is a diff, and `ExpectNonEmptyPlan` simply reports that there is a diff (#482).

### Proposal

Add a `RefreshOnly` test step. This would allow using `resource.TestCheckFunc`s against the refreshed values.

PR #1070, opened 2022-09-29
Adding RefreshState test step

bendbennett:
Closes: #1069


helper/resource/testing.go
```diff
@@ -570,6 +573,19 @@ type TestStep struct {
...
+	//---------------------------------------------------------------
+
+	// RefreshState, if true, will test the functionality of `terraform
+	// refresh` by refreshing the state.
+	RefreshState bool
+
+	// RefreshStateCheck checks state following `terraform refresh`. It
+	// should be used to verify that the state has the expected resources,
+	// IDs, and attributes.
+	RefreshStateCheck RefreshStateCheckFunc
```
gdavison, line 587: For the use case I'm picturing, I'd prefer to be able to reuse the existing ecosystem of `TestCheckFunc`s after the refresh. This would allow tests to run against the whole refreshed state.

gdavison (commented):
Thanks for the quick response, @bendbennett. It looks good. One thing I'd prefer to see is that the RefreshOnly tests could run a full `TestCheckFunc`.

helper/resource/testing.go
```diff
@@ -570,6 +570,14 @@ type TestStep struct {
 	// at the end of the test step that is verifying import behavior.
 	ImportStatePersist bool
 
+	//---------------------------------------------------------------
+	// RefreshState testing
+	//---------------------------------------------------------------
+
+	// RefreshState, if true, will test the functionality of `terraform
+	// refresh` by refreshing the state.
+	RefreshState bool
```
bflad, line 579: We should consider what website documentation we can add around this. It'd also be good to denote here and in the website that its current intention is refresh, (potentially) plan, and optionally running checks.

bflad: This field should also document that it cannot be the first `TestStep` and any other potential restrictions or contradictory definitions (such as `ImportState` + `RefreshState` for example)

bendbennett: I've updated the documentation. In terms of the website docs, where do you think they should be added? The information around `ImportState` is in `/plugin/sdkv2/resources`. As we're just discussing testing here should be add a new section or add/amend an area of the website docs that already exist?

helper/resource/testing_new_refresh_state.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	}, wd, providers)
+	if err != nil {
+		t.Fatalf("Error getting state: %s", err)
+	}
+
+	if step.Config == "" {
```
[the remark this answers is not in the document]
bendbennett: Have refactored accordingly.

helper/resource/testing_new_refresh_state.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	logging.HelperResourceDebug(ctx, "Running Terraform CLI init and refresh")
+
+	err = runProviderCommand(ctx, t, func() error {
+		return wd.Init(ctx)
+	}, wd, providers)
+	if err != nil {
+		t.Fatalf("Error running init: %s", err)
+	}
```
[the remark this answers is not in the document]
bendbennett: Have refactored.

helper/resource/testing.go
bflad: I think https://www.terraform.io/plugin/sdkv2/testing/acceptance-tests/teststep is okay enough for now. Ideally that page would probably be broken up by test mode, but adding information in the current information architecture should be okay. 👍

helper/resource/testing_new_refresh_state.go
```diff
@@ -0,0 +1,99 @@
...
+		var stdout string
+		err = runProviderCommand(ctx, t, func() error {
+			var err error
+			stdout, err = wd.SavedPlanRawStdout(ctx)
+			return err
+		}, wd, providers)
+		if err != nil {
+			return fmt.Errorf("Error retrieving formatted plan output: %w", err)
+		}
+		return fmt.Errorf("After applying this test step, the plan was not empty.\nstdout:\n\n%s", stdout)
```
bflad, line 95:
```suggestion
		return fmt.Errorf("After refreshing state during this test step, a followup plan was not empty.\nstdout:\n\n%s", stdout)
```

.changelog/1070.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+helper/resource: Add RefreshState to allow testing of refresh in isolation
```
bflad, line 2:
```suggestion
helper/resource: Added `TestStep` type `RefreshState` field, which enables a step which refreshes state without an explicit apply or configuration changes
```

helper/resource/teststep_providers_test.go
```diff
@@ -1574,6 +1575,158 @@ func TestTest_TestStep_ProviderFactories_Import_External_WithoutPersistNonMatch(
...
+			{
+				PreConfig:          setTimeForTest(time.Now().Add(time.Hour * 2)),
+				RefreshState:       true,
+				ExpectNonEmptyPlan: true,
+				Check:              TestCheckResourceAttr("random_password.test", "special", "true"),
+			},
+			{
+				Config:             `resource "random_password" "test" { }`,
+				Check:              TestCheckResourceAttr("random_password.test", "special", "false"),
+				ExpectNonEmptyPlan: true,
```
bflad, line 1712: Nit: It'd be nice if the provider resource actually converged after the apply without an unexpected plan difference so it didn't look like this sort of post-RefreshState ExpectNonEmptyPlan was required, but no worries if this is difficult to work out. I think we may be able to reset the time during the `Create` function by explicitly calling something like:

```go
// Ensure test resource can converge after apply.
setTimeForTest(time.Now())
return nil
```

But I'm not sure if that's the least confusing option there.

bendbennett: Good point. Have used `PreConfig` to reset time for `TestStep` to avoid state mutation in `ReadContext` and subsequent diff generation by `CustomizeDiff`.

helper/resource/teststep_validate_test.go
```diff
@@ -79,12 +80,27 @@ func TestTestStepValidate(t *testing.T) {
...
+			expectedError: fmt.Errorf("TestStep cannot have RefreshState as first step"),
+		},
+		"importstate-and-refreshstate-both-true": {
+			testStep: TestStep{
+				ImportState:  true,
+				RefreshState: true,
+			},
+			testStepValidateRequest: testStepValidateRequest{},
+			expectedError:           fmt.Errorf("TestStep cannot have ImportState and RefreshState in same step"),
 		},
```
bflad, line 104: Just to ensure coverage of new validation:

```suggestion
		},
		"destroy-and-refreshstate-both-true": {
			testStep: TestStep{
				Destroy:      true,
				RefreshState: true,
			},
			testStepValidateRequest: testStepValidateRequest{},
			expectedError:           fmt.Errorf("TestStep cannot have RefreshState and Destroy"),
		},
```

bflad (approved):
Looks good to me 🚀

PR #1076, opened 2022-10-07
[FIX]: Errors compile regex and ignore data sources

Amad3eu:
**As of Terraform v1.3, the terraform import command populates an empty state file with the imported resource and any data sources. Previously, only the target resource was written to the state file. This causes acceptance testing steps with ImportState to fail if the configuration contains any data sources.**


bendbennett:
@Amad3eu thank you for your submission.
This looks to be a duplicate of #1074 so I'm going to close this one.

Amad3eu:
Okay, quiet, but my intention was just to join 2 PRs in one, but I ended up not referencing the previous PR

Issue #1066, opened 2022-09-22
Validation of imported resources fails in integrated test


dataclouder:
I am one of the maintainers of [vmware/terraform-provider-vcd](https://github.com/vmware/terraform-provider-vcd). We are evaluating terraform v1.3.0 with SDK v2.23.0, and we have seen that with the updated tool many tests in our suite fail with an error as described in this issue. 
With terraform up to 1.2.9, the tests run fine.

### SDK version

```
github.com/hashicorp/terraform-plugin-sdk/v2 v2.23.0
```

### Relevant provider source code

The problem arises when testing import , like in the test snippet here 
```go
			{
				ResourceName:      "vcd_catalog_item." + TestCatalogItemDS,
				ImportState:       true,           // << ---
				ImportStateVerify: true,           // <---
				ImportStateIdFunc: importStateIdOrgCatalogObject(TestCatalogItemDS),
				// These fields can't be retrieved from catalog item data
				ImportStateVerifyIgnore: []string{"ova_path", "upload_piece_size", "show_upload_progress"},
			},
```
Full test code available from [datasource_vcd_catalog_test.go](https://github.com/vmware/terraform-provider-vcd/blob/main/vcd/datasource_vcd_catalog_item_test.go)

### Terraform Configuration Files
The kind of test that fails is the one that has a few data sources followed by one resource, usually created as a data source clone, like in the snippet below.

```hcl
data "vcd_catalog" "cat-datacloud" {
  org  = "datacloud"
  name = "cat-datacloud"
}

data "vcd_catalog_item" "photon-hw11" {
  org     = "datacloud"
  catalog = data.vcd_catalog.cat-datacloud.name
  name    = "photon-hw11"
}

resource "vcd_catalog_item" "TestCatalogItemDS" {
  org     = "datacloud"
  catalog = data.vcd_catalog.cat-datacloud.name

  name                 = "TestCatalogItemDS"
  description          = data.vcd_catalog_item.photon-hw11.id
  ova_path             = "/some/path/to/terraform-provider-vcd/test-resources/test_vapp_template.ova"
  upload_piece_size    = 5
  show_upload_progress = "true"

  metadata = {
    catalogItem_metadata = "catalogItem Metadata"
    catalogItem_metadata2 = "catalogItem Metadata2"
  }
}

```

### Debug Output

Skipped: see **Root cause and fix** below.

### Expected Behavior

The imported fields should be validated.

### Actual Behavior

[...]

PR #1077, opened 2022-10-10
Adding test coverage for changes to ignore data sources in imported state

bendbennett:
References: #1074 

Closes: #1066 
Closes: #1074 

~~Changes in this PR are to accompany changes in #1074.
@gdavison feel free to pull these changes into your PR and we can merge it all in together.~~

~~Tests in this PR fail as expected, require changes in #1074 to pass.~~

I've cherry-picked @gdavison's changes from #1074 into this PR.

### DIFF
--- a/helper/resource/teststep_providers_test.go
+++ b/helper/resource/teststep_providers_test.go
@@ -1076,6 +1076,7 @@ func TestTest_TestStep_Taint(t *testing.T) {
 	}
 }
 
+//nolint:unparam
 func extractResourceAttr(resourceName string, attributeName string, attributeValue *string) TestCheckFunc {
 	return func(s *terraform.State) error {
 		rs, ok := s.RootModule().Resources[resourceName]
@@ -1253,6 +1254,8 @@ func TestTest_TestStep_ProviderFactories_To_ExternalProviders(t *testing.T) {
 }
 
 func TestTest_TestStep_ProviderFactories_Import_Inline(t *testing.T) {
+	id := "none"
+
 	t.Parallel()
 
 	Test(t, TestCase{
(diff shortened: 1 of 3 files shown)


helper/resource/teststep_providers_test.go
```diff
@@ -1738,43 +1923,41 @@ func composeImportStateCheck(fs ...ImportStateCheckFunc) ImportStateCheckFunc {
 	}
 }
 
-func testExtractResourceAttrInstanceState(attributeName string, attributeValue *string) ImportStateCheckFunc {
+func testExtractResourceAttrInstanceState(id, attributeName string, attributeValue *string) ImportStateCheckFunc {
```
bflad, line 1926: Nit: Instead of fetching via the internal identifier (typically the `id` attribute, but not necessarily guaranteed) can we instead use the resource address string? e.g. `random_string.example` Reasons being:

- This is how other testing framework "check" functions generally work
- Resources such as this one may have duplicate resource instances with the same id
- We're looking at making id less special in #1072

bflad (approved):
Overall looks good, just one minor bit about how we pinpoint specific resource instances in the testing framework. 👍

helper/resource/teststep_providers_test.go
bendbennett: Not sure how we use resource address when we're using a `type ImportStateCheckFunc func([]*terraform.InstanceState) error` as `*terraform.InstanceState` just has the `ID` and `Attributes` and doesn't appear to have any reference to the resource address.

bflad: Sigh, more trouble than its worth then (e.g. mapping those instance states with the resource addresses). We'll be removing these `terraform` bits when we take a look at the next iteration of the test framework and this isn't exported, so fine for now.

bflad:
Please note this also needs a changelog entry before merge.

Issue #1063, opened 2022-09-20
helper/resource: Potential Goroutine Leak with go-plugin


bflad:
### SDK version

```
v2.23.0
```

### Relevant provider source code

Refer to the terraform-provider-random acceptance testing.

### Debug Output

[Go timeout panic stacktrace](https://gist.github.com/bflad/39a7fa79ef8b1f4197df0f07903158b4)

### Expected Behavior

Running `go test -count=1 -parallel=1 -timeout=10m -v ./internal/provider` would yield a relatively small number of go-plugin-related Goroutines because the testing should be serialized and each test should properly close them.

### Actual Behavior

Refer to Gist above, there's a lot of this:

```
goroutine 10 [select, 9 minutes]:
github.com/hashicorp/go-plugin.(*gRPCBrokerServer).Recv(0x0?)
	/Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_broker.go:121 +0x58
github.com/hashicorp/go-plugin.(*GRPCBroker).Run(0x14000046460)
	/Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_broker.go:411 +0x40
created by github.com/hashicorp/go-plugin.(*GRPCServer).Init
	/Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_server.go:85 +0x424

goroutine 75981 [select]:
github.com/hashicorp/go-plugin.(*gRPCBrokerServer).Recv(0x1005ab701?)
	/Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_broker.go:121 +0x58
github.com/hashicorp/go-plugin.(*GRPCBroker).Run(0x140002b2370)
	/Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_broker.go:411 +0x40
created by github.com/hashicorp/go-plugin.(*GRPCServer).Init
	/Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_server.go:85 +0x424

goroutine 5016 [chan receive, 8 minutes]:
testing.(*testContext).waitParallel(0x14000127e50)
	/opt/homebrew/Cellar/go/1.19.1/libexec/src/testing/testing.go:1557 +0x15c
testing.(*T).Parallel(0x140001d1a00)
[...]

PR #1091, opened 2022-11-04
helper/resource: Various small fixes

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-sdk/issues/1063
Reference: https://github.com/hashicorp/go-plugin/pull/220

This includes the following small bug fixes:

- Clarifies/fixes the trace logging during `terraform show` commands
- Improves performance slightly for `terraform show` raw plan usage (terraform-exec already captures and returns stdout for raw plans)
- Prevents a goroutine leak specific to the detached stop context in `schema.GRPCProvider`

Previously, leak detection on testing would contain an entry per Terraform command such as:

```
        [Goroutine 276 in state select, with github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.mergeStop on top of the stack:
        goroutine 276 [select]:
        github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.mergeStop({0x103a5f298?, 0x14000390c00?}, 0x1400033e5f0, 0x140004bb560)
                /Users/bflad/src/github.com/hashicorp/terraform-plugin-sdk/helper/schema/grpc_provider.go:46 +0x64
        created by github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).StopContext
                /Users/bflad/src/github.com/hashicorp/terraform-plugin-sdk/helper/schema/grpc_provider.go:63 +0x178
        ]
```

Leaked goroutines such as:

```
        [Goroutine 53 in state select, with github.com/hashicorp/go-plugin.(*gRPCBrokerServer).Recv on top of the stack:
        goroutine 53 [select]:
        github.com/hashicorp/go-plugin.(*gRPCBrokerServer).Recv(0x29?)
                /Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_broker.go:121 +0x58
        github.com/hashicorp/go-plugin.(*GRPCBroker).Run(0x140003ca190)
                /Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_broker.go:411 +0x40
        created by github.com/hashicorp/go-plugin.(*GRPCServer).Init
                /Users/bflad/go/pkg/mod/github.com/hashicorp/go-plugin@v1.4.4/grpc_server.go:85 +0x424
        ]
```

Will likely require a fix upstream in go-plugin.


SBGoods (approved):

bendbennett (approved):
LGTM 👍

helper/resource/plugin.go
```diff
@@ -430,6 +443,12 @@ func runProviderCommand(ctx context.Context, t testing.T, f func() error, wd *pl
 	// get closed, and we'll hang here.
 	cancel()
 
+	// For legacy providers, call the StopProvider RPC so the StopContext
+	// goroutine is cleaned up properly.
+	for _, legacyProviderServer := range legacyProviderServers {
+		legacyProviderServer.StopProvider(ctx, nil) //nolint:errcheck // does not return errors
```
bflad, line 449: For what it is worth, we could consider calling this RPC on all providers started by the testing framework as a better approximation to normal Terraform executions, but it only makes sense for sdk/framework-based providers if we enable them to implement close/stop logic that gets called during that RPC.

PR #1106, opened 2022-11-18
deps: Remove golang.org/x/tools dependency

bflad:
This dependency was used for calling `cmd/stringer` in `go:generate` directives, mainly in inherited code from the Terraform core codebase which used that practice. The code associated with these generated `String` methods is now generally considered frozen or not difficult to manually update, so this removes the tooling dependency to reduce maintenance/security burden for the Go module.


bendbennett (approved):
LGTM

PR #1108, opened 2022-11-21
deps: Update golang.org/x/crypto@v0.3.0

bflad:
Updates `golang.org/x/crypto` to a tagged release so dependency management tooling such as Dependabot begin suggesting version updates. The latest tagged release, `v0.3.0`, shows no immediate consequential changes over the first tagged release last month.

The only direct `golang.org/x/crypto` usage in this Go module is the `helper/acctest.RandSSHKeyPair()` function. Added unit testing to verify its current implementation assumptions. Updated its Go documentation to clarify that developers should opt to use the `crypto` and `golang.org/x/crypto/ssh` packages directly should there be different requirements for the private key.

Dependency updated via:

```shell
go get golang.org/x/crypto@v0.3.0
go mod tidy
```


bendbennett (approved):
LGTM

PR #1119, opened 2023-01-05
chore: rename nav-data file to align with the docs paths

BRKalow:
Some of out internal tooling assumes that the nav data file name matches the path to the content, this renames the nav data file to align with that.

Content lives at: `docs/plugin/sdkv2`, so the nav data filename should be `plugin-sdkv2-nav-data.json`.


bflad (approved):
Looks good to me 🚀

PR #1146, opened 2023-02-13
helper/schema: Remove timeouts during ImportResourceState

bflad:
Closes #1145
Reference: https://github.com/hashicorp/terraform/pull/32463

Terraform 1.3.8 and later now correctly handles null values for single nested blocks. This means Terraform will now report differences between a null block and known block with null values. This SDK only supported single nested blocks via its timeouts functionality.

This change is a very targeted removal of any potential `timeouts` block values in a resource state from the `ImportResourceState` RPC. Since configuration is not available during that RPC, it is never valid to return any data beyond null for that block. This will prevent unexpected differences on the first plan after import, where Terraform will report the block removal for configurations which do not contain the block.

New unit test failure prior to code updates:

```
--- FAIL: TestImportResourceState_Timeouts_Removed (0.00s)
    /Users/bflad/src/github.com/hashicorp/terraform-plugin-sdk/helper/schema/grpc_provider_test.go:1159: unexpected difference:   cty.Value(
        - 	{
        - 		ty: cty.Type{typeImpl: cty.typeObject{AttrTypes: map[string]cty.Type{...}}},
        - 		v:  map[string]any{"id": string("test"), "string_attribute": nil, "timeouts": nil},
        - 	},
        + 	{
        + 		ty: cty.Type{typeImpl: cty.typeObject{AttrTypes: map[string]cty.Type{...}}},
        + 		v: map[string]any{
        + 			"id":               string("test"),
        + 			"string_attribute": nil,
        + 			"timeouts":         map[string]any{"create": nil, "read": nil},
        + 		},
        + 	},
          )
```

PR #1147, opened 2023-02-13
helper/resource: Remove special timeouts logic in TestStep ImportStateVerify

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-sdk/pull/576
Reference: https://github.com/hashicorp/terraform-plugin-sdk/pull/1146

This change reverts #576 as it prevents acceptance testing from detecting Terraform 1.3.8 and later potentially returning an unexpected plan after import involving the `timeouts` block. #1146 fixes the import logic to never include a known value for `timeouts`, which is what the acceptance testing was trying to say when this testing logic was not in place.

With this change, but without #1146 on a resource known to have the unexpected plan after import on Terraform 1.3.8:

```
=== CONT  TestAccVPCSecurityGroup_basic
    vpc_security_group_test.go:1002: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.

          map[string]string{
        +       "timeouts.%": "2",
          }
--- FAIL: TestAccVPCSecurityGroup_basic (21.54s)
```

With this change and with #1146:

```
=== CONT  TestAccVPCSecurityGroup_basic
--- PASS: TestAccVPCSecurityGroup_basic (20.23s)
```

Verified both ways on Terraform 1.2.9 as well (as a smoke test for a Terraform version prior to 1.3.8).

### DIFF
--- a/.changes/unreleased/BUG FIXES-20230213-155557.yaml
+++ b/.changes/unreleased/BUG FIXES-20230213-155557.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'helper/resource: Ensured ImportStateVerify can detect timeouts differences
+  that may result in an unclean plan after import'
+time: 2023-02-13T15:55:57.259366-05:00
+custom:
+  Issue: "1147"
--- a/helper/resource/testing_new_import_state.go
+++ b/helper/resource/testing_new_import_state.go
@@ -236,26 +236,6 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
 				}
 			}
 
-			// timeouts are only _sometimes_ added to state. To
-			// account for this, just don't compare timeouts at
-			// all.
-			for k := range actual {
-				if strings.HasPrefix(k, "timeouts.") {
-					delete(actual, k)
-				}
-				if k == "timeouts" {
-					delete(actual, k)
-				}
-			}
-			for k := range expected {
-				if strings.HasPrefix(k, "timeouts.") {
-					delete(expected, k)
-				}
-				if k == "timeouts" {
-					delete(expected, k)
-				}
-			}
-
 			if !reflect.DeepEqual(actual, expected) {
 				// Determine only the different attributes
 				// go-cmp tries to show surrounding identical map key/value for


bflad:
I tried writing unit testing to cover this change and I don't think we can make this change.

```go
// Verify that terraform-plugin-sdk automatic timeouts handling does not cause
// unexpected ImportStateVerify differences. This test specifically covers
// when the configuration has timeouts on creation (instead of timeouts being
// added via a later update), since it may have differing state handling.
func TestTest_TestStep_ImportStateVerify_Timeouts_CreateConfig(t *testing.T) {
	t.Parallel()

	UnitTest(t, TestCase{
			"examplecloud": func() (*schema.Provider, error) { //nolint:unparam // required signature
				return &schema.Provider{
					ResourcesMap: map[string]*schema.Resource{
						"examplecloud_thing": {
							CreateContext: func(_ context.Context, d *schema.ResourceData, _ interface{}) diag.Diagnostics {
								d.SetId("resource-test")

								return nil
							},
							DeleteContext: func(_ context.Context, _ *schema.ResourceData, _ interface{}) diag.Diagnostics {
								return nil
							},
							ReadContext: func(_ context.Context, d *schema.ResourceData, _ interface{}) diag.Diagnostics {
								return nil
							},
								"id": {
									Computed: true,
									Type:     schema.TypeString,
								},
							},
							Importer: &schema.ResourceImporter{
								StateContext: schema.ImportStatePassthroughContext,
							},
								Create: schema.DefaultTimeout(10 * time.Minute),
								Update: schema.DefaultTimeout(10 * time.Minute),
							},
						},
					},
				}, nil
			},
		},
		Steps: []TestStep{
			{
				Config: `resource "examplecloud_thing" "test" {
					timeouts {
						create = "20m"
					}
				}`,
			},
			{
				ResourceName:      "examplecloud_thing.test",
				ImportState:       true,
				ImportStateVerify: true,
			},
		},
	})
}
```

Results in:

```
--- FAIL: TestTest_TestStep_ImportStateVerify_Timeouts_CreateConfig (0.86s)
    /Users/bflad/src/github.com/hashicorp/terraform-plugin-sdk/helper/resource/testing_new_import_state_test.go:145: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        - 	"timeouts.%":      "2",
        - 	"timeouts.create": "20m",
          }
```

We don't want to break existing testing for this.

PR #1159, opened 2023-02-28
Prevent destroy resource

vmanilo:
Resolves https://github.com/hashicorp/terraform-plugin-testing/issues/85

Added `PreventPostDestroy` flag, which can be set to `true` when we testing against real resource and want to keep it.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -388,6 +388,10 @@ type TestCase struct {
 	// are tested alongside real resources
 	PreventPostDestroyRefresh bool
 
+	// PreventPostDestroy can be set to true for cases where resources
+	// are tested alongside real resources
+	PreventPostDestroy bool
+
 	// CheckDestroy is called after the resource is finally destroyed
 	// to allow the tester to test that the resource is truly gone.
 	CheckDestroy TestCheckFunc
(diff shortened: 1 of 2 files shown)


bflad:
Hi @vmanilo 👋 Thank you for submitting this change.

I think we want to shy away from introducing an overly broad flag like this into the testing framework. The prior overly broad `PreventPostDestroyRefresh` flag is not a design pattern we would like to use going forward since it has limited utility for most developers.

The intended use for acceptance testing managed resources is for the full lifecycle of an entire Terraform configuration, which may contain multiple resources. A followup enhancement would likely be the ability to select which resource(s) should be ignored during destroy, therefore adding yet more potentially confusing/conflicting options into `TestStep`. There's some proposals outlined for that implementing selective state removal support in https://github.com/hashicorp/terraform-plugin-testing/issues/85#issuecomment-1448927282, but the maintainers have not met to discuss the desired path forward yet.

I'm going to close this pull request for now because of the above and because we are targeting future enhancements over in the [newly introduced terraform-plugin-testing Go module](https://discuss.hashicorp.com/t/announcing-the-terraform-plugin-testing-module/49914). Thank you again for submitting this though. I would encourage adding any further discussion on this over in https://github.com/hashicorp/terraform-plugin-testing/issues/85 -- it would be especially helpful to understand the use cases why developers may desire this functionality to ensure we are designing a solution that covers a wide selection of those use cases.

Issue #1166, opened 2023-03-09
Migration to terraform-plugin-testing causes panic in existing test suite


bendbennett:
### SDK version

```
v2.25.0
```

### Relevant provider source code

```go
// n/a
```

### Terraform Configuration Files

```hcl
// n/a
```

### Expected Behavior

Following [migration](https://developer.hashicorp.com/terraform/plugin/testing/migrating) to using `terraform-plugin-testing` all passing tests should continue to pass.

### Actual Behavior

```
panic: /var/folders/23/p8tx31c53rd16t1pjt03zndc0000gq/T/go-build764025451/b443/provider_sdk.test flag redefined: sweep

goroutine 1 [running]:
flag.(*FlagSet).Var(0x14000030240, {0x10132b6d8, 0x140002219c0}, {0x10103ad88, 0x5}, {0x10105d47b, 0x29})
        /usr/local/go/src/flag/flag.go:982 +0x2a4
flag.(*FlagSet).StringVar(...)
        /usr/local/go/src/flag/flag.go:847
flag.(*FlagSet).String(0x1400031fb98?, {0x10103ad88, 0x5}, {0x0, 0x0}, {0x10105d47b, 0x29})
        /usr/local/go/src/flag/flag.go:860 +0x98
flag.String(...)
        /usr/local/go/src/flag/flag.go:867
github.com/hashicorp/terraform-plugin-testing/helper/resource.init()
        /Users/bdb/go/pkg/mod/github.com/hashicorp/terraform-plugin-testing@v1.1.0/helper/resource/testing.go:53 +0x44
FAIL    github.com/bendbennett/terraform-provider-playground/internal/provider_sdk      0.437s
FAIL
```

### Steps to Reproduce

1. Follow [migration guide](https://developer.hashicorp.com/terraform/plugin/testing/migrating) in any Terraform Provider that uses [github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource](https://developer.hashicorp.com/terraform/plugin/testing/migrating#github-com-hashicorp-terraform-plugin-sdk-v2-helper-resource) in any "non-test" files (i.e., any files that are not `_test.go`).
2. Run tests

### References

- https://github.com/hashicorp/terraform-plugin-testing/issues/93
- https://github.com/hashicorp/terraform-plugin-framework/issues/513

jacobbednarz:
thanks for this one @bendbennett ✌️ appreciate you getting to the bottom of it.

[...]

PR #1167, opened 2023-03-09
Fix migration to terraform-plugin-testing causes panic in existing test suite

bendbennett:
Closes: #1166

### DIFF
--- a/helper/resource/aliases.go
+++ b/helper/resource/aliases.go
@@ -1,0 +1,109 @@
+package resource
+
+import (
+	"context"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+)
+
+// Deprecated: Use helper/retry package instead. This is required for migrating acceptance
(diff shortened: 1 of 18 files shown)


helper/resource/aliases.go
```diff
@@ -0,0 +1,109 @@
+package resource
+
+import (
+	"context"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
```
austinvalle, line 7: Do we need to add the exports from the new `id` package to this set of aliases as well?

image

bflad (approved):
Overall looks good, just some things to consider. 👍

helper/resource/aliases.go
bflad: +1

bendbennett: ++1

bflad (approved):

austinvalle (approved):
🚀

PR #1181, opened 2023-03-29
[wip] Prevent useless strings creations during applyCollectionDiff

freedge:
before:

image

following this change my terraform plan goes from  120s to 100s.

this is more a fyi as this pr is made against v1.17.2 (as used on my provider) but the main branch seems to use the same logic.

### DIFF
--- a/terraform/diff.go
+++ b/terraform/diff.go
@@ -747,11 +747,14 @@ func (d *InstanceDiff) applyCollectionDiff(path []string, attrs map[string]strin
 	}
 
 	currentKey := prefix + name
+	currentKeyDot := currentKey + "."
+	currentKeyDotPct := currentKeyDot + "%"
+	currentKeyDotSharp := currentKeyDot + "#"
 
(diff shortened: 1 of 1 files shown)


bflad:
Hi @freedge 👋 Thank you for submitting this. Please note that [terraform-plugin-sdk v1 has been end-of-life since July 31, 2021](https://discuss.hashicorp.com/t/end-of-life-timeline-for-v1-of-the-terraform-plugin-sdk/24991), therefore we are not accepting any changes to that code since it will never be released. Peeking at the changes, they appear fairly trivial in nature so we would consider something like this for v2, if you're interested. There is a [terraform-plugin-sdk v2 upgrade guide available](https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide), although depending on the provider size, it may be worth skipping over to [terraform-plugin-framework instead](https://developer.hashicorp.com/terraform/plugin/framework-benefits) at this point.

PR #1188, opened 2023-04-10
helper/schema: Remove Default mention from Schema type Optional field Go documentation

### LINKED ISSUE #1185: Inconsistent Optional w/ Default behavior
### SDK version
v2.24.1

```
...
```

```go
schema.Schema{
  Type: schema.TypeString,
  Default:      "Some default",
  Optional:     true,
  Computed:     true,
}
```

### Debug Output
```
resource my_resource: some_attribute: Default must be nil if computed
```

### Expected Behavior
Per the [SDK documentation](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema):

```go
        // Optional indicates whether the practitioner can choose to not enter
	// a value in the configuration for this attribute. Optional cannot be used
	// with Required.
	//
	// If also using Default or DefaultFunc, Computed should also be enabled,
	// otherwise Terraform can output warning logs or "inconsistent result
	// after apply" errors.
	Optional [bool](https://pkg.go.dev/builtin#bool)
```

### Actual Behavior
This recommendation of supplying "Computed" when both "optional" and with a "default" seems to produce an error suggesting that "Computed" attributes must never have a default.

bflad:
Closes #1185

While the intention of the Go documentation matches Terraform's expectations, other parts of the SDK explicitly validate against the schema definition it was recommending. This documentation change is safer than trying to reconcile all the potential behavior changes should this particular validation be removed and a developer attempts to set `Computed`.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -88,10 +88,6 @@ type Schema struct {
 	// Optional indicates whether the practitioner can choose to not enter
 	// a value in the configuration for this attribute. Optional cannot be used
 	// with Required.
-	//
-	// If also using Default or DefaultFunc, Computed should also be enabled,
-	// otherwise Terraform can output warning logs or "inconsistent result
-	// after apply" errors.
 	Optional bool
 
 	// Computed indicates whether the provider may return its own value for


austinvalle (approved):
lgtm 👍🏻

PR #1194, opened 2023-05-12
helper/validation: clarify IsCIDR() error message

ndrpnt:
The old error was misleading for users, as IPv6 CIDR are valid


bflad (approved):
Looks good to me, thank you for this, @ndrpnt 🚀

PR #1224, opened 2023-07-10
helper/logging: Adjust unit test regex flag

austinvalle:
Originally found in #1223 - The unit test `TestNewLoggingHTTPTransport_LogMasking` started failing in a recent dependabot PR: https://github.com/hashicorp/terraform-plugin-sdk/actions/runs/5473296958/jobs/10006381743?pr=1223

The test failing is using a `GET https://www.terraform.io` request to test the masking for the logger on the response body, using a regex `<html>.*</html>` for the mask. I believe the site must have changed recently and introduced some `\n` characters, because `.` in RE2 regex does not match `\n` characters by default: https://github.com/google/re2/blob/a57a1d6462a1613e5e2cfd6fb1ce26d36706a9af/doc/syntax.txt#L68

Adjusting the test's regex to include that flag should fix it -> `(?s)<html>.*</html>`

## Notes
- Added timeout for lint

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,3 +25,7 @@ linters:
     - unparam
     - unused
     - vet
+
+run:
+  # Prevent false positive timeouts in CI
+  timeout: 5m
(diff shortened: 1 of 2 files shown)


austinvalle:
> Seems okay for now. **Ideally, we'd spin up our own web server to control the unit test better**, but that seems like a project for when this is handled in terraform-plugin-log, rather than something to adjust here at this point.

Agreed 😆

PR #1229, opened 2023-08-08
helper/schema: Add `Resource` type flags for disabling legacy type system protocol flags

### LINKED ISSUE #1227: Consider Method for Disabling Legacy SDK Data Consistency Warning Log Demotion
### Terraform Version

```shell
Terraform v1.5.4
on darwin_arm64
```

### Use Cases

Terraform provider resources written with the legacy [terraform-plugin-sdk Go module](https://github.com/hashicorp/terraform-plugin-sdk) have a special flag enabled in the protocol which demotes certain cases of data consistency errors into warning logs:

```
TIMESTAMP [WARN]  Provider "TYPE" produced an invalid plan for ADDRESS, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .ATTRIBUTE: planned value cty.False for a non-computed attribute
```

If the problematic attribute value that raises the warning log is referenced elsewhere in a configuration, Terraform will raise an error diagnostic in the receiving location, which can help with finding these cases, but has a relatively low hit rate as compared to every attribute value that has this issue. This was a pragmatic choice for the existing provider ecosystem since at the time there was not a updated provider development SDK to match Terraform 0.12 and later concepts. It would have been extremely difficult for practitioners to upgrade Terraform if a part of doing so would have been to force all providers to fix all instances of these data inconsistency errors.

Fast forward to the present, the newer [terraform-plugin-framework Go module](https://github.com/hashicorp/terraform-plugin-framework) is available, but it intentionally does not enable the protocol flag since a goal for moving the provider ecosystem forward is to not support errant provider implementations with respect to Terraform's intended data handling rules. Provider developers are embarking on the journey of [migrating their terraform-plugin-sdk based providers to terraform-plugin-framework](https://developer.hashicorp.com/terraform/plugin/framework/migrating), and potentially encountering these data consistency errors for the first time either during initial acceptance testing after migrating their code or in worst case scenarios, after its been released after the migration.

At scale with larger providers, such as a major cloud provider, it is difficult to know about these underlying issues since they only exist in logging. Most practitioners/environments are not running with Terraform's logging capabilities enabled (e.g. setting the `TF_LOG` environment variable), and even for environments where this is enabled, it is difficult for developers to discover and remediate these issues since there are no native provider acceptance testing capabilities to detect this situation short of writing acceptance tests which explicitly check every attribute value in every resource.

### Attempted Solutions

Rely on practitioner bug reports of the downstream resource error (when triggered) or provider developers enabling logging in acceptance testing environments and searching those logs for the relevant log entries.

### Proposal

Introduce an environment variable so provider developers can have these errors raised during provider acceptance testing. In the core logic that checks for the legacy type system flag, conditionalize the warning log demotion on also not having the environment variable set. Normal practitioners should never encounter the newer behavior. Provider developers have the option of setting this environment variable in their provider acceptance testing environments. The terraform-plugin-framework migration guide can explicitly mention this environment variable as a preparation step.

### References

_No response_

bflad:
Closes #1227

Introduces new `Resource` type `EnableLegacyTypeSystemApplyErrors` and `EnableLegacyTypeSystemPlanErrors` fields that prevent the `PlanResourceChange` and `ApplyResourceChange` protocol operation responses from setting the `UnsafeLegacyTypeSystem` flag. When Terraform encounters the `UnsafeLegacyTypeSystem` flag, it will demote certain data consistency issues from error diagnostics that would immediately prevent Terraform workflows from completing to warning logs, which can be difficult for provider developers to discover. Provider developers are not  generally expected to try enabling these settings unless discovering these data consistency issues upfront is desired before migrating resources to terraform-plugin-framework.

This setting is on the individual resource level because these errors can be quite prevalent in existing terraform-plugin-sdk resources and therefore it would be problematic to not give provider developers more control over the behavior.

The settings are split between plan and apply because the protocol flag is on two separate protocol operations and because certain errors for one operation may be unavoidable, so it leaves provider developers the option to still raise errors for the other operation.

This change set includes new website documentation to discuss these data consistency issues and how to potentially resolve them more in-depth. It is expected that the framework migration guide will be updated to recommend enabling these settings before migrating to help provider developers understand the issues before migrating. The error resolution section has one particular error to start and it is expected that this section will grow over time as provider developers report the various data consistency errors that Terraform can raise.

### DIFF
--- a/helper/schema/grpc_provider_test.go
+++ b/helper/schema/grpc_provider_test.go
@@ -575,77 +575,113 @@ func TestUpgradeState_flatmapStateMissingMigrateState(t *testing.T) {
 }
 
 func TestPlanResourceChange(t *testing.T) {
-	r := &Resource{
-		SchemaVersion: 4,
-		Schema: map[string]*Schema{
-			"foo": {
-				Type:     TypeInt,
-				Optional: true,
+	t.Parallel()
+
+	testCases := map[string]struct {
+		TestResource                   *Resource
+		ExpectedUnsafeLegacyTypeSystem bool
+	}{
+		"basic": {
+			TestResource: &Resource{
+				SchemaVersion: 4,
+				Schema: map[string]*Schema{
+					"foo": {
+						Type:     TypeInt,
+						Optional: true,
+					},
+				},
 			},
+			ExpectedUnsafeLegacyTypeSystem: true,
+		},
+		"EnableLegacyTypeSystemPlanErrors": {
+			TestResource: &Resource{
+				EnableLegacyTypeSystemPlanErrors: true,
+				Schema: map[string]*Schema{
+					"foo": {
+						Type:     TypeInt,
+						Optional: true,
+					},
+				},
+			},
+			ExpectedUnsafeLegacyTypeSystem: false,
 		},
 	}
 
-	server := NewGRPCProviderServer(&Provider{
-		ResourcesMap: map[string]*Resource{
-			"test": r,
-		},
-	})
+	for name, testCase := range testCases {
+		name, testCase := name, testCase
 
-	schema := r.CoreConfigSchema()
-	priorState, err := msgpack.Marshal(cty.NullVal(schema.ImpliedType()), schema.ImpliedType())
-	if err != nil {
-		t.Fatal(err)
-	}
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
 
-	// A propsed state with only the ID unknown will produce a nil diff, and
-	// should return the propsed state value.
-	proposedVal, err := schema.CoerceValue(cty.ObjectVal(map[string]cty.Value{
-		"id": cty.UnknownVal(cty.String),
-	}))
-	if err != nil {
-		t.Fatal(err)
-	}
-	proposedState, err := msgpack.Marshal(proposedVal, schema.ImpliedType())
-	if err != nil {
-		t.Fatal(err)
-	}
+			server := NewGRPCProviderServer(&Provider{
+				ResourcesMap: map[string]*Resource{
+					"test": testCase.TestResource,
+				},
+			})
 
-	config, err := schema.CoerceValue(cty.ObjectVal(map[string]cty.Value{
-		"id": cty.NullVal(cty.String),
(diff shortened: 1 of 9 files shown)


helper/schema/grpc_provider_test.go
bflad, line 1: Switching the pull request view to ignore whitespace changes should make this file's changes easier to review (`TestPlanResourceChange` was switched to table-driven testing). 👍

bflad:
No idea why CircleCI is suddenly trying to run against this repository again, when it hasn't done so in years. I will reach out to the team that manages that integration.

website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
```diff
@@ -0,0 +1,112 @@
...
+            },
+            // ... potentially other steps as necessary ...
+        },
+    })
+}
+```
+
+The state value check will raise an error if there is an unexpected value difference, which is the same as if Terraform raised the error. Checking state values of configured values can be removed once the resource is [migrated to terraform-plugin-framework](/terraform/plugin/framework/migrating), as Terraform itself will raise any potential error.
+
+## Resolving Data Consistency Errors
```
austinvalle, line 94: Not sure we need to open the can of worms in this PR, but should we eventually add the other common error `planned value does not match config value or prior value` to this section?

austinvalle (approved):
Overall lgtm, love the docs ❤️

jbardin (approved):

website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
bflad: Yeah, that would eventually be my goal from the PR description. I just don't have all those various errors handy right now and we'll need to figure out how we best want to discuss each one.

austinvalle: Sorry! I missed that part of your PR description 😵‍💫

bendbennett (approved):
LGTM 🚀

PR #1232, opened 2023-08-24
website: Add data consistency error docs for "planned value does not match config value"

austinvalle:
Ref: #1229 

This PR adds additional documentation about a common data consistency error we see during migration to Plugin Framework:
```bash
TIMESTAMP [WARN]  Provider "TYPE" produced an invalid plan for ADDRESS, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .ATTRIBUTE: planned value cty.StringVal("VALUE") does not match config value cty.StringVal("value")
```

I added a simplified example for demonstrating the behavior, but can adjust/add more examples if we think that'd be helpful.

### DIFF
--- a/website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
+++ b/website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
@@ -110,3 +110,140 @@ This occurs if the attribute schema definition is `Optional: true` without `Comp
 If the value is expected to never be set by configuration, the schema attribute `Optional: true` flag should be replaced with `Computed: true`.
 
 Otherwise, this may not be resolvable when the resource is implemented with `terraform-plugin-sdk`. Having `Optional: true` while also setting the attribute's `Computed: true` flag in the schema will also enable this SDK's behavior of keeping the prior state value if the configuration value is removed (set to null) during an update. That SDK behavior is unavoidable. This SDK will also raise an implementation error if both `Computed: true` and `Default` are set, since the value will never reset to the default value because of that behavior. If that behavior is not acceptable, this error is unavoidable until the resource is migrated to terraform-plugin-framework, which does not have implicit behaviors when enabling the `Computed: true` flag and instead provider developers are expected to decide whether the prior state preservation behavior should occur or not by using the `UseStateForUnknown` schema plan modifier.
+
+### Planned Value does not match Config Value
+
+If the resource is raising this type of error or warning log:
+
+```text
+TIMESTAMP [WARN]  Provider "TYPE" produced an invalid plan for ADDRESS, but we are tolerating it because it is using the legacy plugin SDK.
+    The following problems may be the cause of any confusing errors from downstream operations:
+      - .ATTRIBUTE: planned value cty.StringVal("VALUE") does not match config value cty.StringVal("value")
+```
+
+This occurs for attribute schema definitions that are `Optional: true` and `Computed: true`; where the planned value, returned by the provider, does not match the attribute's config value or prior state value. For example, value's for an attribute of type string must match byte-for-byte.
+
+An example root cause of this issue could be from API normalization, such as a JSON string being returned from an API and stored in state with differing whitespace then what was originally in config.
+
+#### SDKv2 Example
+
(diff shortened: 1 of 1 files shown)


website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
```diff
@@ -110,3 +110,140 @@ This occurs if the attribute schema definition is `Optional: true` without `Comp
...
+#### Recommended Solution
+To solve this issue, the provider code must preserve the config value or prior state value when producing the new state. The recommended way to implement this logic is by creating a [custom type](/terraform/plugin/framework/handling-data/types/custom) with [semantic equality logic](/terraform/plugin/framework/handling-data/types/custom#semantic-equality). A custom type can be shared across multiple resource attributes and will ensure that the semantic equality logic is invoked during the `Read`, `Create`, and `Update` methods respectively.
+
+For the above example, the semantic equality implementation below would resolve the resource drift and error:
+```go
+type CaseInsensitive struct {
+	basetypes.StringValue
+}
+
+// ... custom value type implementation
```
bendbennett, line 226: Not sure if it's worth adding further detail about what would need to be implemented for a custom value type, or perhaps linking to the relevant docs?

bendbennett (approved):
LGTM 🚀

bflad (approved):
I don't want to hold this excellent content up from going somewhere, but something to potentially consider here. 👍

website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
austinvalle: Good suggestion. I'm going to move this section over to the migration guide in the framework repo, but I'll add a `<Note>` above this code example that says this isn't a full example and links to the custom value documentation 👍🏻

austinvalle: Added suggestion in new PR, see https://github.com/hashicorp/terraform-plugin-framework/pull/831/files#diff-671bcafb4ced355e43bc6d152c9bef463a5fc27fa42d2ec2eaa4b9409004d33aR245-R249

website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
```diff
@@ -110,3 +110,140 @@ This occurs if the attribute schema definition is `Optional: true` without `Comp
...
+            "word": "VALUE"
+          },
+        }
+      ]
+    }
+  ],
+}
+```
+
+#### Migrating to Plugin Framework
```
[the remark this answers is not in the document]
austinvalle: I opened https://github.com/hashicorp/terraform-plugin-framework/pull/831 and moved the code over in 016f347

austinvalle:
@bflad @bendbennett - I moved the code over to a [new PR in plugin-framework](https://github.com/hashicorp/terraform-plugin-framework/pull/831). I removed the migration section + plugin framework code from this PR and added an SDKv2 example.

I'd love a second review on the SDKv2 example I provided, as I was basing a lot of it off the Semantic Equality RFC and my sandbox testing. Want to make sure this example makes sense 😄

website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
```diff
@@ -110,3 +110,100 @@ This occurs if the attribute schema definition is `Optional: true` without `Comp
...
+          "attributes": {
+            // This is invalid! Config has this stored as "value"
+            "word": "VALUE"
+          },
+        }
+      ]
+    }
+  ],
+}
+```
```
bflad, line 178: I'm not sure folks will be used to the seeing the JSON state, so maybe best to just say what will happen:

```suggestion
A warning log will be produced and the resulting state after applying a new resource it will store the attribute state as `VALUE` instead of `value`.
```

bflad (approved):
This looks good to me 🚀

website/docs/plugin/sdkv2/resources/data-consistency-errors.mdx
austinvalle: Good point -> ⚡

PR #1237, opened 2023-08-29
helper/validation: `StringInSlice`: Single quote slice elements

### LINKED ISSUE #464: Improve validation.StringInSlice error message
### SDK version

```
{
  "Path": "github.com/hashicorp/terraform-plugin-sdk",
  "Version": "v1.13.0"
}
```

### Use-cases

When using the `StringInSlice` SchemaValidateFunc with valid strings that contain white spaces.
For example:
```
ValidateFunc: validation.StringInSlice([]string{"Base Policy", "Allow", "Deny"}, false),
```

If the user inputs an invalid value the error message looks like 
```
Error: expected domain_names.0.action to be one of [Base Policy Allow Deny], got dont_allow
```
It is ambiguous whether "Base Policy" is a single phrase or two separate  words.
### Attempted Solutions

I've written my own version of `StringInSlice` that separates words by comma. So the error message instead looks like
```
 Error: expected domain_names.0.action to be one of [Base Policy, Allow, Deny], got dont_allow
```
### Proposal

By default, `StringInSlice` should separate valid values by comma in the error message. I would be happy to work on this as I've already done it once! 😄 
### References

None.

Frankkkkk:
The validation function's error is ambiguous when dealing with slices containing spaces. For example:
```
expected role_name to be one of [Apache Spark Administrator Synapse Credential User Synapse Administrator]
```

By changing the format verb, the error becomes more user friendly:
```
expected role_name to be one of ["Apache Spark Administrator" "Synapse Credential User" "Synapse Administrator"]
```

This fixes #464

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20230829-141309.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20230829-141309.yaml
@@ -1,0 +1,6 @@
+kind: ENHANCEMENTS
+body: 'helper/validation: Added quoting in `StringInSlice` error diagnostic output to
+  prevent confusion with values that contain spaces'
+time: 2023-08-29T14:13:09.735543+02:00
+custom:
+  Issue: "464"
--- a/helper/validation/strings.go
+++ b/helper/validation/strings.go
@@ -146,7 +146,7 @@ func StringInSlice(valid []string, ignoreCase bool) schema.SchemaValidateFunc {
 			}
 		}
 
-		errors = append(errors, fmt.Errorf("expected %s to be one of %v, got %s", k, valid, v))
+		errors = append(errors, fmt.Errorf("expected %s to be one of %q, got %s", k, valid, v))
 		return warnings, errors
 	}
 }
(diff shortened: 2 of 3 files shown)


bflad (changes requested):
Hi @Frankkkkk 👋 Thank you for submitting this change and makes sense in this case. Please note the associated unit test, `TestValidationStringInSlice`, needs to be updated as well:

```
--- FAIL: TestValidationStringInSlice (0.00s)
    strings_test.go:283: expected test case 2 to produce error matching "expected [\w]+ to be one of \[ValidValue AnotherValidValue\], got VALIDVALUE", got [expected test_property to be one of ["ValidValue" "AnotherValidValue"], got VALIDVALUE]
```

Maintainer note: I think this differs from https://github.com/hashicorp/terraform-plugin-framework-validators/pull/152/files because over there, it was `%q` quoting framework `types.String` values instead of `string`.

Frankkkkk:
Hi @bflad thanks for your feed back. Sorry, I forgot to add the test file to the commit :(. It should be fixed.
Thanks !

bflad:
No worries, @Frankkkkk! If you're interested, this change likely should also have a changelog entry which you can add if you are interested. You can find information about adding one in the [contributing guide](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/.github/CONTRIBUTING.md#changelog). Something like "helper/validation: Added quoting in `StringInSlice` error diagnostic output to prevent confusion with values that contain spaces" -- if that's a little much, we can take care of that post-merge. Thanks again!

bflad (approved):
Looks great, thanks, @Frankkkkk 🚀

PR #1260, opened 2023-10-11
meta: Fix version number, deprecate exported variables and function

### LINKED ISSUE #1257: `meta.SDKVersion` reporting incorrect version
### SDK version

```
v2.29.0
```

https://github.com/hashicorp/terraform-plugin-sdk/blob/ec118d67dca1878e6582e392b0dad368b7092d91/meta/meta.go#L17

n/a

### Example usage

cloudflare/terraform-provider-cloudflare uses this value to append to the UA string.

https://github.com/cloudflare/terraform-provider-cloudflare/blob/658bb74d9a8ee1e258e10936bbe3c6c206ca4fe7/internal/sdkv2provider/provider.go#L380C3-L381

### Expected Behavior

`meta.SDKVersion` should report v2.29.0

### Actual Behavior

`meta.SDKVersion` reports v2.10.1

### References

- #190
- #444

bflad:
Closes #1257

The release process is also updated to automatically overwrite the version variable as it did in the previous release process. While this should ensure this version information is correct going forward, consumers should instead prefer the Go standard library `runtime/debug` package build information, which is guaranteed to always be correct.

### DIFF
--- a/meta/meta.go
+++ b/meta/meta.go
@@ -14,11 +14,17 @@ import (
 )
 
 // The main version number that is being run at the moment.
-var SDKVersion = "2.10.1"
+//
+// Deprecated: Use Go standard library [runtime/debug] package build information
+// instead.
+var SDKVersion = "2.30.0"
 
 // A pre-release marker for the version. If this is "" (empty string)
 // then it means that it is a final release. Otherwise, this is a pre-release
 // such as "dev" (in development), "beta", "rc1", etc.
+//
+// Deprecated: Use Go standard library [runtime/debug] package build information
+// instead.
 var SDKPrerelease = ""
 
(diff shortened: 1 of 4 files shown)


meta/meta.go
```diff
@@ -14,11 +14,17 @@ import (
 )
 
 // The main version number that is being run at the moment.
-var SDKVersion = "2.10.1"
+//
+// Deprecated: Use Go standard library [runtime/debug] package build information
+// instead.
+var SDKVersion = "2.30.0"
```
austinvalle, line 20: super nit: should this be `2.29.0`, then will be updated by the release when it runs

bflad: Yep, let's do that.

```suggestion
var SDKVersion = "2.29.0"
```

PR #1272, opened 2023-11-08
Update unit test doc example in AWS provider

drewmullen:
The included example hasnt been updated since the code reorganization which includes the exporting of the Flatten function. Trying to figure it out took me a couple minutes... figured it might be helpful to someone to just update the docs


bflad (approved):
Hi @drewmullen 👋 Thank you for submitting this. I'm going to approve and merge this under best effort since I feel like you folks would know best if/when other developers might be referencing this particular example. If there is any sort of dependence on this documentation, it might be best if the content is moved into the AWS provider documentation since it is not guaranteed that this generic SDK documentation will always keep such a specific provider example. Thanks!

PR #1307, opened 2024-01-29
all: Initial MoveResourceState implementation

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-go/pull/351
Reference: https://developer.hashicorp.com/terraform/plugin/framework/migrating

The next versions of the plugin protocol (5.5/6.5) include support for moving resource state across resource types. The terraform-plugin-sdk Go module will not be receiving this feature, however this Go module must be updated to handle the new RPC with errors.

Provider developers can implement move resource state in their terraform-plugin-sdk based providers by following the framework migration documentation to introduce terraform-plugin-mux and migrating the resource type to terraform-plugin-framework.


austinvalle (approved):
🚀

PR #1316, opened 2024-02-23
all: Modify CallFunction implementations

bendbennett:
Reference: hashicorp/terraform-plugin-go#380
Reference: https://developer.hashicorp.com/terraform/plugin/framework/migrating

The next versions of the plugin protocol (5.5/6.5) include support for provider defined functions. The terraform-plugin-sdk Go module will not be receiving this feature, however this Go module must be updated to handle the new RPC with errors.

Provider developers can implement provider defined functions by introducing terraform-plugin-mux and using their terraform-plugin-sdk based providers in combination with a terraform-plugin-framework provider that contains the provider defined functions.


go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/hashicorp/logutils v1.0.0
 	github.com/hashicorp/terraform-exec v0.20.0
 	github.com/hashicorp/terraform-json v0.21.0
-	github.com/hashicorp/terraform-plugin-go v0.21.0
+	github.com/hashicorp/terraform-plugin-go v0.22.0
```
bendbennett, line 17: Running `go get github.com/hashicorp/terraform-plugin-go` results in the addition of the following entry to `go.mod`:

```
toolchain go1.21.6
```

Which would result in an analogous [CI failure](https://github.com/hashicorp/terraform-plugin-mux/actions/runs/8017456549/job/21901361119#step:4:7) to that seen on _terraform-plugin-mux_

Given that our CI runs both Go `1.20` and `1.21`, I removed this directive, but perhaps this should be handled differently?

bflad (approved):
Looks good to me 🚀

PR #1335, opened 2024-05-06
all: Add automatic deferred action support for unknown provider configuration

austinvalle:
Ref: https://github.com/hashicorp/terraform-plugin-go/pull/403

This PR implements deferred action support for `PROVIDER_CONFIG_UNKNOWN` during the `ConfigureProvider` RPC. This functionality will allow a provider to defer all resources and data sources associated with the SDKv2 provider during the `ReadResource`, `ReadDataSource`, `ImportResourceState`, and `PlanResourceChange` RPCs.

Deferred actions is an experimental feature introduced in [`Terraform v1.9.0-alpha20240404`](https://github.com/hashicorp/terraform/releases/tag/v1.9.0-alpha20240404) and is only available in experimental Terraform builds.

More granular deferred action support will be available in `terraform-plugin-framework`.


go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/hashicorp/logutils v1.0.0
 	github.com/hashicorp/terraform-exec v0.20.0
 	github.com/hashicorp/terraform-json v0.21.0
-	github.com/hashicorp/terraform-plugin-go v0.22.2
+	github.com/hashicorp/terraform-plugin-go v0.22.3-0.20240506143625-1296f392daef
```
austinvalle, line 19: TODO: update when `terraform-plugin-go` is released

austinvalle: ⚡ done

.changes/unreleased/FEATURES-20240506-152018.yaml
bflad, line 1: Do we want to have a NOTE changelog that calls out deferred actions, its experimental-ness, and maybe mention working with HashiCorp if interested (and that's something we want)? I think we'll just want to be sure if someone sees the new feature/enhancement that they understand its probably not for them at the moment or feel like its something they should/need to implement. Hope that makes sense.

helper/schema/deferred_response.go
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package schema
+
+// MAINTAINER NOTE: Only PROVIDER_CONFIG_UNKNOWN (enum value 2 in the plugin-protocol) is relevant
+// for SDKv2. Since (DeferredResponse).DeferredReason is mapped directly to the plugin-protocol,
+// the other enum values are intentionally omitted here.
```
bflad, line 8: ❤️

helper/schema/grpc_provider.go
```diff
@@ -607,12 +607,37 @@ func (s *GRPCProviderServer) ConfigureProvider(ctx context.Context, req *tfproto
...
+
 	logging.HelperSchemaTrace(ctx, "Calling downstream")
 	diags := s.provider.Configure(ctxHack, config)
 	logging.HelperSchemaTrace(ctx, "Called downstream")
 
 	resp.Diagnostics = convert.AppendProtoDiag(ctx, resp.Diagnostics, diags)
 
+	if s.provider.providerDeferred != nil {
+		// Check if a deferred response was incorrectly set on the provider. This would cause an error during later RPCs.
+		if !configureDeferralAllowed(req.ClientCapabilities) {
```
bflad, line 623: Nit: Can this just check `!s.provider.deferralAllowed` here?

helper/schema/grpc_provider.go
```diff
@@ -607,12 +607,37 @@ func (s *GRPCProviderServer) ConfigureProvider(ctx context.Context, req *tfproto
...
+				Summary:  "Invalid Deferred Provider Response",
+				Detail: "Provider configured a deferred response for all resources and data sources but the Terraform request " +
+					"did not indicate support for deferred actions. This is an issue with the provider and should be reported to the provider developers.",
+			})
+		} else {
+			logging.HelperSchemaDebug(
+				ctx,
+				"Provider has configured a deferred response, all associated resources and data sources will automatically return a deferred response.",
+				map[string]interface{}{
+					"deferred_reason": s.provider.providerDeferred.Reason.String(),
```
bflad, line 635: Nit: Should this logging key be `tf_deferred_reason` to match terraform-plugin-go here? https://github.com/hashicorp/terraform-plugin-go/blob/05a54cd0f10761f941562bbf8846932d551e0c99/internal/logging/keys.go#L67

This Go module does have an internal constants file for logging keys, if that helps! https://github.com/hashicorp/terraform-plugin-sdk/blob/main/internal/logging/keys.go

Same with the other usage of this key.

helper/schema/grpc_provider.go
```diff
@@ -731,6 +772,25 @@ func (s *GRPCProviderServer) PlanResourceChange(ctx context.Context, req *tfprot
...
+	if s.provider.providerDeferred != nil && !res.ResourceBehavior.ProviderDeferred.EnablePlanModification {
+		logging.HelperSchemaDebug(
+			ctx,
+			"Provider has deferred response configured, automatically returning deferred response.",
+			map[string]interface{}{
+				"deferred_reason": s.provider.providerDeferred.Reason.String(),
+			},
+		)
+
+		resp.PlannedState = req.ProposedNewState
```
bflad, line 786: I think this is the best we can do in this situation without doing some real logic refactoring gymnastics 👍

helper/schema/grpc_provider.go
```diff
@@ -1145,6 +1220,60 @@ func (s *GRPCProviderServer) ImportResourceState(ctx context.Context, req *tfpro
...
+		}
+
+		resp.ImportedResources = []*tfprotov5.ImportedResource{
+			{
+				TypeName: req.TypeName,
+				State: &tfprotov5.DynamicValue{
+					MsgPack: unknownStateMp,
+				},
+			},
+		}
```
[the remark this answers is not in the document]
bflad: If Terraform will accept unknown for the resource state, I think that's preferable. 👍

helper/schema/grpc_provider_test.go
```diff
@@ -3337,6 +4977,240 @@ func TestApplyResourceChange_bigint(t *testing.T) {
...
+								Required: true,
+							},
+							"test_string": {
+								Type:     TypeString,
+								Computed: true,
+							},
+						},
+						Importer: &ResourceImporter{
+							StateContext: func(ctx context.Context, d *ResourceData, meta interface{}) ([]*ResourceData, error) {
+								t.Fatal("Test failed, import shouldn't be called")
```
bflad, line 5058: Nit: Do you want to use `(*testing.T).Error()` or `error` return so other test cases can run?

bflad (approved):
Overall looks good to me 🚀 Just some minor considerations along the way.

helper/schema/grpc_provider.go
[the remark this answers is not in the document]
austinvalle: Okay cool, I'll keep it like this and remove the comment. Also worth noting (not sure if this is obvious), but deferral is only possible with the `import` blocks (i.e. plannable import).

Using the `import` command I believe Terraform shouldn't be able to send an `AllowDeferral: true`. Luckily our corner tests "should" be able to test that scenario.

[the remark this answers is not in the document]
austinvalle: ⚡ aed628d9a0a01c4f6943aa6a344c16a1c918ae31

helper/schema/grpc_provider.go
austinvalle: I'm assuming you're just referring to the 2nd `if`, which can definitely be replaced by that -> 05985636e2f256b17c2f7149a624c4e084ea57b8 ⚡ 

The initial `s.provider.providerDeferred != nil` is still needed because we only want to raise a diagnostic if a deferral is returned by the provider incorrectly (since it won't be returned to `terraform-plugin-go`, this is really just preventing an error in later RPCs)

helper/schema/grpc_provider.go
austinvalle: Thanks for catching that! Added it to the constants -> bd21c5e19779c42be5748e84b62dff2bd77cee01 ⚡

helper/schema/grpc_provider.go
austinvalle: Yeah 🙁, we can always come back and give it a real attempt later if we think it's worth it 😆

helper/schema/grpc_provider_test.go
austinvalle: Yeah that's the proper way to do it 😆  -> c31bc0060e839dcc7dd7432c2a580be51a28f99c ⚡

.changes/unreleased/FEATURES-20240506-152018.yaml
austinvalle: Thanks for mentioning this, I was about half-way through the Framework PR and had a comment written up about it (will apply whatever we decide here over there)

Do you think 207c950e468b7ff8eeae648afaa35f88cee7c5b9 is enough? I'm not 100% sure where to direct the interest, but if you have a suggestion we can add more.

bflad: I think that is good for now! If someone really is curious or something, they can reach out via their normal channels whether that be talking to someone at HashiCorp or raising an issue in the repository.

bflad (approved):
Followup changes all look good to me still 😄

PR #1350, opened 2024-06-05
feat: add validation function for string length in bytes with min and max constraints

akhn1004:
### Summary

Adds a new validation function for validating string lengths in bytes with specified minimum and maximum constraints. The new function, `StringLenBytesBetween`, ensures that the length of a given string falls within the specified byte range.

### Background

In many cases, especially when dealing with APIs or data storage, it is crucial to validate that string inputs meet specific length requirements to prevent errors and ensure data integrity. The existing validation functions did not provide a way to validate string lengths based on byte count, which is important for handling multi-byte characters correctly.

### Related Issue

This issue has been raised in the context of ALB listener rules in Terraform AWS provider, as outlined in [hashicorp/terraform-provider-aws#37802](https://github.com/hashicorp/terraform-provider-aws/issues/37802). The problem manifests when attempting to set fixed responses with multi-byte characters. For example, Japanese characters, which are typically 2-3 bytes each, quickly exceed the byte limits imposed by ALB, leading to validation errors even though the number of characters might be within acceptable limits.

### DIFF
--- a/helper/validation/strings.go
+++ b/helper/validation/strings.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/structure"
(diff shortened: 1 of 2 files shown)


yutachaos:
@akhn1004 
Your PR is looks good.But looks like he hasn't signed the CLA.

PR #1355, opened 2024-07-02
feat(terraform): add ability to suppress github etag while being noisy when no actual state has been changed

aamkye:
Issues and PRs reference:
- https://github.com/integrations/terraform-provider-github/issues/796
- https://github.com/integrations/terraform-provider-github/issues/910
- https://github.com/hashicorp/terraform/issues/28803
- https://github.com/integrations/terraform-provider-github/pull/911
- https://github.com/integrations/terraform-provider-github/pull/909
- https://github.com/integrations/terraform-provider-github/pull/2296

The ETags are shown in diff on almost every plan, even after apply just before the next plan. After this change, ETag diff disappears. I've also added tests.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -980,7 +980,8 @@ func (m schemaMap) internalValidate(topSchemaMap schemaMap, attrsOnly bool) erro
 				return fmt.Errorf("%s: DefaultFunc is for configurable attributes,"+
 					"there's nothing to configure on computed-only field", k)
 			}
-			if v.DiffSuppressFunc != nil {
+			// Allow this behavior for etag, which is a special case (being too noisy)
+			if v.DiffSuppressFunc != nil && k != "etag" {
 				return fmt.Errorf("%s: DiffSuppressFunc is for suppressing differences"+
 					" between config and state representation. "+
 					"There is no config for computed-only field, nothing to compare.", k)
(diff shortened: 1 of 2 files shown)


aamkye:
@catsby could you look at this? :)

maciejlakomski (approved):

austinvalle:
Hey there @aamkye 👋🏻 , sorry you're running into trouble with the GitHub provider etag diffs.

Unforuntately, we won't be able to accept this contribution as this SDK is used by all Terraform providers. Computed attributes have no practical use with `DiffSupressFunc` as the main intention of that function is to prevent inconsequential differences between "config" and prior state. Computed attributes aren't defined in config, which is why that [original PR](https://github.com/integrations/terraform-provider-github/pull/909) in the GitHub provider caused an internal validation error.

Looking briefly through the GitHub provider issues, it seems like the intended usage of the etag is for internal use by the provider, so the ideal solution would be to use [private state](https://developer.hashicorp.com/terraform/plugin/framework/resources/private-state) to store that etag. That way, users can't reference `etag`, any drift in the value would not be reported to users, and the provider would still be able to reference that value for caching purposes.

Unfortunately, it seems like the GitHub provider resources referenced here are written in the older `terraform-plugin-sdk/v2`, which doesn't support private state. We don't have any plans to introduce private state to the older SDK, so the only way forward would be to [migrate](https://developer.hashicorp.com/terraform/plugin/framework/migrating) the GitHub provider resources over to [`terraform-plugin-framework`](https://developer.hashicorp.com/terraform/plugin/framework). I see there is an open issue to start that migration here: https://github.com/integrations/terraform-provider-github/issues/1912

PR #1375, opened 2024-09-10
all: Add support for write-only attributes

SBGoods:
Adds support for new write-only attribute feature in Terraform `v1.11` or higher. Write-only attributes are managed resource attributes whose values are not saved to the Terraform plan or state artifacts and can accept ephemeral values.

A provider can declare an attribute as write-only in the schema using the `WriteOnly` field. Write-only attributes values are only available in the configuration, and the prior state, planned state, and final state values for write-only attributes should always be `null`.


helper/schema/grpc_provider.go
```diff
@@ -281,6 +282,32 @@ func (s *GRPCProviderServer) ValidateResourceTypeConfig(ctx context.Context, req
 		resp.Diagnostics = convert.AppendProtoDiag(ctx, resp.Diagnostics, err)
 		return resp, nil
 	}
+	if req.ClientCapabilities == nil || !req.ClientCapabilities.WriteOnlyAttributesAllowed {
```
SBGoods, line 285: We might need to automatically set the client capabilities in `terraform-plugin-go`, so that we don't have to do this null check.

helper/validation/write_only.go
```diff
@@ -0,0 +1,88 @@
...
+			lastStep := attrPath[pathLen-1]
+
+			// Only attribute steps have a Name field
+			attrStep, ok := lastStep.(cty.GetAttrStep)
+			if !ok {
+				resp.Diagnostics = diag.Diagnostics{
+					{
+						Severity:      diag.Error,
+						Summary:       "Invalid oldAttributePath",
+						Detail:        "The specified oldAttribute path must point to an attribute.",
```
SBGoods, line 65: Since this is technically an error for the provider developer and not the practitioner, we need to add to this message indicating that it should be reported to the provider developers.

helper/validation/write_only.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// PreferWriteOnlyAttribute is a ValidateResourceConfigFunc that returns a warning
+// if the Terraform client supports write-only attributes and the old attribute is
+// not null.
+// The last step in the path must be a cty.GetAttrStep{}.
+// When creating a cty.IndexStep{} to into a nested attribute, use an unknown value
+// of the index type to indicate any key value.
```
austinvalle, line 21: Mentioned this offline, but I'm wondering what core/James might say about modeling this type of path with unknowns? Could also consider just accepting a slice of `AttrStep`? Both seem awkward to manage but we don't have a ton of options unless we want to roll a new path expression implementation (which is also an option 😆)

helper/validation/write_only.go
```diff
@@ -0,0 +1,88 @@
...
+// For lists: cty.Index(cty.UnknownVal(cty.Number)),
+// For maps: cty.Index(cty.UnknownVal(cty.String)),
+// For sets: cty.Index(cty.UnknownVal(cty.Object(nil))),
+func PreferWriteOnlyAttribute(oldAttribute cty.Path, writeOnlyAttributeName string) schema.ValidateResourceConfigFunc {
```
austinvalle, line 25: Should `writeOnlyAttributeName` be a cty.Path as well?

helper/validation/write_only.go
```diff
@@ -0,0 +1,88 @@
...
+				}
+				return
+			}
+
+			if !attr.value.IsNull() {
+				resp.Diagnostics = append(resp.Diagnostics, diag.Diagnostic{
+					Severity: diag.Warning,
+					Summary:  "Available Write-Only Attribute Alternative",
+					Detail: fmt.Sprintf("The attribute %s has a WriteOnly version %s available. "+
+						"Use the WriteOnly version of the attribute when possible.", attrStep.Name, writeOnlyAttributeName),
```
austinvalle, line 77: We may want to bikeshed this warning message wording with the provider development teams

helper/validation/write_only.go
```diff
@@ -0,0 +1,88 @@
...
+						Severity:      diag.Error,
+						Summary:       "Invalid oldAttributePath",
+						Detail:        "The specified oldAttribute path must point to an attribute.",
+						AttributePath: attrPath,
+					},
+				}
+				return
+			}
+
+			if !attr.value.IsNull() {
```
austinvalle, line 72: Will be interesting to test this, but I have a feeling for practitioners that are already using the resource, the SDK won't be able to completely null this value out. We may have to check the cty.Value against zero values as well 😞 (empty string, zero, etc.)

I don't know off the top of my head if that's valid, but I'm pretty sure there's nothing a provider dev/practitioner can do to null a value in SDKv2 once it has a value 🤔

austinvalle: A corner provider test could shine some light on this problem 💭

helper/validation/write_only_test.go
```diff
@@ -0,0 +1,780 @@
...
+						cty.IndexStep{Key: cty.ObjectVal(map[string]cty.Value{
+							"oldAttribute":       cty.StringVal("value3"),
+							"writeOnlyAttribute": cty.NullVal(cty.String),
+						})},
+						cty.GetAttrStep{Name: "oldAttribute"},
+					},
+				},
+			},
+		},
+		"map nested block: oldAttribute and writeOnlyAttribute map returns warning diags": {
```
austinvalle, line 490: Similar question to one I had above, are map nested blocks in SDKv2? Or is this a map attribute equivalent

helper/validation/path.go
```diff
@@ -0,0 +1,55 @@
...
+package validation
+
+import (
+	"github.com/hashicorp/go-cty/cty"
+)
+
+// PathEquals compares two Paths for equality. For cty.IndexStep,
+// unknown key values are treated as an Any qualifier and will
+// match any index step of the same type.
+func PathEquals(p cty.Path, other cty.Path) bool {
```
austinvalle, line 13: nit: Just a naming thing, since this is kind of fuzzy/expression matching, maybe something like `PathMatches`? Perhaps we can do a formal bikeshed for naming later 😄

austinvalle (commented):
All of this is looking awesome, great work! I left some initial comments, but in addition to that:
- I'm wondering what will eventually show up in the plan renderer for Terraform... since we are planning non-null values that will eventually be set to null. Feels like it should either explicitly mention that the value is write-only, or we should be planning `null` for write-only attributes as well. We should make a note to come back to this once core gets further along.
- It'd be cool to get some corner testing for SDKv2 to catch any regressions we might introduce in the future
- I'd love to run this development branch against a major cloud TF provider (aws/gcp/azure) if we can get a hand from their teams finding the appropriate CI jobs to do so. They have so many SDKv2 schemas + tests that it would be a good smoke test.
  - We would need them to update their plugin-go/mux/etc, so you might need to adjust the plugin-go branch to make that all compile.

helper/schema/grpc_provider.go
```diff
@@ -1182,6 +1219,8 @@ func (s *GRPCProviderServer) ApplyResourceChange(ctx context.Context, req *tfpro
 
 	newStateVal = copyTimeoutValues(newStateVal, plannedStateVal)
 
+	newStateVal = setWriteOnlyNullValues(newStateVal, schemaBlock)
```
austinvalle, line 1222: I wonder if we should do this in the `ReadResource` RPC as well? It's possible the provider could attempt to refresh the state with a value in a write-only attribute. Terraform core doesn't typically do any validation/assertions then, so there wouldn't be an error.

That seems like being potentially defensive, but we did have a bug about unknown values making it state during refresh recently lol: 
- See outcome and linked issues for more context: https://github.com/hashicorp/terraform-plugin-framework/issues/902

SBGoods: Will it be possible for this to affect state? According to the current proposal, write-only attributes will be completely omitted from the state file, so I would be surprised that Terraform Core would write the value even if we do send one across the wire.

helper/schema/grpc_provider_test.go
```diff
@@ -5857,1099 +5740,2877 @@ func TestReadDataSource(t *testing.T) {
...
-			configBytes, err := msgpack.Marshal(config, schema.ImpliedType())
-			if err != nil {
-				t.Fatal(err)
-			}
-
-			testReq := &tfprotov5.ApplyResourceChangeRequest{
-				TypeName: "test",
-				PriorState: &tfprotov5.DynamicValue{
-					MsgPack: priorState,
+		"Map nested block, WriteOnly attribute with multiple values returns multiple diags": {
```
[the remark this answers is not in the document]
SBGoods: I honestly don't know what type combinations are possible in the SDKv2 😂. I'll circle back to this after I get some SDKv2 regression tests in the corner provider.

helper/validation/write_only.go
SBGoods: I originally had this as a `cty.Path` but I changed it since the only thing we would be doing with the path is getting the attribute name. I didn't want the parameter type to give the impression that we are actually walking/validating this path, but we can change it back to a `cty.Path` for future proofing.

helper/schema/grpc_provider.go
austinvalle: The current relationship between core and the providers is that the provider is responsible for producing data that satisfies Terraform's constraints. In the case that the provider doesn't, Terraform raises an error diagnostic, rather than silently ignoring the incorrect data.

So if that same principle is applied here, I would expect an error diagnostic if the provider incorrectly sets a write-only attribute. However, we are discussing logic that doesn't exist yet in core, so it's possible they won't take this route. Perhaps we can just note to revisit this when core is closer to implementation

helper/schema/resource_data.go
```diff
@@ -604,6 +607,67 @@ func (d *ResourceData) GetRawConfig() cty.Value {
...
+				Severity: diag.Error,
+				Summary:  "Invalid config path",
+				Detail: "The Terraform Provider unexpectedly provided a path that does not match the current schema. " +
+					"This can happen if the path does not correctly follow the schema in structure or types. " +
+					"Please report this to the provider developers. \n\n" +
+					"Cannot find config value for given path.",
+				AttributePath: valPath,
+			},
+		}
+	}
```
austinvalle, line 666: (adding comments as I run into stuff)

I don't think we should return an error in this scenario. This would also trigger if you had an optional attribute that does exist in the schema, the value is found during walk, but the actual value is null.

We could consider adding a boolean to determine if we found something during the walk.

helper/schema/resource_data.go
```diff
@@ -604,6 +607,67 @@ func (d *ResourceData) GetRawConfig() cty.Value {
...
 
+// GetRawConfigAt is a helper method for retrieving specific values
+// from the RawConfig returned from GetRawConfig. It returns the cty.Value
+// for a given cty.Path or an error diagnostic if the value at the given path does not exist.
+//
+// GetRawConfigAt is considered advanced functionality, and
+// familiarity with the Terraform protocol is suggested when using it.
+func (d *ResourceData) GetRawConfigAt(valPath cty.Path) (cty.Value, diag.Diagnostics) {
+	rawConfig := d.GetRawConfig()
+	configVal := cty.NullVal(cty.EmptyObject)
```
austinvalle, line 618: I wonder what cty typically uses as the default "we don't know what the type is" value.

I'd imagine it's something like an unknown dynamic pseudo type

helper/schema/schema.go
```diff
@@ -2353,6 +2384,36 @@ func (m schemaMap) validateType(
 	return diags
 }
 
+// hasWriteOnly returns true if the schemaMap contains any WriteOnly attributes.
```
austinvalle, line 2387: One additional usage of this `hasWriteOnly` I think we missed is applying this to managed resource blocks that are not write-only themselves. For example, ATM, this schema is considered valid for a managed resource:

```go
map[string]*schema.Schema{
	"nested_list_block": {
		Type:     schema.TypeList,
		Optional: true,
		Computed: true,
		Elem: &schema.Resource{
				"writeonly_string": {
					Type:      schema.TypeString,
					Required:  true,
					WriteOnly: true,
				},
			},
		},
	},
},
```

austinvalle: In the RFC, this specific rule was: `Block types with Computed set to true cannot contain WriteOnly attributes`

helper/schema/resource_data.go
SBGoods: It seems to be [DynamicVal](https://github.com/zclconf/go-cty/blob/main/cty/unknown.go#L83), I've updated the default value to that which should also fix the false positive "Invalid config path" error thrown above.

helper/schema/schema.go
austinvalle, line 1: Leaving a comment here to make sure we circle back on this before merging.

I think we should consider adding more internal validation to explicitly prevent using schema fields like `ForceNew` with `WriteOnly`. One of the awkward parts of write-only attributes is that they always appear in the resource diff when a value is present (since prior state is always null). Combining that with something like `ForceNew` means that the provider always indicates the resource should be replaced in the plan response.

At the moment, Terraform is ignoring that because the path in `requires_replace` is a write-only attribute, although I'm not sure if that is on purpose or not. SDKv2 also has some legacy behavior of always adding `id` to the `requires_replace`, although Terraform doesn't seem to be acknowledging that either 🤔 .

Overall, it doesn't make sense for write-only attributes to use the `ForceNew` field since Terraform can't do anything with that information, so restricting it will at least prevent any awkward side effects from occurring. Unforuntately, I don't think we can't restrict `CustomizeDiff` to not being used with `WriteOnly` (which can also indicate replacement), as that's the only place a provider can do validation that is aware of any of the resource's prior state (for example, to know if that resource is being created or not).

helper/schema/resource_diff.go
```diff
@@ -480,6 +481,67 @@ func (d *ResourceDiff) GetRawConfig() cty.Value {
 	return cty.NullVal(schemaMap(d.schema).CoreConfigSchema().ImpliedType())
 }
 
+// GetRawConfigAt is a helper method for retrieving specific values
+// from the RawConfig returned from GetRawConfig. It returns the cty.Value
+// for a given cty.Path or an error diagnostic if the value at the given path does not exist.
+//
+// GetRawConfigAt is considered advanced functionality, and
+// familiarity with the Terraform protocol is suggested when using it.
+func (d *ResourceDiff) GetRawConfigAt(valPath cty.Path) (cty.Value, diag.Diagnostics) {
```
austinvalle, line 490: I added a duplicate implementation here for resource diff for usage with `CustomizeDiff` (most likely usage would be running some type of validation that requires knowing about prior state (which can be checked with `(*ResourceDiff).GetRawState`)

austinvalle:
@SBGoods Since we are already "managing" the setting of `null` for write-only attributes during plan/apply/read/etc. we need to ensure that we do this logic in every RPC where TF core is validating the `null` values in state (which is every RPC that has state as a response value). Currently I think this PR is missing the nulling logic on:
- ImportResourceState
- UpgradeResourceState

helper/schema/resource_data.go
```diff
@@ -604,6 +607,67 @@ func (d *ResourceData) GetRawConfig() cty.Value {
 	return cty.NullVal(schemaMap(d.schema).CoreConfigSchema().ImpliedType())
 }
 
+// GetRawConfigAt is a helper method for retrieving specific values
+// from the RawConfig returned from GetRawConfig. It returns the cty.Value
+// for a given cty.Path or an error diagnostic if the value at the given path does not exist.
+//
+// GetRawConfigAt is considered advanced functionality, and
+// familiarity with the Terraform protocol is suggested when using it.
+func (d *ResourceData) GetRawConfigAt(valPath cty.Path) (cty.Value, diag.Diagnostics) {
```
jbardin, line 616: I would check that this works for nested blocks; specifically traversing across `cty.SetVal` values is weird and I think `cty.Walk` is going to fail there.

helper/validation/write_only.go
```diff
@@ -0,0 +1,116 @@
...
+					AttributePath: writeOnlyAttribute,
+				},
+			}
+			return
+		}
+
+		var oldAttrs []attribute
+
+		err := cty.Walk(req.RawConfig, func(path cty.Path, value cty.Value) (bool, error) {
+			if PathMatches(path, oldAttribute) {
```
jbardin, line 59: should all `cty.Walk' calls use `PathMatches`? Or are they expected to only have paths from what should be matching objects?

helper/schema/resource_data.go
SBGoods: Good callout! I've added more test cases in commit [d1d4c14](https://github.com/hashicorp/terraform-plugin-sdk/pull/1375/commits/d1d4c14d859e91a7dbbd8bb12aad9cfadf0377c0). The Set and Set nested block traversal test cases are passing 👍🏾

website/docs/plugin/sdkv2/resources/index.mdx
```diff
@@ -30,3 +30,12 @@ Terraform has data consistency rules for resources, which may not be easily disc
...
 When resource implementations change (due to bug fixes, improvements, or changes to the backend APIs Terraform interacts with), they can sometimes become incompatible with existing state. When this happens, a migration is needed for resources provisioned in the wild with old schema configurations. Terraform resources support migrating state values in these scenarios via [State Migration](/terraform/plugin/sdkv2/resources/state-migration).
+
+## Write-only Attributes
+
+~> **NOTE:** Write-only attributes are only supported in Terraform `v1.11` or higher
+
+Write-only attributes are a special type of resource attribute
+whose values are not sent to Terraform and do not persist in the Terraform plan or state artifacts.
+Write-only attributes can accept [ephemeral values](/terraform/language/resources/ephemeral).
+The [Write-only attribute] page discusses how to create these attributes.
```
rkoron007, line 41: @SBGoods I think you are missing the link here?

helper/schema/resource_data.go
```diff
@@ -604,6 +607,67 @@ func (d *ResourceData) GetRawConfig() cty.Value {
...
+	rawConfig := d.GetRawConfig()
+	configVal := cty.DynamicVal
+
+	if rawConfig.IsNull() {
+		return configVal, diag.Diagnostics{
+			{
+				Severity: diag.Error,
+				Summary:  "Invalid config path",
+				Detail: "The Terraform Provider unexpectedly provided a path that does not match the current schema. " +
+					"This can happen if the path does not correctly follow the schema in structure or types. " +
```
austinvalle, line 626: If `rawConfig.IsNull()`, does that mean the provided `valPath` is invalid? Seems more like the message should be geared towards "unexpected error", since raw config should always be non-null.

helper/schema/resource_diff.go
```diff
@@ -480,6 +481,67 @@ func (d *ResourceDiff) GetRawConfig() cty.Value {
...
+	rawConfig := d.GetRawConfig()
+	configVal := cty.DynamicVal
+
+	if rawConfig.IsNull() {
+		return configVal, diag.Diagnostics{
+			{
+				Severity: diag.Error,
+				Summary:  "Invalid config path",
+				Detail: "The Terraform Provider unexpectedly provided a path that does not match the current schema. " +
+					"This can happen if the path does not correctly follow the schema in structure or types. " +
```
austinvalle, line 500: Same note as `ResourceData` about the error message

helper/schema/resource_data_test.go
```diff
@@ -3915,6 +3919,162 @@ func TestResourceData_nonStringValuesInMap(t *testing.T) {
...
+						"ConfigAttribute": cty.StringVal("valueA"),
+					}),
+					cty.ObjectVal(map[string]cty.Value{
+						"ConfigAttribute": cty.StringVal("valueB"),
+					}),
+				}),
+			}),
+			Path: cty.GetAttrPath("set_nested_block").Index(cty.ObjectVal(map[string]cty.Value{
+				"ConfigAttribute": cty.StringVal("valueB"),
+			})).GetAttr("ConfigAttribute"),
```
austinvalle, line 4031: Feels like for sets it'd just be easier for provider developers to grab the root, but glad it works!

helper/schema/resource_data_test.go
```diff
@@ -3915,6 +3919,162 @@ func TestResourceData_nonStringValuesInMap(t *testing.T) {
 	}
 }
 
+func TestResourceDataGetRawConfigAt(t *testing.T) {
```
austinvalle, line 3922: Another test to consider here is when the attribute is valid in the schema, but the value is actually null (should be  `cty.NullVal(<cty.Type>)` and not be `cty.DynamicVal`)

helper/schema/write_only.go
```diff
@@ -0,0 +1,214 @@
...
+	sort.Strings(attrNames)
+
+	for _, name := range attrNames {
+		attr := schema.Attributes[name]
+		v := valMap[name]
+
+		if attr.WriteOnly && !v.IsNull() {
+			diags = append(diags, diag.Diagnostic{
+				Severity: diag.Error,
+				Summary:  "WriteOnly Attribute Not Allowed",
```
austinvalle, line 145:
```suggestion
				Summary:  "Write-only Attribute Not Allowed",
```

helper/schema/write_only.go
```diff
@@ -0,0 +1,214 @@
...
+				Detail: fmt.Sprintf("The resource contains a non-null value for WriteOnly attribute %q ", name) +
```
austinvalle, line 146:
```suggestion
				Detail: fmt.Sprintf("The resource contains a non-null value for write-only attribute %q ", name) +
```

helper/validation/write_only.go
```diff
@@ -0,0 +1,116 @@
...
+				}
+				return
+			}
+
+			if !attr.value.IsNull() {
+				resp.Diagnostics = append(resp.Diagnostics, diag.Diagnostic{
+					Severity: diag.Warning,
+					Summary:  "Available Write-Only Attribute Alternative",
+					Detail: fmt.Sprintf("The attribute %s has a WriteOnly version %s available. "+
+						"Use the WriteOnly version of the attribute when possible.", attrStep.Name, writeOnlyAttrStep.Name),
```
austinvalle, line 105: Small wording suggestion

```suggestion
					Summary:  "Available Write-only Attribute Alternative",
					Detail: fmt.Sprintf("The attribute %s has a write-only alternative %s available. "+
						"Use the write-only alternative of the attribute when possible.", attrStep.Name, writeOnlyAttrStep.Name),
```

helper/validation/write_only.go
austinvalle: I still think this will be a problem, but we don't have to solve it at this exact moment. Can always come back later and add detecting default/zero values if it makes sense.

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
+---
+page_title: Resources - Write-only Attributes
+description: Implementing write-only attributes within resources.
+---
+
+# Resources - Write-only Attributes
+
+~> **NOTE:** Write-only attributes are only supported in Terraform `v1.11` or higher
+
+Write-only attributes are managed resource attributes whose values are not saved to the Terraform plan or state artifacts. Write-only attributes
```
austinvalle, line 10:
```suggestion
Write-only attributes are managed resource attributes that are configured by practitioners but are not persisted to the Terraform plan or state artifacts. Write-only attributes
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+should be used to handle secret values that do not need to be persisted in Terraform state, such as passwords, API keys, etc.
+The provider is expected to be the terminal point for an ephemeral value,
+which should either use the value by making the appropriate change to the API or ignore the value. Write-only attributes can accept [ephemeral values](/terraform/language/resources/ephemeral).
```
austinvalle, line 13: Maybe something like this?

```suggestion
which should either use the value by making the appropriate change to the API or ignore the value. Write-only attributes can accept [ephemeral values](/terraform/language/resources/ephemeral) and are not required to be consistent between plan and apply operations.
```

website/docs/plugin/sdkv2/resources/index.mdx
```diff
@@ -30,3 +30,12 @@ Terraform has data consistency rules for resources, which may not be easily disc
...
 
 When resource implementations change (due to bug fixes, improvements, or changes to the backend APIs Terraform interacts with), they can sometimes become incompatible with existing state. When this happens, a migration is needed for resources provisioned in the wild with old schema configurations. Terraform resources support migrating state values in these scenarios via [State Migration](/terraform/plugin/sdkv2/resources/state-migration).
+
+## Write-only Attributes
+
+~> **NOTE:** Write-only attributes are only supported in Terraform `v1.11` or higher
+
+Write-only attributes are a special type of resource attribute
+whose values are not sent to Terraform and do not persist in the Terraform plan or state artifacts.
+Write-only attributes can accept [ephemeral values](/terraform/language/resources/ephemeral).
```
austinvalle, line 40:
```suggestion
Write-only attributes can accept [ephemeral values](/terraform/language/resources/ephemeral) and are not required to be consistent between plan and apply operations..
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+## General Concepts
+
+The following are high level differences between regular attributes and write-only attributes:
```
austinvalle, line 17: Perhaps instead of "regular", use language like "configuration" or just directly reference `Required` / `Optional` attributes. "Non-computed" is also a potential option

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+- Write-only attributes can accept ephemeral and non-ephemeral values
+
+- Write-only attribute values are only available in the configuration, and the prior state, planned state, and final state values for
```
austinvalle, line 21:
```suggestion
- Write-only attribute values are only available in the configuration. The prior state, planned state, and final state values for
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+write-only attributes should always be `null`.
+    - Provider developers do not need to “clean up” write-only attribute values after using them as the SDKv2 will handle the nullification of write-only attributes for all RPCs.
```
austinvalle, line 23:
```suggestion
    - Provider developers do not need to explicitly set write-only attribute values to `null` after using them as the SDKv2 will handle the nullification of write-only attributes for all RPCs.
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+- Any value that is set for a write-only attribute using `(*ResourceData).Set()` by the provider will be set to `null` by SDKv2 before the RPC response is sent to Terraform
```
austinvalle, line 25:
```suggestion
- Any value that is set for a write-only attribute using `(*ResourceData).Set()` by the provider will be reverted to `null` by SDKv2 before the RPC response is sent to Terraform
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+- Write-only attribute values on its own will never cause a Terraform diff.
```
austinvalle, line 27:
```suggestion
- Write-only attribute values cannot produce a Terraform plan difference.
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+  WriteOnly: true,
+},
+```
+
+**Restrictions:**
+
+- Cannot be used in data source or provider schemas.
+- Must be set with either `Required` is `true` or `Optional` is `true`
+- Cannot be used when `Computed` is `true`
+- Cannot be used with aggregate schema types (e.g. `typeMap`, `typeList`, `typeSet`)
```
austinvalle, line 50: Some considerations:
- Can't be used with `ForceNew`
- WriteOnly can't be set directly on an aggregate schema type, but it can be used within one (as long as it's not computed). We should make that distinction somewhere

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+// Set traversal
+cty.GetAttrPath("set_attribute").Index(cty.String("set_string_val"))
+
+
+// Set nested block traversal
+cty.GetAttrPath("set_nested_block").
+Index(cty.ObjectVal(map[string]cty.Value{"set_nested_block_attribute": cty.StringVal("valueA")}))
+.GetAttr("set_nested_block_attribute") // returns path to cty.String("valueA")
+
+```
```
austinvalle, line 111: This documentation is great! One potential consideration is maybe linking each `cty.Path` example to a Terraform config example. That might help make readers make the mental jump

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+### cty.Value
+
+When working with `cty.Value`, you must always check the type of the value before converting it to a Go value or else the conversion will cause a panic.
```
austinvalle, line 115:
```suggestion
When working with `cty.Value`, you must always check the type of the value before converting it to a Go value or else the conversion could cause a panic.
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Check if the value is not null
+if !woVal.IsNull() {
+    // Now we can safely convert to a Go string
+    encryptedValue = woVal.AsString()
+}
+```
+
+## PreferWriteOnly Validator
```
austinvalle, line 130:
```suggestion
## PreferWriteOnlyAttribute Validator
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+`PreferWriteOnlyAttribute()` is a validator that takes a `cty.Path` to a regular attribute and a `cty.Path` to a write-only attribute.
```
austinvalle, line 132: Not sure on the wording, but just avoiding regular

```suggestion
`PreferWriteOnlyAttribute()` is a validator that takes a `cty.Path` to an existing configuration attribute (required/optional) and a `cty.Path` to a write-only attribute.
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+When using `cty.Path` to traverse into a nested block, use an unknown value to indicate any key value:
+
+- For lists: `cty.Index(cty.UnknownVal(cty.Number))`,
+- For maps: `cty.Index(cty.UnknownVal(cty.String))`,
+- For sets: `cty.Index(cty.UnknownVal(cty.Object(nil)))`,
+
+
+## Best Practices
+
+Since write-only attributes have no prior values, we cannot determine user intent with a write-only attribute alone. To determine when to use/not use a write-only attribute value in your provider, we recommend using other attributes in the provider. For example:
```
austinvalle, line 173: Not sure the wording 🤔 

```suggestion
Since write-only attributes have no prior values, user intent cannot be determined with a write-only attribute alone. To determine when to use/not use a write-only attribute value in your provider, we recommend using other non-write-only attributes in the provider. For example:
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+- Pair write-only attributes with a regular attribute to “trigger” the use of the write-only attribute
```
austinvalle, line 175:
```suggestion
- Pair write-only attributes with a configuration attribute (required or optional) to “trigger” the use of the write-only attribute
```

website/docs/plugin/sdkv2/resources/write-only-attributes.mdx
```diff
@@ -0,0 +1,177 @@
...
+    - For example, a `password_wo` write-only attribute can be paired with a configured `password_wo_version` attribute. When the `password_wo_version` updates, the provider will send the `password_wo` value to the API.
```
austinvalle, line 176:
```suggestion
    - For example, a `password_wo` write-only attribute can be paired with a configured `password_wo_version` attribute. When the `password_wo_version` is modified, the provider will send the `password_wo` value to the API.
```

austinvalle (approved):
Looks good! I left some comments, mostly just considerations for documentation. (will re-approve if there are more changes pushed to the PR)

.changes/unreleased/FEATURES-20250121-165644.yaml
```diff
@@ -0,0 +1,6 @@
+kind: FEATURES
+body: 'helper/schema: Added `WriteOnly` schema behavior for managed resource schemas to indicate a write-only attribute. 
+Write-only attributes are managed resource attributes whose values are not saved to the Terraform plan or state artifacts.'
```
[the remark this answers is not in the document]
rkoron007: @SBGoods Adding on to Austin's suggestion and to get us on the same page, in the docs we renamed these "write-only arguments" because that is what the [syntax docs recommend](https://developer.hashicorp.com/terraform/language/syntax/configuration#arguments) and [got the rest of the team's sign-off](https://hashicorp.slack.com/archives/C071HC4JJCC/p1736793639409839?thread_ts=1736359755.647449&cid=C071HC4JJCC) on using "arguments" too. 

Happy to discuss if you have any feedback on that though!

```suggestion
Write-only arguments are managed resource attributes whose values are not saved to the Terraform plan or state artifacts.'
```

website/docs/plugin/sdkv2/resources/index.mdx
```diff
@@ -30,3 +30,12 @@ Terraform has data consistency rules for resources, which may not be easily disc
 Resources define the data types and API interactions required to create, update, and destroy infrastructure with a cloud vendor, while the [Terraform state](/terraform/language/state) stores mapping and metadata information for those remote objects.
 
 When resource implementations change (due to bug fixes, improvements, or changes to the backend APIs Terraform interacts with), they can sometimes become incompatible with existing state. When this happens, a migration is needed for resources provisioned in the wild with old schema configurations. Terraform resources support migrating state values in these scenarios via [State Migration](/terraform/plugin/sdkv2/resources/state-migration).
+
+## Write-only Attributes
```
rkoron007, line 34: Commented on this above, but suggest renaming everywhere to  "Write-only arguments" for consistency!

rkoron007 (commented):
I see there are a bunch of suggestions here, so I kept my review light. Happy to go through this again after you get a chance to review Austin's comments! ✨

helper/schema/grpc_provider_test.go
```diff
@@ -4252,23 +5236,23 @@ func TestPlanResourceChange(t *testing.T) {
 					MsgPack: mustMsgpackMarshal(
 						cty.Object(map[string]cty.Type{
 							"id":  cty.String,
-							"foo": cty.Number,
+							"foo": cty.String,
 						}),
 						cty.ObjectVal(map[string]cty.Value{
 							"id":  cty.UnknownVal(cty.String),
-							"foo": cty.NullVal(cty.Number),
+							"foo": cty.StringVal("bar"),
```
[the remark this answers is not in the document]
SBGoods: I thought Terraform Core is setting proposed new state with the config for write-only attributes? It's only that we can't populate it in the `(*ResourceData).Get()` in the CRUD methods.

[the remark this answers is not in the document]
austinvalle: Ah yes, you're correct. During Plan `ProposedNewState` will have the config values. I think I was thinking about `PlannedNewState` during Apply 🫤

pbortnick (commented):
All good from Web Presence POV

austinvalle (approved):
Great work! 🚀

PR #1426, opened 2025-02-17
fix: Correctly validate the length of a UTF8 string with `StringLenBetween` validator

### LINKED ISSUE #1425: Validating strings with `StringLenBetween` incorrectly fails for UTF-8 encoded values
### SDK version

```
HEAD
```

```go
...
```

```hcl
...
```

### Debug Output

n/a

### Expected Behavior

With a provider that wants to validate a String with `StringLenBetween(1, 5)` and a value of `"你好世界"` is provided, the validation should be successful.

### Actual Behavior

Fails with `expected length of test_property to be in the range (1 - 5), got 你好世界`.

### Steps to Reproduce

Unit test in https://github.com/hashicorp/terraform-plugin-sdk/commit/2f48bd87b4954d8ad1956cf6f4cdce4eea25a35a

### References

n/a

goodtune:
Fixes #1425.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20250217-181939.yaml
+++ b/.changes/unreleased/BUG FIXES-20250217-181939.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'tfsdk: Correctly validate the length of a UTF8 string with `StringLenBetween` validator'
+time: 2025-02-17T18:19:39.713656+11:00
+custom:
+    Issue: "1425"
--- a/helper/validation/meta_test.go
+++ b/helper/validation/meta_test.go
@@ -71,7 +71,7 @@ func TestValidationAll(t *testing.T) {
 		{
 			val: "你好世界",
 			f: All(
-				StringLenBetween(1,5),
+				StringLenBetween(1, 5),
 			),
 		},
 	})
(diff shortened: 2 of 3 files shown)


goodtune:
@austinvalle sorry for a direct ping, but I noticed you had reviewed a recent community contribution.

I came to this solution because of a third-party provider which simply doesn’t work in the real world because of this issue (the NetBox API has a field of length 200, we have Chinese characters in the field and can return it from the API, but the provider won’t play with it) that is preventing me from encouraging further adoption of Terraform.

What is the process for getting pull requests and issues on the radar for feedback?

austinvalle:
Hey there @goodtune 👋🏻, thanks for filing the issue + PR and apologies for the slow response.

Unfortunately, I think it's unlikely we'd accept this change only for the fact that this exported function has existed for a couple years, and any usage of it in all existing providers are relying on this behavior (whether deemed correct or not, it's not explicitly documented that the string length being validated is UTF-8 based). 

As well, this specific helper function is just implementing an interface which is exported https://github.com/hashicorp/terraform-plugin-sdk/blob/fe25e82bcf8d6744ff336c66d3cc35164eb15271/helper/schema/schema.go#L486-L494 so it's easy to copy and adjust it to fit your UTF-8 use-case.

------

I'd recommend just copying the function into your provider codebase and creating your own UTF-8 alternative.

PR #1427, opened 2025-02-18
helper/schema: Add validation to prevent write-only attributes in set nested blocks

SBGoods:
Write-only attributes when used with ephemeral data cannot be used within any set and set-based data due to limitations in Terraform Core. This PR adds validation to block write-only attributes from being used within set nested blocks.

### DIFF
--- a/.changes/unreleased/NOTES-20250218-172625.yaml
+++ b/.changes/unreleased/NOTES-20250218-172625.yaml
@@ -1,0 +1,5 @@
+kind: NOTES
+body: Write-only attribute support is in technical preview and offered without compatibility promises until Terraform 1.11 is generally available.
+time: 2025-02-18T17:26:25.941391-05:00
+custom:
+    Issue: "1375"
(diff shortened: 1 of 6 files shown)


SBGoods (commented):
Failing CI checks will be resolved after https://github.com/hashicorp/terraform-plugin-sdk/pull/1428 is merged

.changes/unreleased/NOTES-20250218-172625.yaml
austinvalle, line 1: I don't think we **_have_** to repeat this changelog, but I'm not expressly against it 😆

helper/schema/schema_test.go
```diff
@@ -5303,7 +5303,7 @@ func TestSchemaMap_InternalValidate(t *testing.T) {
 			},
 			true,
 		},
-		"Set configuration block nested attribute with WriteOnly set returns no errors": {
+		"Set configuration block nested attribute with WriteOnly set returns error": {
```
austinvalle, line 5306: Nice that we already had tests for it!

austinvalle (approved):
LGTM :shipit:

PR #1444, opened 2025-03-14
Resource Identity: Add identity data to existing RPCs and add two new identity related RPCs"

ansgarm:
This PR adds initial support for Resource Identity to the SDKv2. It's in an early stage, meaning it's only lightly tested and mostly covering the happy path.

- **update plugin go**
- **initial generate of RPC + small shim test**
- **wip - resource identity schema**
- **Added initial testing for getresourceidentityschemas**
- **Added initial attempt for GetResourceIdentitySchemas and UpgradeResourceIdentity.**
- **start implementing identity in ReadResource – test doesn't work yet, but getting close**
- **Added more for GetResourceIdentitySchemas and UpgradeResourceIdentity.**
- **Copied CoreIdentitySchema to CoreResourceIdentitySchema and started ConfigIdentitySchemaToProto and coreConfigIdentitySchema to use in GetResourceIdentitySchemas**
- **Continuing addition of functions for Resource Identity**
- **Added TODO for internal validation later**
- **Tests for GetResourceIdentity passing**
- **Initial test written for UpgradeResourceIdentity passing**
- **make ReadResource test pass**
- **use the right identity upgraders**
- **support get, set for identity data**
- **support get, set for identity data in diffs in PlanResourceChange customize diff functions**
- **Added to the internal validation in schema and corresponding tests, need to check if the expected behavior is correct**
- **Added TestUpgradeResourceIdentity_removedAttr and TestUpgradeResourceIdentity_jsonStateBigInt however BigInt doesn't work just yet**
- **safeguard empty identity sent by Terraform**
- **implement identity in ApplyResourceChange**
- **allow using Identity() when there wasn't one yet**
- **fix check for previous state**
- **Added scaffolding for internalIdentityValidate in resource.go to validate resource identities**
- **update terraform-plugin-go to latest main**
- **fix diff possibly being empty**
- **update RawIdentity to RawState as the former was retired**
- **update code comments and remove obsolete todos**


internal/configs/configschema/schema.go
```diff
@@ -95,6 +95,16 @@ type Attribute struct {
...
+
+	// RequiredForImport, if set to true, specifies that an omitted or null value is
+	// not permitted. This field conflicts with OptionalForImport.
+	// Only valid for identity schemas.
+	RequiredForImport bool
+
+	// OptionalForImport, if set to true, specifies that an omitted or null value is
+	// permitted. This field conflicts with RequiredForImport.
+	// Only valid for identity schemas.
+	OptionalForImport bool
```
austinvalle, line 107: Just a nit

```suggestion
	// RequiredForImport, if set to true, specifies that an omitted or null value is
	// not permitted when importing by the identity. This field conflicts with OptionalForImport.
	// Only valid for identity schemas.
	RequiredForImport bool

	// OptionalForImport, if set to true, specifies that an omitted or null value is
	// permitted when importing by the identity. This field conflicts with RequiredForImport.
	// Only valid for identity schemas.
	OptionalForImport bool
```

helper/schema/schema.go
```diff
@@ -846,7 +956,11 @@ func (m schemaMap) internalValidate(topSchemaMap schemaMap, attrsOnly bool) erro
 			return fmt.Errorf("%s: Cannot be both Required and Computed", k)
 		}
 
-		if !v.Required && !v.Optional && !v.Computed {
+		if v.RequiredForImport {
+
+		} else if v.OptionalForImport {
+
+		} else if !v.Required && !v.Optional && !v.Computed {
```
austinvalle, line 963: It would be nice to validate the identity schema separate from the other schemas, it would make logic like this a little cleaner.

Would this allow you to create invalid resource schemas that use `OptionalForImport` or `RequiredForImport`, and therefore have `Required`, `Optional`, and `Computed` as false?

helper/schema/resource_identity.go
```diff
@@ -0,0 +1,71 @@
...
+// upfront is recommended to prevent potential panics.
+//
+// The interface{} parameter is the result of the Provider type
+// ConfigureFunc field execution. If the Provider does not define
+// a ConfigureFunc, this will be nil. This parameter is conventionally
+// used to store API clients and other provider instance specific data.
+//
+// The map[string]interface{} return parameter should contain the upgraded
+// schema version state data for a managed resource instance. Values must
+// align to the typing mentioned above.
```
austinvalle, line 70: Just a nit that these docs should reference identity (and the correct types) rather than managed resource

helper/schema/identity_data.go
austinvalle, line 1: I'll need to eventually run through this logic with a debugger to get it straight in my head what's going on lol 😕

austinvalle (commented):
Took a first look, good stuff so far!

helper/schema/grpc_provider.go
```diff
@@ -717,6 +826,25 @@ func (s *GRPCProviderServer) ReadResource(ctx context.Context, req *tfprotov5.Re
...
+		if err != nil {
+			resp.Diagnostics = convert.AppendProtoDiag(ctx, resp.Diagnostics, err)
+			return resp, nil
+		}
+		// Step 2: Turn cty.Value into flatmap representation
+		identityAttrs := hcl2shim.FlatmapValueFromHCL2(identityVal)
+		// Step 3: Well, set it in the instanceState
+		instanceState.Identity = identityAttrs
+	} else {
+		// TODO: add diagnostic or trace in case there's no identity?
```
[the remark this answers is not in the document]
ansgarm: as far as I've observed the behaviour and if memory serves well, it does handle this gracefully

helper/schema/grpc_provider_test.go
```diff
@@ -4637,24 +5031,52 @@ func TestReadResource(t *testing.T) {
...
+								"instance_id": cty.String,
+								"region":      cty.String,
+							}),
+							cty.ObjectVal(map[string]cty.Value{
+								"instance_id": cty.StringVal("test-id"),
+								"region":      cty.StringVal("test-region"),
+							}),
+						),
+					},
+				},
```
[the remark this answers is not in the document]
ansgarm: Might be a good addition for when there's no identity to start with – in this test it's changed from `test-region` to `new-region`. But maybe that's further from reality than setting something for a previous nil value 😅

helper/schema/resource_data.go
```diff
@@ -409,6 +411,36 @@ func (d *ResourceData) State() *terraform.InstanceState {
...
+	// attribute set as a map[string]interface{}, write it to a MapFieldWriter,
+	// and then use that map.
+	if d.identitySchema != nil {
+		rawMapIdentity := make(map[string]interface{})
+		identityData, err := d.Identity()
+		// This error shouldn't happen, as we check for the identity schema first
+		if err == nil {
+			for k := range d.identitySchema {
+				raw := identityData.get([]string{k})
+				if raw.Exists && !raw.Computed {
```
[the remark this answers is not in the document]
ansgarm: Oh, yeah, it's not needed. It is copy-pasta

helper/schema/resource_identity.go
ansgarm: good catch, done!

helper/schema/identity_data.go
ansgarm: yeah, it's tough 😅 happy to walk you through the usual paths it takes :)

helper/schema/grpc_provider_test.go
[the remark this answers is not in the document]
ansgarm: Might be a good addition for when there's no identity to start with – in this test it's changed from test-region to new-region. But maybe that's further from reality than setting something for a previous nil value 😅

helper/schema/core_schema.go
```diff
@@ -370,3 +373,22 @@ func (r *Resource) CoreConfigSchema() *configschema.Block {
 func (r *Resource) coreConfigSchema() *configschema.Block {
 	return schemaMap(r.SchemaMap()).CoreConfigSchema()
 }
+
+func (r *Resource) CoreIdentitySchema() *configschema.Block {
+	block := r.coreIdentitySchema()
+
+	if block.Attributes == nil {
+		// TODO: we should error instead and callers should handle the error appropriately
+		// and error would hint at an invalid provider implementation
```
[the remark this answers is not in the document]
ansgarm: noted as a follow-up for another PR (and we also have this todo)

helper/schema/schema.go
SBGoods: Technically, this `internalValidate()` should be schema type agnostic. I don't think it's even possible to pass information about the schema type to this method without some major refactoring. If we have resource identity schema-specific validation that might need to be a separate method that's called in `(*provider).InternalValidate()`. This is where the validation that blocks write-only attributes in non-managed resource schemas is: https://github.com/hashicorp/terraform-plugin-sdk/blob/19e5b30b5502fc58d3a7008f6ca5954aaef21717/helper/schema/provider.go#L235-L238

SBGoods (commented):
Looks pretty good so far! Left a couple of comments

helper/schema/grpc_provider.go
```diff
@@ -779,6 +907,30 @@ func (s *GRPCProviderServer) ReadResource(ctx context.Context, req *tfprotov5.Re
 		MsgPack: newStateMP,
 	}
 
+	// TODO: if schema defines identity, we should error if there's none written to newInstanceState
+
+	if newInstanceState.Identity != nil {
+		identityBlock := s.getResourceIdentitySchemaBlock(req.TypeName)
```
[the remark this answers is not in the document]
ansgarm: added a todo – this would come once we add an error in the method that's eventually called inside this call. so seems to fit together

helper/schema/grpc_provider.go
```diff
@@ -717,6 +821,24 @@ func (s *GRPCProviderServer) ReadResource(ctx context.Context, req *tfprotov5.Re
...
+		if err != nil {
+			resp.Diagnostics = convert.AppendProtoDiag(ctx, resp.Diagnostics, err)
+			return resp, nil
+		}
+		// Step 2: Turn cty.Value into flatmap representation
+		identityAttrs := hcl2shim.FlatmapValueFromHCL2(identityVal)
+		// Step 3: Well, set it in the instanceState
+		instanceState.Identity = identityAttrs
+	}
+	// TODO: ELSE add diagnostic or trace in case there's no identity?
```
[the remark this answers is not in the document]
ansgarm: yeah, that makes sense. I removed the todo comments – I don't think we desperately need tracing at this point

helper/schema/schema.go
ansgarm: We'll implement validation in a follow-up PR – I'll leave this discussion open for reference.

helper/schema/grpc_provider_test.go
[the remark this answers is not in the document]
austinvalle: Ah okay, I misunderstood the test 😅

helper/schema/schema.go
```diff
@@ -714,11 +760,36 @@ func (m schemaMap) Diff(
...
+			}
+
+			mapWIdentity := &MapFieldWriter{Schema: d.identitySchema}
+			if err := mapWIdentity.WriteField(nil, rawMapIdentity); err != nil {
+				log.Printf("[ERR] Error writing identity fields: %s", err)
+				return nil, err
+			}
+
+			result.Identity = mapWIdentity.Map()
+		} // TODO: else log error?
```
austinvalle, line 792: ➕ to eventually returning errors in situations like this. This would indicate that either Terraform has a bug or the SDK has a bug, so we want to explicitly call that out.

helper/schema/identity_data.go
austinvalle, line 1: Eventually we should add some basic tests for this! 👍🏻

austinvalle (approved):
LGTM 🚀 

We should follow up this PR with a bump to the `terraform-plugin-go@v0.27.0-alpha.1` and maybe a pre-release changelog similar to the framework one -> https://github.com/hashicorp/terraform-plugin-framework/blob/main/.changes/1.15.0-alpha.1.md

helper/schema/grpc_provider.go
```diff
@@ -717,6 +821,23 @@ func (s *GRPCProviderServer) ReadResource(ctx context.Context, req *tfprotov5.Re
 	}
 	instanceState.RawState = stateVal
 
+	// TODO: is there a more elegant way to do this? this requires us to look for the identity schema block again
+	if req.CurrentIdentity != nil && req.CurrentIdentity.IdentityData != nil {
```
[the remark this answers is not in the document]
ansgarm: Oh, that comment is mostly there to indicate that this code is repeated multiple times – so extracting it into a function might make sense, but yeah, schema would need to be passed into it to work.

helper/schema/identity_data.go
ansgarm: good call! took a note ✅

PR #1447, opened 2025-03-19
Resource Identity: Add internal validation

rainkwan:
Addresses https://github.com/hashicorp/terraform-plugin-sdk/pull/1444#discussion_r1999465390 and continues #1444 by adding internal identity validation, which is mostly contained in [this commit](https://github.com/hashicorp/terraform-plugin-sdk/pull/1447/commits/253b60ef96c766decdf99f0a34ea84d151e5aa14) minus a [linting fix](https://github.com/hashicorp/terraform-plugin-sdk/pull/1447/commits/a2d383bb6ea0f24e14bfbf2ae83afe123443bf88)

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -1482,3 +1482,135 @@ func RemoveFromState(d *ResourceData, _ interface{}) error {
 	d.SetId("")
 	return nil
 }
+
+// Internal validation of provider implementation
+func (r *ResourceIdentity) InternalIdentityValidate() error {
+	if r == nil {
+		return fmt.Errorf(`The resource identity is empty`)
+	}
+
+	if r.Schema == nil {
+		return fmt.Errorf(`The resource identity schema is empty`)
+	}
+
+	for k, v := range r.Schema {
+		if !v.OptionalForImport && !v.RequiredForImport {
+			return fmt.Errorf(`OptionalForImport or RequiredForImport must be set for resource identity`)
+		}
+		if v.OptionalForImport && v.RequiredForImport {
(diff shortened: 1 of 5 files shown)


helper/schema/schema.go
austinvalle, line 1: Now that we have a separate function that will fully validate identity schemas, I think we should be able to assume anything calling this function is "not an identity schema" right?

If so, maybe we can simplify all of this logic by just checking if either of these new fields are set, then returning an error:
```go
if v.RequiredForImport {
  return fmt.Errorf("%s: RequiredForImport is only valid for resource identity schemas", k)
}
if v.OptionalForImport {
  return fmt.Errorf("%s: OptionalForImport is only valid for resource identity schemas", k)
}
```

austinvalle (commented):
This is looking good! I have one specific question about the original schema validate logic

helper/schema/resource.go
```diff
@@ -1482,3 +1482,135 @@ func RemoveFromState(d *ResourceData, _ interface{}) error {
...
+
+// Internal validation of provider implementation
+func (r *ResourceIdentity) InternalIdentityValidate() error {
+	if r == nil {
+		return fmt.Errorf(`The resource identity is empty`)
+	}
+
+	if r.Schema == nil {
+		return fmt.Errorf(`The resource identity schema is empty`)
+	}
```
ansgarm, line 1494: I think we should also check whether the length of the Schema is zero and return an error that at least one attribute is required. What do you think?

ansgarm (dismissed):

ansgarm (approved):

PR #1450, opened 2025-03-19
Prepare 2.37.0-alpha.1 release

ansgarm:
~Updated the file as per the `CONTRIBUTING.md` guide (at the very bottom)~

Adds support for pre-releases to Github Action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,9 +20,26 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.changelog-version.outputs.version }}
+      version_only: ${{ steps.changelog-version.outputs.version_only }}
+      prerelease: ${{ steps.changelog-version.outputs.prerelease }}
     steps:
       - id: changelog-version
-        run: echo "version=$(echo "${{ inputs.versionNumber }}" | cut -c 2-)" >> "$GITHUB_OUTPUT"
+        run: |
+          version="${{ inputs.versionNumber }}"
+          version="${version#v}"  # Remove leading "v" if present
+          version_only="${version%%-*}"
+          prerelease="${version#*-}"
+
+          # If there's no dash, set prerelease to empty
+          if [ "$version" = "$version_only" ]; then
+            prerelease=""
+          fi
+
+          {
+            echo "version=$version"
+            echo "version_only=$version_only"
+            echo "prerelease=$prerelease"
+          } >> "$GITHUB_OUTPUT"
 
   changelog:
     needs: [ changelog-version, meta-version ]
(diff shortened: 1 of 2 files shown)


meta/meta.go
```diff
@@ -17,15 +17,15 @@ import (
 //
 // Deprecated: Use Go standard library [runtime/debug] package build information
 // instead.
-var SDKVersion = "2.36.1"
+var SDKVersion = "2.37.0"
```
austinvalle, line 20: I believe our CI should do this for us? Not sure about the pre-release portion: https://github.com/hashicorp/terraform-plugin-sdk/commit/9a1ad1dbc68125df2491ac25ac1447d21fdf70d2

ansgarm: Oh, neat, I did not know!

austinvalle: Yeah it looks like it probably won't update pre-release 😆 , not sure exactly who is using this and where, might not be a big deal ATM

https://github.com/hashicorp/terraform-plugin-sdk/blob/9eb7da0be3e9a35771af84fae0a37a53d96fb833/.github/workflows/release.yml#L59-L81

ansgarm: I'll update the contributing docs then once we know what'll happen with the pre-release

ansgarm: not entirely sure what'll end up in `changelog-version.outputs.version` but I highly suspect the pre-release identifier might be part of it 🤨

ansgarm: Oh damn, Github didn't show your comment 😂 So it now seems like I ignored what you said..

bbasata: Here's one data point -- a search across official provider repos. Two providers use this in a user-agent string, which seems fine.
```
$ ag meta\.SDKVersion $(ag -l SDKVersion | grep -v vendor)
terraform-provider-azuread/internal/common/client_options.go
109:	tfUserAgent := fmt.Sprintf("HashiCorp Terraform/%s (+https://www.terraform.io) Terraform Plugin SDK/%s", o.TerraformVersion, meta.SDKVersionString()) //nolint:staticcheck

terraform-provider-azurestack/internal/common/client_options.go
62:	tfUserAgent := fmt.Sprintf("HashiCorp Terraform/%s (+https://www.terraform.io) Terraform Plugin SDK/%s", tfVersion, meta.SDKVersionString())
```

ansgarm: Added a commit that fixes the action nevertheless – who knows what else is happening out there 😅

bbasata: I like it! I had not seen the existing `var SDKPrerelease`.

.github/workflows/release.yml
```diff
@@ -70,8 +87,10 @@ jobs:
           # Avoid persisting GITHUB_TOKEN credentials as they take priority over our service account PAT for `git push` operations
           # More details: https://github.com/actions/checkout/blob/b4626ce19ce1106186ddf9bb20e706842f11a7c3/adrs/0153-checkout-v2.md#persist-credentials
           persist-credentials: false
-      - name: Update meta package SDKVersion
-        run: sed -i "s/var SDKVersion =.*/var SDKVersion = \"${{ needs.changelog-version.outputs.version }}\"/" meta/meta.go
+      - name: Update meta package SDKVersion and SDKPrerelease
+        run: |
+          sed -i "s/var SDKVersion =.*/var SDKVersion = \"${{ needs.changelog-version.outputs.version_only }}\"/" meta/meta.go
+          sed -i "s/var SDKPrerelease =.*/var SDKPrerelease = \"${{ needs.changelog-version.outputs.prerelease }}\"/" meta/meta.go
```
austinvalle, line 93: Honestly.... I'm wondering if we should just include the full version in `SDKVersion` and then the `SDKPrerelease` as the separated portion..

Because the `SDKVersion` shouldn't be `v2.37.0` when we release this, it should really be `v2.37.0-alpha.1`. I don't think `SDKPrerelease` is super useful given that context, but since it exists, probably best to keep it up to date 😆 

So when we release I think it'd make sense to be:
```go
var SDKVersion = "v2.37.0-alpha.1"
var SDKPrerelease = "alpha.1"
```

Rather than:
```go
var SDKVersion = "v2.37.0"
var SDKPrerelease = "alpha.1"
```

austinvalle: I guess also, all of this is deprecated anyways, so it doesn't really matter 🤷🏻

bbasata: I'm okay with prioritizing `SDKVersionString()` correctness. This is consistent with the doc comment for `SDKVersion`: `The main version number that is being run at the moment.`

bbasata (approved):

ansgarm:
> could we use `go generate`

I'd love such a PR!

PR #1451, opened 2025-03-19
chore: Update alpha version of TF mentioned in changelog

austinvalle:
It's too late for the other alphas, but Terraform just released this new alpha, which has a single bug fix for ReadResource / the JSON command: https://github.com/hashicorp/terraform/releases/tag/v1.12.0-alpha20250319


ansgarm (approved):
nice

PR #1452, opened 2025-03-20
error if there's no identity schema (or an empty one) defined when it should be

ansgarm:
Return error diagnostics if there's no identity schema (or an empty one) defined when it should be.

Includes tests for this (`GetResourceIdentitySchemas`, `ReadResource`, `PlanResourceChange`, `ApplyResourceChange`)

Oh, and also adds an identity test for `ApplyResourceChange` that didn't exist yet.


rainkwan (approved):
Seems sensible

austinvalle (approved):
Nice!

PR #1458, opened 2025-03-26
TFECO-9158: Add tests for ResourceData#Identity method

ansgarm:
~depends on #1453~

### DIFF
--- a/helper/schema/resource_data_test.go
+++ b/helper/schema/resource_data_test.go
@@ -4178,6 +4178,142 @@ func TestResourceDataSetType(t *testing.T) {
 	}
 }
 
+func TestResourceDataIdentity(t *testing.T) {
+	d := &ResourceData{
+		identitySchema: map[string]*Schema{
+			"foo": {
+				Type:              TypeString,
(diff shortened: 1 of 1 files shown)


helper/schema/resource_data_test.go
```diff
@@ -4178,6 +4178,142 @@ func TestResourceDataSetType(t *testing.T) {
...
+				"foo": "baz",
+			},
+		},
+	}
+	d.SetId("bar") // just required to be able to call .State()
+	identity, err := d.Identity()
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	err = identity.Set("foo", "qux")
```
austinvalle, line 4292: TIL, the following word after "foo", "bar", "baz" 😆

austinvalle (approved):
Nice tests! 🚀

helper/schema/resource_data_test.go
ansgarm: I always have to google it 😂 I only retain that it exists but not which word it was 😆

PR #1460, opened 2025-03-27
TFECO-9206 Add validation to check whether the identity that we're about to send back to Terraform is complete

ansgarm:
based on #1459

### DIFF
--- a/helper/schema/grpc_provider.go
+++ b/helper/schema/grpc_provider.go
@@ -929,6 +929,15 @@ func (s *GRPCProviderServer) ReadResource(ctx context.Context, req *tfprotov5.Re
 			return resp, nil
 		}
 
+		// validate every attribute exists and is not null
+		identityAttrs := newIdentityVal.AsValueMap()
+		for k := range identityBlock.Attributes {
+			if v, ok := identityAttrs[k]; !ok || v.IsNull() || !v.IsWhollyKnown() {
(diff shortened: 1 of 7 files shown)


helper/schema/grpc_provider.go
```diff
@@ -929,6 +929,15 @@ func (s *GRPCProviderServer) ReadResource(ctx context.Context, req *tfprotov5.Re
...
 		}
 
+		// validate every attribute exists and is not null
+		identityAttrs := newIdentityVal.AsValueMap()
+		for k := range identityBlock.Attributes {
+			if v, ok := identityAttrs[k]; !ok || v.IsNull() || !v.IsWhollyKnown() {
+				resp.Diagnostics = convert.AppendProtoDiag(ctx, resp.Diagnostics, fmt.Errorf("identity data missing for attribute '%s'. This is a bug in the provider, which should be reported in the provider's own issue tracker.", k))
+				return resp, nil
+			}
+		}
```
austinvalle, line 939: Wondering if we should duplicate this validation in the provider? We know core needs/will have this logic at some point and we don't typically do validation like this in the SDKs. (For example, ReadResource it's invalid to return a `NewState` that has unknown values, which core validates, but the SDKs do not)

----------

If we do decide that we want to validate this in the providers, we might consider doing it at the `terraform-plugin-go` level, so it applies to `terraform-plugin-framework` as well

ansgarm: Good call!

austinvalle:
@ansgarm I think we can close this one? Might have been spun off of an early idea we had 🤔

ansgarm:
Oh, yeah! It's obsolete! Good catch

PR #1474, opened 2025-04-22
TFECO-9286: add `ImportStatePassthroughWithIdentity`

ansgarm:
This PR adds support for `ImportStatePassthroughWithIdentity` which makes it easier to implement support for import by identity if the fallback `Id` attribute is part of the identity object (e.g. if it's a UUID or ARN that can just be copied as is from an attribute of the identity object into the Id string).

### DIFF
--- a/helper/schema/resource_importer.go
+++ b/helper/schema/resource_importer.go
@@ -6,6 +6,7 @@ package schema
 import (
 	"context"
 	"errors"
+	"fmt"
 )
 
 // ResourceImporter defines how a resource is imported in Terraform. This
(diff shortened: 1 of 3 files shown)


helper/schema/resource_importer.go
```diff
@@ -83,3 +84,30 @@ func ImportStatePassthrough(d *ResourceData, m interface{}) ([]*ResourceData, er
 func ImportStatePassthroughContext(ctx context.Context, d *ResourceData, m interface{}) ([]*ResourceData, error) {
 	return []*ResourceData{d}, nil
 }
+
+func ImportStatePassthroughWithIdentity(idAttributePath string) StateContextFunc {
```
austinvalle, line 88: Since this is an exported function, we should add some general docs here (use when you need to support identity AND id-only refresh).

Since identity attributes are "flat", the only real accepted path for `idAttributePath` here should be an attribute name, but there's no real way to enforce it, just document 😄. It also must be a string identity attribute, which might be worth noting.

helper/schema/resource_importer_test.go
```diff
@@ -14,3 +18,168 @@ func TestInternalValidate(t *testing.T) {
...
+		},
+	}
+
+	tests := []struct {
+		name                 string
+		idAttributePath      string
+		resourceData         *ResourceData
+		expectedResourceData *ResourceData
+		expectedError        string
+	}{
```
austinvalle, line 41: We should also write a failing test case where the identity attribute path does exist, but it's not a string 🙂

austinvalle (commented):
Looking good! I have a suggestion for documentation and one additional test case

helper/schema/resource_importer.go
ansgarm: Oh, of course – totally missed those docs 😅

ansgarm: Oh, of course – totally missed those docs 😅

austinvalle (approved):
🚀

PR #1476, opened 2025-04-24
Add changie entry for write-only update method validation bugfix

SBGoods:
Adds changie entry for  #1472


austinvalle (approved):
Good catch! 🚀

PR #1499, opened 2025-06-26
Fix typo in diagnostic summary for identity change error message

NateBrady23:
- Corrected the word "then" to "than" in the error message displayed during the read operation in TestReadResource.

Related: https://github.com/hashicorp/terraform-plugin-framework/pull/1172


austinvalle (changes requested):
Thanks for the PR @NateBrady23!

Looks like there is one test that needs to be updated, but the change itself looks good 👍🏻

PR #1504, opened 2025-08-07
helper/schema: Export `tfprotov5` schema and identity schema for consumption by other Go modules

austinvalle:
## Related Issue

N/A

## Description

This PR introduces new methods on `schema.Resource` that return the protocol v5 schema and identity schema. This can be useful for other Go modules that need the schema to decode/encode data for a managed resource. The logic is the same as what is returned from `GetProviderSchema` and `GetResourceIdentitySchemas` RPCs:
- https://github.com/hashicorp/terraform-plugin-sdk/blob/8984e1e62a17bd607f069101a760e23a649ac2f8/helper/schema/grpc_provider.go#L241-L244
- https://github.com/hashicorp/terraform-plugin-sdk/blob/8984e1e62a17bd607f069101a760e23a649ac2f8/helper/schema/grpc_provider.go#L92-L104

Both methods return a function, which allows consuming Go modules to defer retrieving the schema until needed.

## Rollback Plan

No

### DIFF
--- a/helper/schema/core_schema.go
+++ b/helper/schema/core_schema.go
@@ -288,13 +288,6 @@ func (s *Schema) coreConfigSchemaType() cty.Type {
 	}
 }
 
-func (r *Resource) ProtoSchema(ctx context.Context) *tfprotov5.Schema {
-	return &tfprotov5.Schema{
-		Version: int64(r.SchemaVersion),
-		Block:   convert.ConfigSchemaToProto(ctx, r.CoreConfigSchema()),
-	}
-}
-
 // CoreConfigSchema is a convenient shortcut for calling CoreConfigSchema on
 // the resource's schema. CoreConfigSchema adds the implicitly required "id"
 // attribute for top level resources if it doesn't exist.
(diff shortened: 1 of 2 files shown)


helper/schema/core_schema_test.go
```diff
@@ -34,15 +34,6 @@ func testResource(block *configschema.Block) *configschema.Block {
...
 
 func TestSchemaMapCoreConfigSchema(t *testing.T) {
-	// these are global so if new tests are written we should probably employ a mutex
-	DescriptionKind = StringMarkdown
-	SchemaDescriptionBuilder = func(s *Schema) string {
-		if s.Required && s.Description != "" {
-			return fmt.Sprintf("**Required** %s", s.Description)
-		}
-		return s.Description
-	}
```
austinvalle, line 44: This is technically reducing the coverage of this test, but it's so specific that I don't think we're losing too much 😆

stephybun (approved):
This all looks reasonable to me, LGTM 🦒

PR #1508, opened 2025-08-19
list: add conversion functions for cty values to tftypes values

stephybun:
## Description

 - Adds a new public package with functions that convert a cty value to a tftypes value
 - Adds two receiver functions on `ResourceData` that convert resource identity or resource data in cty to tftypes

TODO:
- Additional tests
- Better error handling.. I'm discarding errors in certain places 😅 

## Rollback Plan

None


helper/convert/type.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// An private version of this function to convert cty.Type to tftypes.Type exists in
+// internal/plugin/convert/schema.go but this is a simplified and public version
+// that maps the types one to one without any error handling, under the assumption
+// the incoming data will always be valid
+func ToTfType(in cty.Type) tftypes.Type {
+	switch {
+	case in.IsPrimitiveType():
```
austinvalle, line 17: For completeness, the final primitive that we'd need to add here is `DynamicPseudoType`, which should never actually show up for the use-cases we have, but it's a pretty simple add

```go
	case in.IsPrimitiveType():
		if in == cty.DynamicPseudoType {
			return tftypes.DynamicPseudoType
		}
```

helper/convert/type.go
```diff
@@ -0,0 +1,54 @@
...
+	case in.IsTupleType():
+		elemTypes := make([]tftypes.Type, 0)
+
+		for _, v := range in.TupleElementTypes() {
+			elemTypes = append(elemTypes, ToTfType(v))
+		}
+
+		return tftypes.Tuple{ElementTypes: elemTypes}
+	}
+	return nil
```
austinvalle, line 53: I feel like if we should either panic (something like "unknown cty.Type <print type>"), or return `DynamicPseudoType`, which is typically a fill-in for "we don't know the type"

I'm leaning towards panic being the right behavior. In the very unlikely chance that cty adds a new type that we don't add support for here, it'll be found immediately with a clear message vs. causing confusing errors downstream.

helper/convert/type.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// An private version of this function to convert cty.Type to tftypes.Type exists in
+// internal/plugin/convert/schema.go but this is a simplified and public version
+// that maps the types one to one without any error handling, under the assumption
+// the incoming data will always be valid
```
austinvalle, line 14: Does this function need to be duplicated and public? If it's only being consumed by `ResourceData` then perhaps we could just re-use that internal one and remove this?

Or if we need this exported version we could call the internal one and ignore/panic on the error?

helper/convert/value.go
```diff
@@ -0,0 +1,131 @@
...
+	case ty.IsSetType():
+		return SetTfValue(in)
+	case ty.IsTupleType():
+		return TupleTfValue(in)
+	}
+
+	return emptyTfValue(ToTfType(in.Type()))
+}
+
+func emptyTfValue(ty tftypes.Type) tftypes.Value {
```
austinvalle, line 129: nit: Just a general terminology thing, I'd probably call this `nullTfValue`

helper/convert/value.go
```diff
@@ -0,0 +1,131 @@
...
+	if in.IsNull() {
+		return emptyTfValue(ToTfType(in.Type()))
+	}
+
+	var val tftypes.Value
+	switch in.Type() {
+	case cty.String:
+		val = tftypes.NewValue(tftypes.String, in.AsString())
+	case cty.Bool:
+		val = tftypes.NewValue(tftypes.Bool, in.True())
```
austinvalle, line 21: This looks like it'll always return true, rather than whatever the boolean value is

helper/convert/value.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+func PrimitiveTfValue(in cty.Value) tftypes.Value {
+	if in.IsNull() {
+		return emptyTfValue(ToTfType(in.Type()))
+	}
+
+	var val tftypes.Value
+	switch in.Type() {
```
austinvalle, line 17: I'm not exactly sure what the best way to handle this scenario, but what if the type is `cty.DynamicPseudoType`.

I'm fairly certain the only way that can be possible is if `in.IsUnknown()` or `in.IsNull()` are true, so perhaps a panic would be fine? 🤔

helper/convert/value.go
```diff
@@ -0,0 +1,131 @@
...
+	vals := make([]tftypes.Value, 0)
+
+	for _, v := range in.AsValueSlice() {
+		vals = append(vals, ToTfValue(v))
+	}
+
+	return tftypes.NewValue(tupleType, vals)
+}
+
+func ToTfValue(in cty.Value) tftypes.Value {
```
austinvalle, line 109: Similar question to `ToTfType`, do these functions needs to be exported since they are all accessed via `ResourceData`? If not, it might make sense to just move this to the internal `plugin/convert` package.

If they do need to be exported, we should add some package docs 🙂

helper/convert/value.go
```diff
@@ -0,0 +1,131 @@
...
+	ty := in.Type()
+	switch {
```
austinvalle, line 111: Similar note about switch statements here, if we want to keep the clean API (not returning an error) I think that's fine, but we should probably favor a default "panic" statement if none of the statements are true.

It would be really difficult for us to track a bug down if it's a value being ignored and replaced by a null.

helper/convert/value_test.go
```diff
@@ -0,0 +1,420 @@
...
+			Value: cty.NumberFloatVal(3.14),
+			Want:  tftypes.NewValue(tftypes.Number, 3.14),
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.Value.GoString(), func(t *testing.T) {
+			t.Parallel()
+
+			got := PrimitiveTfValue(test.Value)
```
austinvalle, line 43: nit: I think for tests like this, you can probably just call the top-level `ToTfValue` since it's how all consumers would call it anyways, plus there isn't a ton of setup needed to call that function anyways.

helper/convert/value_test.go
```diff
@@ -0,0 +1,420 @@
...
+			got := ListTfValue(test.Value)
+
+			if diff := cmp.Diff(test.Want, got); diff != "" {
+				t.Errorf("unexpected differences: %s", diff)
+			}
+		})
+	}
+}
+
+func TestSetTfType(t *testing.T) {
```
austinvalle, line 187: > Same comment for the map + tuple tests

nit: I think it'd be nice to do what you did for the list test, which is add a test that goes through at least one level of recursion (a set of objects with a set attribute)

helper/schema/resource_data.go
```diff
@@ -59,6 +63,93 @@ type getResult struct {
 	Schema         *Schema
 }
 
+// TfTypeIdentity returns the identity data as a tftypes.Value.
```
austinvalle, line 66: We should add some tests for these 🙂 , same comment for the other method.

It doesn't need to be as robust as the tests directly against `ToTfValue`, but at least testing the additional logic in this method (like adding attributes, or the flatmapping of data)

helper/schema/resource_data.go
```diff
@@ -59,6 +63,93 @@ type getResult struct {
...
+
+	return convert.ToTfValue(stateVal)
+}
+
+// TfTypeResource returns the resource data as a tftypes.Value.
+func (d *ResourceData) TfTypeResource() tftypes.Value {
+	s := schemaMap(d.schema).CoreConfigSchema()
+
+	// The CoreConfigSchema method on schemaMaps doesn't automatically handle adding the id
+	// attribute or timeouts like the method on Resource does
```
austinvalle, line 85: Nice comment 🔥

helper/convert/value.go
stephybun: It was intentional to add these as public functions since this type of `cty` -> `tftypes` conversion doesn't really exist anywhere else (at least not publicly) but since we want to shift away from extending SDKv2 functionality I felt like this could be a useful utility/puzzle piece for provider developers to have access to if they're wanting to mix and match things from SDKv2 and Framework (and even plugin-go?) even though the use cases might be quite rare. 

There's obviously quite a lot of speculation in that view though, perhaps this is something that we should hide away and only expose in a controlled manner. I'm not precious about it in any case.

WDYT?

helper/convert/value.go
stephybun: I originally thought so too, but 

```
// True returns true if the receiver is True, false if False, and panics if
// the receiver is not of type Bool.
```

helper/convert/value_test.go
stephybun: Totally forgot maps 🤦‍♀️

helper/schema/resource_data.go
```diff
@@ -59,6 +63,101 @@ type getResult struct {
...
 }
 
+// TfTypeIdentityState returns the identity data as a tftypes.Value.
+func (d *ResourceData) TfTypeIdentityState() tftypes.Value {
+	s := schemaMap(d.identitySchema).CoreConfigSchema()
+
+	state := d.State()
+
+	if state == nil {
+		log.Panicf("state is nil")
```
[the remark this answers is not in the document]
austinvalle: Interesting, I'm not fully understanding the scenario you're describing, but it does sound like if there are potential setup mishaps, maybe we should be returning an error from this function? 🤔

helper/convert/value.go
austinvalle: Lolol, what an odd name, I'm sure there's a fun history lesson on why that is named that way 😆

internal/plugin/convert/value_test.go
```diff
@@ -0,0 +1,613 @@
...
+					AttributeTypes: map[string]tftypes.Type{
+						"fruits": tftypes.Map{ElementType: tftypes.String},
+					}}}, map[string]tftypes.Value{
+				"foo": tftypes.NewValue(tftypes.Object{
+					AttributeTypes: map[string]tftypes.Type{
+						"fruits": tftypes.Map{ElementType: tftypes.String},
+					}}, map[string]tftypes.Value{
+					"fruits": tftypes.NewValue(tftypes.Map{ElementType: tftypes.String}, map[string]tftypes.Value{
+						"ananas":   tftypes.NewValue(tftypes.String, "pineapple"),
+						"erdbeere": tftypes.NewValue(tftypes.String, "strawberry"),
```
rainkwan, line 369: Fun lesson in German

Issue #1502, opened 2025-07-11
SDKv2 allows resource identities to be set to null, eventually triggering mutable identity error


austinvalle:
### Module version
```
v2.37.0
```

### Context
The AWS provider recently found a bug in `v6.0.0` with a resource identity implementation where the identity was not being set, which resulted in the identity object being stored in Terraform state as all `null` attributes:
```jsonc
{
    "module": "module.simple-s3-cdn",
      "mode": "managed",
      "type": "aws_s3_bucket",
      "name": "origin",
      "provider": "provider[\"registry.terraform.io/hashicorp/aws\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            // ...
          },
          "sensitive_attributes": [],
          "identity_schema_version": 0,
          "identity": { // <---------------- all identity attributes are null
            "account_id": null,
            "bucket": null,
            "region": null
          }
        }
      ]
    }
```

As a result of this bug, the AWS provider attempted to fix the identities in `v6.1.0`, however because the identity had already been stored (as an invalid null object), they received the mutable identity error message: https://github.com/hashicorp/terraform-provider-aws/issues/43199

```bash
Error: Unexpected Identity Change: During the read operation,
  the Terraform Provider unexpectedly returned a different identity
  then the previously stored one.
```

### Expected Behavior

An identity that is `null` or has all `null` attributes is invalid, so the SDK should have returned an error to the provider indicating that identity was supported, however the provider did not return an identity. This would have helped the provider developers catch this bug earlier, before identity was stored in any state files.

The only time an identity is valid to be null is in a read operation immediately following import, where the identity might be null because the practitioner is importing via the ID string.

### References
- https://github.com/hashicorp/terraform-plugin-framework/issues/1184
- https://github.com/hashicorp/terraform-provider-aws/issues/43199
- https://github.com/hashicorp/terraform-provider-aws/issues/43204

PR #1513, opened 2025-08-27
identity: add validation to check if a resource identity is null 

stephybun:
## Related Issue

#1502

## Description

Adds validation to the ApplyResource and ReadResource RPC to check whether a resource identity is fully null i.e. all attributes within the identity are null.

## Rollback Plan

None


helper/schema/resource.go
```diff
@@ -679,6 +679,11 @@ type ResourceBehavior struct {
 	// resource's lifecycle. Setting this flag to true will disable the SDK validation that ensures identity
 	// data doesn't change during RPC calls.
 	MutableIdentity bool
+
+	// AllowNullIdentity toggles whether the managed resource allows identities to be null. Setting this flag to true
+	// disables the SDK validation that ensures the identity cannot be null at the end of ApplyResourceChange and ReadResource
+	// RPC calls in certain situations.
+	AllowNullIdentity bool
```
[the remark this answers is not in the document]
austinvalle: Continuing our offline conversation, I think I'm on the side of removing this flag and just always applying it. In lieu of a real example of a resource that has an identity that is allowed to be fully null (while still supporting identity), I think any bugs we introduce would require a re-release of the provider anyways, so we could always come back and add a flag if we find a scenario or adjust the validation logic

helper/schema/grpc_provider.go
```diff
@@ -1544,6 +1563,24 @@ func (s *GRPCProviderServer) ApplyResourceChange(ctx context.Context, req *tfpro
 			return resp, nil
 		}
 
+		if !res.ResourceBehavior.AllowNullIdentity {
+			isFullyNull := true
+			for _, v := range newIdentityVal.AsValueMap() {
```
austinvalle, line 1568: ❓ - I'm not 100% sure what the value of `newIdentityVal` will be during a destroy operation (which still calls `ApplyResourceChange`). I'm assuming because of the flatmap this will still just be an object, but could potentially have all null values?

Perhaps there is no logic in providers or the SDKs that actually nulls the values during destroy 🤔 -> I believe state will be nulled out before returning from `ApplyResourceChange` , so this must be different

helper/schema/grpc_provider.go
```diff
@@ -1544,6 +1563,24 @@ func (s *GRPCProviderServer) ApplyResourceChange(ctx context.Context, req *tfpro
...
+					isFullyNull = false
+					break
+				}
+			}
+
+			if isFullyNull {
+				resp.Diagnostics = convert.AppendProtoDiag(ctx, resp.Diagnostics, fmt.Errorf(
+					"Missing Resource Identity After Create: The Terraform provider unexpectedly returned no resource identity after having no errors in the resource create. "+
+						"This is always a problem with the provider and should be reported to the provider developer",
+				))
```
austinvalle, line 1579: nit on message: This isn't an error that can only happen during create right? Technically this logic is run during any call to `ApplyResoureChange`, which could be a `create`, `update`, or `delete/destroy`

helper/schema/grpc_provider.go
```diff
@@ -970,6 +971,24 @@ func (s *GRPCProviderServer) ReadResource(ctx context.Context, req *tfprotov5.Re
 			return resp, nil
 		}
 
+		if !res.ResourceBehavior.AllowNullIdentity {
```
austinvalle, line 974: nit: In read this check happens after the mutable identity check, in apply it happens before. Should they be consistent (either way)

austinvalle (commented):
Left some comments! Also one suggestion to perhaps increase our confidence in this change.

Perhaps we can reach out to the AWS team to run these changes (perhaps after we merge to main, before releasing the module) against their acceptance test suite? They have the most resource identity implementations and have a lot of tests that might help us see if there are any edges we missed.

helper/schema/grpc_provider.go
stephybun: If the operation is a destroy `ApplyResourceChange` will return early and we'll never reach this part of the code

https://github.com/hashicorp/terraform-plugin-sdk/blob/b3515012a37a0648c7cfa2ecdfe50d3ead5fd99b/helper/schema/grpc_provider.go#L1508-L1521

PR #1582, opened 2026-04-23
helper/schema: preserve no-op planning shortcut when only identity is carried over

ansgarm:
_Assisted-by: Claude opus-4.7 (via OpenCode)_

## Summary

- Fix a regression introduced in #1444 where adopting Resource Identity caused SDKv2 providers to emit spurious plan diffs (e.g. `- timeouts {}`) on no-op updates.
- `schemaMapWithIdentity.Diff` always copies the prior identity into `diff.Identity`, so a non-empty `diff.Identity` is not on its own evidence of an identity change. The previous `len(diff.Identity) == 0` guard in `PlanResourceChange` therefore caused the no-op shortcut to be skipped for every resource that adopted `Identity`, forcing the planned state to be re-built from the legacy flatmap diff. That round-trip drops state shapes the flatmap path can't faithfully reproduce — most visibly an empty `timeouts {}` block left behind by older provider versions.
- Compare `diff.Identity` against `priorState.Identity` instead, so the shortcut still fires when identity is merely carried over unchanged.

Reported in hashicorp/terraform-provider-google#27055.

## Test plan

- New regression test `TestPlanResourceChange/identity-noop-preserves-empty-timeouts-block` exercises the timeouts/identity scenario; verified failing on `main` and passing with the fix.
- `go test ./...` passes across the module.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20260423-110543.yaml
+++ b/.changes/unreleased/BUG FIXES-20260423-110543.yaml
@@ -1,5 +1,5 @@
 kind: BUG FIXES
-body: 'helper/schema: Fix `PlanResourceChange` skipping the no-op shortcut when a resource has `Identity` set, which caused legacy state shapes (e.g. an empty `timeouts {}` block) to be re-shaped via the flatmap path and produce spurious diffs.'
+body: 'helper/schema: Fixed a bug that caused spurious diffs in the plan when resource identity is set but didn''t change.'
 time: 2026-04-23T11:05:43.475731+02:00
 custom:
     Issue: "1582"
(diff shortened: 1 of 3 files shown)


ansgarm:
Thinks I (as a human) did to further test this change:
* ran the unit test suite of the Google Terraform provider locally with a replace directive pointing at this branch (all tests passed)
* created a fake local tfstate that had `"timeouts": {"create": null, "update": null, "delete": null}` and ran the Doogle Terraform provider compiled locally with and without this SDKv2 fix against it. I was able to reproduce the diff with `- timeouts {}` with the non-patched SDKv2 and saw it gone (aka "no changes") after replacing the SDKv2 dependency with this branch.

.changes/unreleased/BUG FIXES-20260423-110543.yaml
```diff
@@ -0,0 +1,5 @@
+kind: BUG FIXES
+body: 'helper/schema: Fix `PlanResourceChange` skipping the no-op shortcut when a resource has `Identity` set, which caused legacy state shapes (e.g. an empty `timeouts {}` block) to be re-shaped via the flatmap path and produce spurious diffs.'
```
SBGoods: I think this changelog entry is a bit verbose and goes too much into implementation details. Maybe something along the lines of "Fixed a bug that caused spurious diffs in the plan when resource identity is set." The more specific we can be about the nature of the bug from the user's perspective, the better.

ansgarm: Thanks! Changed it 👍

SBGoods (approved):

PR #1583, opened 2026-04-24
Backport #1582 and 1581 to v2.40.x

ansgarm:
We already merged changes to `main` that we don't want to release alongside this patch yet, this is why we release from a backport branch even though `2.40.1` would be the natural successor of the current `2.40.0` release.

This PR mainly backports the fix for the bug that caused spurious diffs in the plan when resource identity was set but didn't change. For good measure, it also includes the dependency updates related to `hc-install` failing for expired old gpg keys when downloading Terraform (can happen during testing when there's no locally installed Terraform version).

This PR includes:
- #1582
- #1581
- #1579

After this PR is approved and merged, we'll run the release Github action from `release/2.40.x` for the version `2.40.1`.


SBGoods (dismissed):

SBGoods:
Actually, while we're at it, can we also backport #1567 to resolve #1570?

ansgarm:
pushed!

SBGoods (approved):

PR #1598, opened 2026-07-06
helper/retry: omit last error from UnexpectedStateError message when nil

### LINKED ISSUE #1359: Don't print `UnexpectedStateError.LastError` when it's nil
### SDK version

```
...
github.com/hashicorp/terraform-plugin-sdk/v2 v2.34.0
```

### Use-cases

In case when the retry func faces an unexpected resource status it returns `unexpected state 'ACTIVE', wanted target 'DELETED'. last error: %!s(<nil>)`.

### Proposal

There should be more cleaner error report, when LastError is nil: `unexpected state 'ACTIVE', wanted target 'DELETED'`, e.g.

```go
func (e *UnexpectedStateError) Error() string {
        if e.LastError == nil {
                return fmt.Sprintf( 
                        "unexpected state '%s', wanted target '%s'.",
                        e.State,
                        strings.Join(e.ExpectedState, ", "),
                )
        }
        return fmt.Sprintf(
                "unexpected state '%s', wanted target '%s'. last error: %s",
                e.State,
                strings.Join(e.ExpectedState, ", "),
                e.LastError,
        )
}```

<!---
If you have an idea for a way to address the problem via a change to SDK features, please describe it below.

In this section, it's helpful to include specific examples of how what you are suggesting might look in configuration files, or on the command line, since that allows us to understand the full picture of what you are proposing.

If you're not sure of some details, don't worry! When we evaluate the feature request we may suggest modifications as necessary to work within the design constraints of the SDK and Terraform Core.
-->

### References
<!--
Are there any other GitHub issues, whether open or closed, that are related to the problem you've described above or to the suggested solution? If so, please create a list below that mentions each of them. For example:

- #6017
-->

https://github.com/hashicorp/terraform-plugin-sdk/blob/4c604df5f03572d01e7c32777b339519213eb475/helper/retry/error.go#L45

emmayusufu:
`UnexpectedStateError.Error()` formats `LastError` with `%s` unconditionally, so when the retry gave up without an underlying error the message ends in `last error: %!s(<nil>)`.

This only includes the suffix when a last error actually exists, the same way `TimeoutError.Error()` right below it already handles its nil case. Added a small table test covering both paths (the nil case reproduces the `%!s(<nil>)` output before the fix), plus a changie entry.

Closes #1359

### DIFF
--- a/.changes/unreleased/BUG FIXES-20260706-152904.yaml
+++ b/.changes/unreleased/BUG FIXES-20260706-152904.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'helper/retry: Omitted the last error suffix from the UnexpectedStateError message when there is no last error, instead of printing a nil value'
+time: 2026-07-06T15:29:04+03:00
+custom:
+    Issue: "1359"
--- a/helper/retry/error.go
+++ b/helper/retry/error.go
@@ -41,11 +41,19 @@ type UnexpectedStateError struct {
 }
 
 func (e *UnexpectedStateError) Error() string {
+	if e.LastError != nil {
+		return fmt.Sprintf(
+			"unexpected state '%s', wanted target '%s'. last error: %s",
+			e.State,
+			strings.Join(e.ExpectedState, ", "),
+			e.LastError,
+		)
+	}
+
 	return fmt.Sprintf(
-		"unexpected state '%s', wanted target '%s'. last error: %s",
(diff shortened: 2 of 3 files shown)


ansgarm (approved):
Hi @emmayusufu 👋

Thank you for your (first!) contribution 💯 This looks good to me 👍

emmayusufu:
Thanks for the quick review!

PR #1609, opened 2026-08-20
helper/schema: apply provider and practioner set timeouts during ReadDataApply

### LINKED ISSUE #1038: Timeouts are not applied to a Data Source read during apply

sreallymatt:
## Related Issue

Fixes #1038

## Description

Reads timeouts from `(*terraform.InstanceDiff).Meta` in `ReadDataApply` so that provider, and practitioner, defined timeouts apply rather than defaulting to the SDK's 20 minutes. This matches the logic in the resource's `Apply` func

Within the AzureRM and AzureAD providers, the majority of our data source `Read` timeouts are set to 5 minutes, 20 minutes is excessive for data source reads in our case and this allows the specified timeouts to apply.

CC: @katbyte / @jackofallops 

## Rollback Plan

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.


helper/schema/resource.go
```diff
@@ -905,7 +905,8 @@ func (r *Resource) Apply(
 	ctx context.Context,
 	s *terraform.InstanceState,
 	d *terraform.InstanceDiff,
-	meta interface{}) (*terraform.InstanceState, diag.Diagnostics) {
+	meta interface{},
+) (*terraform.InstanceState, diag.Diagnostics) {
```
sreallymatt, line 909: these changes are due to my pre-commit hook that runs gofumpt, I can revert this if preferred

mildwonkey:
@sreallymatt - hi! I'd like to help you get this in. Out of curiosity, have you tested this out in the azure provider already? The core team "owns" this repo now, and we're willing to help you get this merged, but I'm not familiar enough with the SDK to give more than a basic review. Seeing this in action in your provider would go a long way to reassure me :)

Also - while I'm not asking you to fix everything! - is this the only method that isn't honoring datasource timeouts, or should we do a review? There's a code comment (which is clearly incorrect) suggesting timeouts are honored for data sources, and I'm curious if this was the only spot it's missing from, or just the one that was causing issues for you. 

(reply here or in slack, whatever works for you)

sreallymatt:
Hi @mildwonkey - I have tested this in the AzureRM provider by simply erroring on a data source Read and returning the timeout value that SDKv2 returns to providers. I'll push this to a branch. (sent on Slack)

As far as I can tell (based on limited experience with the inner workings of the SDK), this is the only method that needs to be modified, for resources `Apply` is called which already contains the correct logic to load provider and user-defined timeouts.

PR #1611, opened 2026-08-24
helper/schema: fix validation diagnostics pointing at the wrong sibling attribute path

### LINKED ISSUE #1175: Incorrect diagnostics location for ValidateDiagFunc
### SDK version

```
v2.25.0
```

```go
"aaa": {
	Optional:    true,
	Type:        schema.TypeList,
			MaxItems:    1,
			Elem: &schema.Resource{
					"start_date": {
						ValidateDiagFunc: validateRegex("^[1-9]\\d*$"),
						Optional:         true,
						Type:             schema.TypeString,
					},
					"end_date": {
						ValidateDiagFunc: validateRegex("^[1-9]\\d*$"),
						Optional:         true,
						Type:             schema.TypeString,
					},
					// some other fields (let's call them "abc", "def")
				},
			},
		},
```

```hcl
      aaa {
        end_date   = "test1"
        start_date  = "test2"
        abc = "abc_value"
        def = "def_value"
      }
```
Values for start and end_date does not pass the validation.
### Expected Behavior

╷
│ Error: value start_date: "test2" does not match the pattern "^[1-9]\\d*$"
│
│   with data.datasourcename.name,
│   on filename.tf line 132, in data "datasourcename" "name":
│  132:         start_date   = "test2"
│
╵
╷
│ Error: value end_date: "test1" does not match the pattern "^[1-9]\\d*$"
│
│   with data.datasourcename.name,
│   on filename.tf line 133, in data "datasourcename" "name":
│  133:         end_date  = "test1"
│
Both validation fails and the location of failed places is correct. It's especially useful with big and flat schemas.
### Actual Behavior

╷
│ Error: value start_date: "test2" does not match the pattern "^[1-9]\\d*$"
│
│   with data.datasourcename.name,
│   on filename.tf line 131, in data "datasourcename" "name":
│  131:         abc                  = "abc_value"
│
╵
╷
│ Error: value end_date: "test1" does not match the pattern "^[1-9]\\d*$"
│
│   with data.datasourcename.name,
│   on filename.tf line 131, in data "datasourcename" "name":
│  131:         abc                  = "abc_value"
│
The location where are the issues are incorrect: it's pointing that it failed on "abc" and "abc" (my duplication here was intended) while it should point to "start_date" and "end_date" where both validations failed. Please note that when trying the terraform plan several times, the incorrect lines reported by diagnostics changes (the bigger the schema is, the more it changes).
### Steps to Reproduce

1. `terraform init`
2. `terraform plan` (often can be tried several times depending on the schema size)
### Preanalysis ?

I did brief debugging on that topic and it seems that the root cause is in the `validateObject` function of the `helper/schema/schema.go` file. As a parameter it receives `path cty.Path` (where Path is a slice). As You know slice is actually a view over an array (it'll be needed later). Later (when there is iteration over `schema`), that path is used to build new path for validate function. If the validate returns something notempty, the it's saved into the diags.
And here where the root cause imo occurs: on next iteration it appends new schema element to path. As a result it overwrites the path value that was already stored into the diags as it's based on the same array. That causes incorrect path and duplication as it'll always show the last element of iterating over schema. 
So probably copying `path` slice in the-over-`schema` loop and using that fresh copy every time should fix the issue.

Regardless if You find results of my debugging useful it'd be great if You could please take care of that issue.

Best regards,
Wojciech

emmayusufu:
Validation diagnostics could point at the wrong attribute. The validate helpers build child paths with `append(path, step)` inside loops, and once an ancestor append has grown the backing array past its length, sibling iterations write into the same spare slot. Any `AttributePath` already stored on a diagnostic then silently changes to the last sibling visited, which is what #1175 reports: two errors pointing at the same attribute, moving between runs.

Fixed by copying the path before appending, via a small `appendPath` helper, at the five sites in `schema.go` and the two in `write_only.go` that store paths on diagnostics. `validateConfigNulls` in `grpc_provider.go` has the same append pattern but converts paths to protobuf at creation, so nothing stored can be corrupted; I left it alone.

The new test fails on main with both diagnostics reporting index 1.

Fixes #1175


emmayusufu:
The golangci-lint failures are all SA1019 deprecation findings in grpc_provider_test.go and helper/validation, none of which this PR touches. They come from the linter action running latest against the recently deprecated UnsafeToUseLegacyTypeSystem, and a clean checkout of main produces the identical 50 findings. Happy to fix those separately if wanted.

